Issue #20, opened 2021-02-26
Documentations


fuxingloh:
#### What would you like to be added:

Documentation to be implemented along side with code in the `docs/` directory. Docs to be deployed automatically on merged.

#### Why is this needed:

1. Allow feature to be released together with documentation on release.
2. Sharing the same concern, easy for community to contribute

Issue #19, opened 2021-02-26
Implementation core with error handling and protocol adapter


fuxingloh:
#### What would you like to be added:

`jellyfish-core` implementation must be:

1. modern with near zero dependencies to mitigate risk
2. support interceptors implementation but not implemented on day one
3. error handling at core, `Promise.reject()`
4. ability to adapt to any protocol when introduced, ws, https etc while being simple to use

/triage accepted
/area core

Issue #18, opened 2021-02-26
IEEE-754 conscious design, json-bigint, bignumbers & native number


fuxingloh:
#### What would you like to be added:

Built-in JSON/Object parser must parse number as `bignumber`, and development of library must take a beginner friendly stance of interoperability with big-number and number.

> https://www.npmjs.com/package/json-bigint

#### Why is this needed:

While most developers are aware of the IEEE 754 double standards, JSON specification does not say anything about number precision. Jellyfish implementation must be accommodating to various developer experience while preventing number precision mistake at large.

```txt
THIS IS VERY IMPORTANT FOR THE LIBRARY TO BE ACCOMMODATING TO VARIOUS DEVELOPER EXPERIENCE. 
```

```node
> parseFloat("0.000008") * 10
0.00007999999999999999
```

/area core
/triage accepted
/priority important-soon

fuxingloh:
> Choosing between simplicity over correctness.

Painfully, the default number implementation will lose precision when converting value. JavaScript has no internal `BigNumber` implementation, an external library must be used. This greatly increased complexity for your up and coming programmer. This is perplexing... 

```js
it('should not lose precision converting DFI', () => {
  const n = 1200000000.00000001
  const a = JSON.parse(JSON.stringify(n)) * 1.0e8
  expect(a.toString()).toBe("120000000000000001")
});
```

```diff
+ Expected: "120000000000000001"
- Received: "120000000000000000"
```

fuxingloh:
Don't think I have much choice on now this. What I can do is spent a significant amount of time educating programmings of the internal `BigNumber` implementation in the documentations https://github.com/DeFiCh/jellyfish/issues/20.

fuxingloh:
https://github.com/josdejong/lossless-json

Found this, this is awesome. There is no risk of losing numeric information when parsing JSON containing big numbers. `LosslessNumber` implements `valueOf()` and will throw an error when this conversion would result in a truncation of the number. This library does not force any specific implementation, you can bring your favourite BigNumber library.

> **How does it work?** The library works exactly the same as the native JSON.parse and JSON.stringify. The difference is that lossless-json preserves information of big numbers. lossless-json parses numeric values not as a regular number but as a LosslessNumber, a data type which stores the numeric value as a string. One can perform regular operations with a LosslessNumber, and it will throw an error when this would result in losing information.

[...]

Issue #17, opened 2021-02-26
Write once run everywhere; for browser, node and RN


fuxingloh:
#### What would you like to be added:

Jellyfish implementation must be platform agonistic and support all possible use case.

1. `<script src="jellyfish.js">` for browser with bundling
2. Electron
3. Node 10+ CJS and ESM
4. Reactive Native

#### Why is this needed:

To support the developer ecosystem, jellyfish implementation must plug and play for all platform.

/area core
/triage accepted

PR #29, opened 2021-03-02
RFC: monorepo jellyfish development

fuxingloh:
> A collection of JavaScript tools and libraries for DeFiChain to build decentralized finance on Bitcoin.

This is an early [1st milestone](https://github.com/DeFiCh/jellyfish/milestone/1) RFC to tackle **DeFiChain JS Library Protocol Implementation/Interfacing** technical design. There will be 4 milestones in total (protocols, wallet, remote & documentation) and this being the biggest unit of work due to the mono repo and utility and toolchain setup.

The purpose of this RFC is to provide a feedback loop before new features or significant change/implementation get introduced to the DeFiChain ecosystem. Extra attention is required for this as this repo is meant to be used by the masses. **As this is an RFC, NOT A CODE REVIEW, some codes are missing and this is not to be merged.** 

For the reviewer, the question is '**does this make sense for the JS ecosystem**' & '**is this easy to use**'?

# Jellyfish Monorepo

> All to be npm published. As jellyfish is all JavaScript related, this mono repo represent a collection of JavaScript tools and libraries for DeFiChain to build decentralized finance on Bitcoin. We can expect more to be added in the future.

- @defichain/jellyfish
- @defichain/jellyfish-core
- @defichain/jellyfish-jsonrpc
- @defichain/jellyfish-wallet (next milestone, by Q1 🤞)
- @defichain/testcontainers

## `@defichain/testcontainers`

Similar to testcontainers in the Java ecosystem, this library provide a lightweight, throwaway instances of `regtest`, `testnet` or `mainnet` provisioned automatically in Docker container. `@defichain/testcontainers` encapsulate on top of `defi/defichain:latest` and directly interface with the Docker REST API. see #20 

With `@defichain/testcontainers`, it allows the JS ecosystem to:

1. E2E application integration test without hassle
2. Parallel unit test runner as port number and container are dynamically generated on demand
3. Supercharge CI/CD workflow, run locally, run anyway, run on CI
4. Supercharge `jellyfish-core` development with integration test guarantees
5. Easy for any developer to test dApp on the DeFiChain JS ecosystem

### Usage Example

```ts
import {MainNetDocker, RegTestDocker, TestNetDocker} from '@defichain/testcontainers'

describe('reg test', () => {
  const node = new RegTestDocker()

  beforeEach(async () => {
    await node.start({
      user: 'foo',
      password: 'bar',
    })
    await node.ready()
  })

  afterEach(async () => {
    await node.stop()
  })

  it('should getmintinginfo and chain should be regtest', async () => {
    const client = Jellyfish.jsonrpc(await node.getRpcUrl())
    const result = await client.call('getmintinginfo', [])
    expect(result.result.chain).toBe('regtest')
  })
})

```

## `@defichain/jellyfish-core`

`@defichain/jellyfish-core` the protocol agnostic DeFiChain client implementation with APIs separated into their category.

### Implementation Details

- Strongly typed `*.d.ts` file will be exported for developer QOL
- Categorically separated concerns, for rapid develop of new features when we upgrade the nodes.
- Protocol agonistic implementation, we can easily swap out and add new REST/WS/MQ providers. see #19 
- Promise based implementation, see #16 

|category|method|JS|CCP|
|---|---|---|---|
|mining|getmintinginfo|`client.mining.getMintingInfo()`|`rpc/mining.cpp#getmintinginfo`|
|rawtransaction|sendrawtransaction|`client.transaction.sendRawTransaction(tx)`|`rpc/sendrawtransaction.cpp#sendrawtransaction`|

#### The client:

https://github.com/DeFiCh/jellyfish/blob/36d11134b74e9285ae36229fdaa3f33876bf3cee/packages/jellyfish-core/src/core.ts#L8-L12

#### The protocol interfacing, with strongly typed interfaces:

https://github.com/DeFiCh/jellyfish/blob/36d11134b74e9285ae36229fdaa3f33876bf3cee/packages/jellyfish-core/src/methods/mining.ts#L3-L23

#### This allow library consumer to use it as follow:

```js
const result = await client.blockchain.getBestBlockHash()
// or
client.blockchain.getBestBlockHash()
  .then((result) => {
    console.log(result)
  }).catch((err)=> {
    console.log('panic!')
  }).finally(() => {
    console.log('cleanup')  
  })
```

## `@defichain/jellyfish-jsonrpc`

Nothing really special about this package, it just imlements `@defichain/jellyfish-core` in the JSON-RPC 1.0 specification. 2 dependencies with 4 deeply are selected, `jsonbig` as JS number is implemented with IEEE-754, defi must not use floating point arithmetic for obvious reason (need to write TDD to guarantee this behavior) see #18. `cross-fetch` a isomorphic `fetch` client that is RN, Node.JS and browser compatible. See #17 

https://github.com/DeFiCh/jellyfish/blob/36d11134b74e9285ae36229fdaa3f33876bf3cee/packages/jellyfish-jsonrpc/src/jsonrpc.ts#L1-L3

https://github.com/DeFiCh/jellyfish/blob/36d11134b74e9285ae36229fdaa3f33876bf3cee/packages/jellyfish-jsonrpc/src/jsonrpc.ts#L21-L29

## `@defichain/jellyfish`

This is the entrypoint for most dApp developer. Distributed as `@defichain/jellyfish`, it is a bundler that creates 4 types of JavaScript packages for public use. This package provides conventional defaults and bundle all code required for dApps building. For library consumer, it is plug and play, they don't need to care how it works underneath.

### 1. `dist/jellyfish.cjs.js` for node.js

Common JS for Node.JS users, plug and play no transpiler required.

```js
const express = require('express')
const Jellyfish = require('@defichain/jellyfish')

const app = express()
const client = Jellyfish.jsonrpc('http://?.defichain.com')

app.get('/', (req, res) => {
  client.mining.getMintingInfo().then((info) => {
    res.send(JSON.stringify(info))
  })
})

app.listen(3000)
```

### 2. `dist/jellyfish.umd.js` for browser

For the global CDN users, this allows developer to just declare jellyfish.js script and use it on any webpage.

1. ReactJS Global Mode
2. VueJS Global Mode
3. WordPress websites
4. The plain old HTML & JavaScript

All library code that are required for jellyfish to work on their browser are automatically bundled into the fat javascript file. (20 KB+) This included `bignumber`, `JSONBig`, `jellyfish-jsonrpc` and in the future all other jellyfish tools e.g. `jellyfish-wallet`. The bundler is set to polyfill on >0.25% of all browser and target node 12 and above implementation.

```html

<script src="https://unpkg.com/@defichain/jellyfish@latest/dist/jellyfish.umd.js"/>
<script lang="js">
  var client = Jellyfish.jsonrpc('http://?.defichain.com')
  client.mining.getMintingInfo().then(info => {
    console.log(info)
  })
</script>
```

### 3. `dist/jellyfish.esm.js` for ES module

For webpack, parcel, and variety of users that want to use their own bundler (tree shaking), ES modules packages are also provided.

```js
import * as jf from '@defichain/jellyfish'

const client = jf.jsonrpc('http://?.defichain.com')
```

### 4. `dist/jellyfish.d.ts` for TS types

TypeScript's types are also exported for convenience’s sake, removes the need for `@types/jellyfish`.

```ts
import {Jellyfish, JellyfishClient} from '@defichain/jellyfish'

const client: JellyfishClient = Jellyfish.jsonrpc('http://?.defichain.com')
```

# Next?

If all is approved this implementation style will be internally frozen and slowly patched into main.

### DIFF
--- a/packages/jellyfish-core/src/core.ts
+++ b/packages/jellyfish-core/src/core.ts
@@ -1,0 +1,20 @@
+import * as methods from './methods'
+import {Client, Payload, JellyfishError} from './api'
+export {methods, Client, Payload, JellyfishError}
+
+/**
+ * Protocol agnostic DeFiChain node client, APIs separated into their category.
+ */
+export abstract class JellyfishClient implements Client {
+  readonly blockchain = new methods.Blockchain(this)
+  readonly mining = new methods.Mining(this)
+  readonly network = new methods.Network(this)
+  readonly wallet = new methods.Wallet(this)
+
+  // TODO(fuxingloh): all rpc categories to be implemented after RFC
+
+  /**
+   * Abstracted isomorphic promise based procedure call handling
+   */
+  abstract call<T>(method: string, payload: Payload): Promise<T>
+}
--- a/packages/jellyfish-core/src/methods/mining.ts
+++ b/packages/jellyfish-core/src/methods/mining.ts
@@ -1,0 +1,23 @@
+import {Client} from '../api'
+
+export interface MintingInfo {
+  blocks: number
+  difficulty: number
+  isoperator: boolean
+  networkhashps: number
+  pooledtx: number
+  chain: 'main' | 'test' | 'regtest'
+  warnings: string
+}
+
+export class Mining {
+  private readonly client: Client
+
+  constructor(client: Client) {
+    this.client = client
+  }
+
+  async getMintingInfo(): Promise<MintingInfo> {
+    return this.client.call('getmintinginfo', [])
+  }
+}
--- a/packages/jellyfish/src/jellyfish.ts
+++ b/packages/jellyfish/src/jellyfish.ts
@@ -1,0 +1,14 @@
+import {JellyfishJsonRpc} from '@defichain/jellyfish-jsonrpc'
+
+export const Jellyfish = {
+  /**
+   * Initialize a DeFiChain client and use jsonrpc for interfacing
+   *
+   * @param url
+   */
+  jsonrpc(url: string): JellyfishJsonRpc {
+    return new JellyfishJsonRpc(url)
+  },
+}
+
+export default Jellyfish
--- a/packages/testcontainers/src/docker.ts
+++ b/packages/testcontainers/src/docker.ts
@@ -1,0 +1,192 @@
+import Dockerode, {DockerOptions, Container} from 'dockerode'
+import fetch from 'node-fetch'
+import JSONBig from 'json-bigint'
+
+export type Network = 'mainnet' | 'testnet' | 'regtest'
+
+export interface StartOptions {
+  // TODO(fuxingloh): change to cookie based auth soon
+  user: string
+  password: string
+}
+
+/**
+ * DeFiChain defid node managed in docker
+ */
+export abstract class DeFiChainDocker {
+  protected static readonly PREFIX = 'defichain-testcontainers-'
+  protected readonly docker: Dockerode
+  protected readonly image = 'defi/defichain:1.5.0'
+  protected readonly network: Network
+
+  protected container?: Container
+  protected startOptions?: StartOptions
+
+  protected constructor(network: Network, options?: DockerOptions) {
+    this.docker = new Dockerode(options)
+    this.network = network
+  }
+
+  /**
+   * Generate a name for a new docker container with network type and random number
+   */
+  protected generateName(): string {
+    const rand = Math.floor(Math.random() * 10000000)
+    return `${DeFiChainDocker.PREFIX}-${this.network}-${rand}`
+  }
+
+  protected getCmd(opts: StartOptions): string[] {
+    return [
+      'defid',
+      '-printtoconsole',
+      '-rpcallowip=172.17.0.0/16',
+      '-rpcbind=0.0.0.0',
+      `-rpcuser=${opts.user}`,
+      `-rpcpassword=${opts.password}`,
+    ]
+  }
+
+  /**
+   * Start defid node on docker
+   *
+   * @param startOptions
+   */
+  async start(startOptions: StartOptions): Promise<void> {
+    this.startOptions = startOptions
+    this.container = await this.docker.createContainer({
+      name: this.generateName(),
+      Image: this.image,
+      Tty: true,
+      Cmd: this.getCmd(startOptions),
+      HostConfig: {
+        PublishAllPorts: true,
+      },
+    })
+    await this.container.start()
+  }
+
+  /**
+   * Get host machine port
+   *
+   * @param name of ExposedPorts e.g. '80/tcp'
+   */
+  async getPort(name: string): Promise<string> {
+    return new Promise((resolve, reject) => {
+      this.container!.inspect(function (err, data) {
+        if (err) {
+          reject(err)
+        } else {
+          resolve(data!.NetworkSettings.Ports[name][0].HostPort)
+        }
+      })
+    })
+  }
+
+  /**
+   * Get host machine port used for defid rpc
+   */
+  public abstract getRpcPort(): Promise<string>
+
+  /**
+   * Get host machine url used for defid rpc calls with auth
+   */
+  async getRpcUrl() {
+    const port = await this.getRpcPort()
+    const user = this.startOptions?.user
+    const password = this.startOptions?.password
+    return `http://${user}:${password}@127.0.0.1:${port}/`
+  }
+
+  /**
+   * Utility rpc function for the current node.
+   * This is not error checked, it will just return the raw result.
+   *
+   * @param method
+   * @param params
+   */
+  async call(method: string, params: any): Promise<any> {
+    const url = await this.getRpcUrl()
+    const response = await fetch(url, {
+      method: 'POST',
+      body: JSONBig.stringify({
+        jsonrpc: '1.0',
+        id: Math.floor(Math.random() * 10000000000000000),
+        method: method,
+        params: params,
+      }),
+    })
+    const text = await response.text()
+    return JSONBig.parse(text)
+  }
+
+  /**
+   * @param timeout millis, default to 15000 ms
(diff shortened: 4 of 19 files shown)


monstrobishi (approved):

packages/jellyfish-core/src/core.ts
```diff
@@ -0,0 +1,20 @@
+import * as methods from './methods'
+import {Client, Payload, JellyfishError} from './api'
+export {methods, Client, Payload, JellyfishError}
+
+/**
```
thedoublejay, line 5: Question, does this comment enables usage via JSDoc in VSC? Or it should have proper tags like @description, etc.

packages/jellyfish-core/src/methods/mining.ts
```diff
@@ -0,0 +1,23 @@
+import {Client} from '../api'
+
+export interface MintingInfo {
+  blocks: number
+  difficulty: number
+  isoperator: boolean
+  networkhashps: number
+  pooledtx: number
+  chain: 'main' | 'test' | 'regtest'
```
thedoublejay, line 9: I suggest if we can have an enum/const for this network variable. I guess it will also be widely used on apps.

packages/testcontainers/src/docker.ts
```diff
@@ -0,0 +1,192 @@
+import Dockerode, {DockerOptions, Container} from 'dockerode'
+import fetch from 'node-fetch'
+import JSONBig from 'json-bigint'
+
+export type Network = 'mainnet' | 'testnet' | 'regtest'
```
thedoublejay, line 5: ex. here, I think you have defined it as well on another file, might be good to reuse this

packages/testcontainers/src/docker.ts
```diff
@@ -0,0 +1,192 @@
...
+  /**
+   * Generate a name for a new docker container with network type and random number
+   */
+  protected generateName(): string {
+    const rand = Math.floor(Math.random() * 10000000)
+    return `${DeFiChainDocker.PREFIX}-${this.network}-${rand}`
+  }
+
+  protected getCmd(opts: StartOptions): string[] {
+    return [
```
thedoublejay, line 39: I think this function might need to accept some parameters. Like if they need to do `-reindex` or maybe if they can pass their config file like in the app `-conf`. You can ignore this if you're planning to do it in the future.

thedoublejay (changes requested):

packages/testcontainers/src/docker.ts
fuxingloh: I am using the polymorphic construct for this, it can be adapted as such

```ts
protected getCmd(opts: StartOptions): string[] {
  return [...super.getCmd(opts), '-regtest=1']
}
```

packages/jellyfish/src/jellyfish.ts
```diff
@@ -0,0 +1,14 @@
+import {JellyfishJsonRpc} from '@defichain/jellyfish-jsonrpc'
+
+export const Jellyfish = {
+  /**
+   * Initialize a DeFiChain client and use jsonrpc for interfacing
+   *
+   * @param url
+   */
+  jsonrpc(url: string): JellyfishJsonRpc {
+    return new JellyfishJsonRpc(url)
```
uzyn, line 10: I would call it `new Jellyfish()` if it makes sense so it's clean when it's used externally.

```js
new Jellyfish(url, {
    useNodeWallet: false, // default to false
}
```

packages/jellyfish-core/src/core.ts
fuxingloh: For IntelliJ IDEA it does, not sure about VSCode.

Additionally for JSDoc exported symbols work well for JS but in the TS environment it's kind of irritating above all else.

So e.g. like @param, @return paired with @description are useful symbols but TypeScript is naturally strongly typed, which nuanced the need for export symbols. So people argued for the sake of brevity it should just be plain old text.

[my motivation was from here](https://github.com/denoland/deno/blob/main/docs/contributing/style_guide.md#use-jsdoc-for-exported-symbols)

packages/jellyfish-core/src/methods/mining.ts
fuxingloh: Ultimately the `methods/*.ts` and their interfaces is to allow high separation of concern between rpc categories. There are a few benefits for not putting it into a const/enum.

1. enum sucks in an agnostic language ideology, it's basically a integer/string construct with added salt
2. code coupling, unnecessary coupling of variables that can be separated. someone wants to change something but this is owned by 100x code owner. 
3. poor developer experience, especially for declaring an interface. It's should be self contained. Check out the interface once without the need to travel down 10x files to find something
4. TypeScript already support interoperability/compatibility so user can declare their enum ontop

uzyn: Make sure to support `custom` too for privately-run chain. We also have an official `devnet` that's more for private internal testing use. https://github.com/DeFiCh/ain/blob/c5e2338eb540c8623bf5725284de1ef4eb67b898/src/chainparamsbase.cpp#L16

uzyn (approved):
Love the overall structure on the planning.

packages/jellyfish/src/jellyfish.ts
fuxingloh: Definitely will pay more attention to this, I rewrote that part a few times.

packages/jellyfish-core/src/methods/mining.ts
thedoublejay: Should be ok for me. It's more of preference I think. Main idea is to reuse especially these commonly used terms.

thedoublejay (approved):

fuxingloh:
Thanks for taking part in this RFC, the implementation style will be locked and slowly patched into main.

#### Things under consideration are:
1. structuring of interfaces readability vs reusability
2. constructing of `Jellyfish` client

PR #41, opened 2021-03-08
jellyfish-core protocol adapter and error handling

### LINKED ISSUE #16: Promise based client

Jellyfish to return `Promise<BlockInfo>` instead of using callback.

#### Why is this needed:

To prevent "callback hell" problem where code get nested and messy, making it harder to understand. As `async/await` is very mature, it brings better quality of life to development.

/area core
/triage accepted

fuxingloh:
/kind feature

This pull request implements the jellyfish-core abstraction for protocol adapting and error handling. 

`JellyfishClient` being an abstract class allows the ability to adapt to any protocol when introduced, ws, https etc while being simple to use. This implementation structure can be observed in `ContainerAdapterClient` where it is used to test `jellyfish-core` implementations.

RPC categories are grouped into `jellyfish-core/src/category/*.ts` e.g. `category/mining.ts` this allows a protocol agnostic implementation of the RPC. All concerns are grouped within one `ts` file for better developer experience of browsing and maintaining code. (might refactor this in the future, but we can leave it as it is now)

`JellyfishError` encapsulate RPC errors from DeFiChain within a structure. This allow for `instancesof` or `type of` error handling with rich structure.

Fixes #16 and #19

#### Additional comments?:

Also made changes to testcontainers:
1. removed `json-bigint` in favour of a new cleaner implementation that is to be introduced
2. added a raw `container.post(string): string` method for test adapting


fullstackninja864 (approved):

packages/jellyfish-core/src/category/mining.ts
```diff
@@ -0,0 +1,48 @@
...
+
+export interface MintingInfo {
+  blocks: number
+  currentblockweight?: number
+  currentblocktx?: number
+  difficulty: string
+  isoperator: boolean
+  masternodeid?: string
+  masternodeoperator?: string
+  masternodestate?: 'PRE_ENABLED' | 'ENABLED' | 'PRE_RESIGNED' | 'RESIGNED' | 'PRE_BANNED' | 'BANNED'
```
thedoublejay, line 11: Nice. I need to add this on my masternode PR (I only have enabled, pre_resigned and resigned)

thedoublejay (approved):

PR #50, opened 2021-03-10
implements JSON-RPC 1.0 specification

fuxingloh:

/kind feature

This PR implements the [`JSON-RPC 1.0`](https://www.jsonrpc.org/specification_v1) specification in `jellyfish-jsonrpc` for `jellyfish-core` abstraction of `JellyfishClient` & `JellyfishError`.

Implements parts of #17

#### Additional comments?:

Expanded on `JellyfishError` in core with `JellyfishRPCError` and `JellyfishClientError`. 

This error raising design follows many popular implementations such as the AmazonSDK, where user config problem & issues are surfaced as ClientError, while server/validation issues are categorized as another type, RPCError in this context.

### DIFF
--- a/packages/jellyfish-core/__tests__/container_adapter_client.ts
+++ b/packages/jellyfish-core/__tests__/container_adapter_client.ts
@@ -1,4 +1,4 @@
-import { JellyfishJSON, JellyfishClient, JellyfishError, Precision } from '../src/core'
+import { JellyfishJSON, JellyfishClient, Precision, JellyfishRPCError } from '../src/core'
 import { DeFiDContainer } from '@defichain/testcontainers'
 
 /**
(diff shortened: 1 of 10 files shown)


packages/jellyfish-core/__tests__/container_adapter_client.ts
```diff
@@ -29,8 +29,8 @@ export class ContainerAdapterClient extends JellyfishClient {
 
     const { result, error } = response
 
-    if (error !== null) {
-      throw new JellyfishError(error)
+    if (error !== undefined && error !== null) {
```
thedoublejay, line 32: is error != null better here? so it captures both undefined and null

thedoublejay (approved):

packages/jellyfish-core/__tests__/container_adapter_client.ts
fuxingloh: In JavaScript `undefined !== null` as the linting tool was complaining too, `null` works for that context but I added `undefined` to be extra sure. Usually people do this: `if (error) {` but the linting tool didn't like that either 😥.

PR #51, opened 2021-03-10
jellyfish bundles for umd, cjs, esm & d.ts

### LINKED ISSUE #17: Write once run everywhere; for browser, node and RN

fuxingloh:
/kind feature

This is the entrypoint for most dApp developer. Distributed as `@defichain/jellyfish`, it is a bundler that creates 4 types of JavaScript packages for public use. This package provides conventional defaults and bundle all code required for dApps building. For library consumer, it is plug and play, they don't need to care how it works underneath.

1. `dist/jellyfish.cjs.js` for node.js
2. `dist/jellyfish.umd.js` for browser
3. `dist/jellyfish.esm.js` for ES6 module
4. `dist/jellyfish.d.ts` for TS types

* `new jellyfish.Client('http://')` or `jellyfish.Client('http://...')` both works
* `export default {}` is used instead of `export default class`
* theses are exported
```ts
export default {
  Client,
  HttpProvider,
  OceanProvider
}
```

Fixes #17

#### Additional comments?:
I am using a nightly version of parcel, this is intentional because I want to follow the latest v2 release of the parcel-builder. As there is <100 LOC of actual code for `jellyfish`. There is nothing significant being exported and `tsc` is used for other packages in the mono repo. 

There is a feature I need (global mode) that is not yet supported, I am going wait for it to be supported. I have created an additional ticket https://github.com/DeFiCh/jellyfish/issues/52 to track it.

Might need to test the dist build work as expected.


packages/jellyfish/src/jellyfish.ts
```diff
@@ -1,3 +1,116 @@
...
+ */
+export const Jellyfish: {
+  Client: ClientConstructor
+  HttpProvider: HttpProviderConstructor
+  OceanProvider: OceanProviderConstructor
+} = {
+  Client,
+  HttpProvider,
+  OceanProvider
+}
```
fuxingloh, line 114: I decided to use a modern approach so that it's more friendly for >ES6 user but annoying for ES3 users. Future first.

packages/jellyfish/package.json
```diff
@@ -53,16 +53,13 @@
   },
   "dependencies": {
     "@defichain/jellyfish-core": "0.0.0",
-    "@defichain/jellyfish-jsonrpc": "0.0.0"
+    "@defichain/jellyfish-jsonrpc": "0.0.0",
+    "bignumber.js": "^9.0.1",
+    "lossless-json": "^1.0.4"
```
fuxingloh, line 58: Added for connivence for the library users, on core it's `peerDeps`.

fuxingloh:
> ## size-limit report 📦
> Path	Size	Loading time (3g)	Running time (snapdragon)	Total time
> packages/jellyfish/dist/jellyfish.umd.js	15.66 KB (+15622.55% 🔺)	314 ms (+3032.23% 🔺)	140 ms (+92.26% 🔺)	454 ms

This will keep growing as we add more features, I think we don't really have much choice for this. web3.js is 1.2 MB before gzip. Jellyfish is 44 KB before gzip, will keep growing.

thedoublejay (approved):

PR #88, opened 2021-03-27
update cross-fetch to 3.1.2

fuxingloh:
/kind dependencies

Updated `cross-fetch` to `3.1.2`, this introduces better typescript support but causes breaking change to the exported types. Requiring a manual patch.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -15467,9 +15467,9 @@
       }
     },
     "node_modules/cross-fetch": {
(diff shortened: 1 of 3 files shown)


fuxingloh:
**CI / Size Limit**: failed due to changed dependency, that workflow can be ignored for now.

thedoublejay (approved):

Issue #48, opened 2021-03-09
DeFiChain Jellyfish RPC Implementations


fuxingloh:
> This tracks all the RPC methods and categories being implemented in `jellyfish-core`. Will list all the methods here, and strike out those that are not applicable.

### `blockchain.cpp`
- [ ] getblockchaininfo
- [ ] getchaintxstats
- [ ] getblockstats
- [ ] getbestblockhash
- [ ] getblockcount
- [ ] getblock
- [ ] getblockhash
- [ ] getblockheader
- [ ] getchaintips
- [ ] getdifficulty
- [ ] getmempoolancestors
- [ ] getmempooldescendants
- [ ] getmempoolentry
- [ ] getmempoolinfo
- [ ] clearmempool
- [ ] getrawmempool
- [ ] gettxout
- [ ] gettxoutsetinfo
- [ ] pruneblockchain
- [ ] savemempool
- [ ] verifychain
- [ ] preciousblock
- [ ] scantxoutset
- [ ] getblockfilter
- [ ] invalidateblock
- [ ] reconsiderblock
- [ ] waitfornewblock
- [ ] waitforblock
- [ ] waitforblockheight
- [ ] syncwithvalidationinterfacequeue

### `mining.ccp`
* [x] getnetworkhashps
* [x] getmintinginfo
* [ ] prioritisetransaction
* [ ] getblocktemplate
* [ ] submitblock
* [ ] submitheader
* [ ] generatetoaddress
* [ ] estimatesmartfee
* [ ] estimaterawfee

### `misc.cpp`
- [ ] getmemoryinfo
- [ ] logging
- [ ] validateaddress
- [ ] createmultisig
- [ ] deriveaddresses
- [ ] getdescriptorinfo
- [ ] verifymessage
- [ ] signmessagewithprivkey
- [ ] setmocktime
- [ ] echo
- [ ] echojson

### `net.cpp`
- [ ] getconnectioncount
- [ ] ping
- [ ] getpeerinfo
- [ ] addnode
- [ ] disconnectnode
- [ ] getaddednodeinfo
- [ ] getnettotals
- [ ] getnetworkinfo
- [ ] setban
- [ ] listbanned
- [ ] clearbanned
- [ ] setnetworkactive
- [ ] getnodeaddresses

### `rawtransaction.cpp`
- [ ] getrawtransaction
- [ ] createrawtransaction
- [ ] decoderawtransaction
- [ ] decodescript
- [ ] sendrawtransaction
[...]

PR #91, opened 2021-04-01
blockchain.getBlock, getBlockHeight, getBlockHash

canonbrother:

/kind feature

Implemented type for RPC: #48
1. blockchain.getBlock
2. blockchain.getBlockHash
3. blockchain.getBlockCount


packages/jellyfish-api-core/src/category/blockchain.ts
```diff
@@ -17,6 +20,36 @@ export class Blockchain {
...
+   * Get block data with particular header hash.
+   * @param blockHash
+   * @param verbosity optional, default is 1, 0 for hex encoded
+   * @return Promise<BlockVerbo>
+   */
+  async getBlock (blockHash: string, verbosity: number): Promise<BlockVerbo2>
+  async getBlock (blockHash: string, verbosity: number): Promise<BlockVerbo1>
+  async getBlock (blockHash: string, verbosity: number): Promise<BlockVerbo0>
+  async getBlock (blockHash: string, verbosity: number): Promise<BlockVerbo> {
+    return await this.client.call('getblock', [blockHash, verbosity], 'number')
```
fuxingloh:
```ts
async getBlock (blockHash: string, verbosity: 2): Promise<BlockVerbo2>
async getBlock (blockHash: string, verbosity: 1): Promise<BlockVerbo1>
async getBlock (blockHash: string, verbosity: 0): Promise<string>
async getBlock (blockHash: string, verbosity: 0 | 1 | 2): Promise<string | BlockBase> {
}
```

You can do something like this so that the type will always be correct based on the verbosity.

packages/jellyfish-api-core/__tests__/category/blockchain.test.ts
```diff
@@ -55,26 +56,89 @@ describe('masternode', () => {
...
-    const info = await client.blockchain.getBlockchainInfo()
+  describe('getBlock', () => {
+    const height: number = 1
+    let blockHash: string = ''
+    let verbosity: number = 0
 
-    expect(info.chain).toBe('regtest')
-    expect(info.blocks).toBeGreaterThan(0)
-    expect(info.headers).toBeGreaterThan(0)
+    beforeAll(async () => {
```
fuxingloh, line 89: Since the height is always gonna be 1, it's easier to read if it's just like this.
```ts
    beforeAll(async () => {
      blockHash = await client.blockchain.getBlockHash(1)
    })
```

packages/jellyfish-api-core/__tests__/category/blockchain.test.ts
```diff
@@ -55,26 +56,89 @@ describe('masternode', () => {
...
+      expect(info.size_on_disk).toBeGreaterThan(0)
+      expect(info.pruned).toBe(false)
     })
+  })
 
-    const info = await client.blockchain.getBlockchainInfo()
+  describe('getBlock', () => {
+    const height: number = 1
+    let blockHash: string = ''
+    let verbosity: number = 0
```
fuxingloh, line 87: Moving verbosity to the `it('')` is cleaner, making it easier to read + making the test logic and code referencing experience less tiring also. (don't need to scroll up and down to figure out a test case, if all the logic is grouped together.)

e.g.
```ts
      const data = await client.blockchain.getBlock(blockHash, 0)
      expect(data.length).toBe(...)
```

fuxingloh: You can do too
```ts
const verbosity = 0
const data = await client.blockchain.getBlock(blockHash, verbosity)
expect(data.length).toBe(...)
```

packages/jellyfish-api-core/src/category/blockchain.ts
fuxingloh:
```ts
type BlockVerbo0 = 'string'
export type BlockVerbo = BlockVerbo2 | BlockVerbo1 | BlockVerbo0 
```

will not be required when it's done as above.

packages/jellyfish-api-core/src/category/blockchain.ts
```diff
@@ -17,6 +20,36 @@ export class Blockchain {
...
+    return await this.client.call('getblockhash', [height], 'number')
+  }
+
+  /**
+   * Get the height of the most-work fully-validated chain.
+   * @return Promise<number>
+   */
+  async getBlockCount (): Promise<number> {
+    return await this.client.call('getblockcount', [], 'number')
+  }
```
fuxingloh, line 39: Missing test coverage for this.

packages/jellyfish-api-core/src/category/blockchain.ts
```diff
@@ -43,3 +76,73 @@ export interface BlockchainInfo {
...
+  hash: string
+  confirmations: number
+  strippedsize: number
+  size: number
+  weight: number
+  height: number
+  masternode: string
+  minter: string
+  mintedBlocks: any
+  stakeModifier: any
```
fuxingloh, line 90: What are this `any`?

fuxingloh (changes requested):

packages/jellyfish-api-core/src/category/blockchain.ts
```diff
@@ -43,3 +76,73 @@ export interface BlockchainInfo {
...
+  nextblockhash: string
+}
+
+export interface BlockVerbo1 extends BlockBase {
+  tx: string[]
+}
+
+export interface BlockVerbo2 extends BlockBase {
+  tx: RawTx[]
+}
```
fuxingloh, line 110: you might want to use generics also to keep it simple.

`Block<RawTx>`
`Block<string>`

```ts
export interface Block<T> {
  //...
  chainwork: string
  tx: T[]
  nTx: number
  //...
}
```

packages/jellyfish-api-core/src/category/blockchain.ts
canonbrother: ohh.. i missed out this!!

packages/jellyfish-api-core/src/category/blockchain.ts
```diff
@@ -17,6 +17,33 @@ export class Blockchain {
...
+
+  /**
+   * Get block data with particular header hash.
+   * @param blockHash
+   * @param verbosity optional, default is 1, 0 for hex encoded
+   * @return Promise<string | Block>
+   */
+  async getBlock<T> (blockHash: string, verbosity?: 0 | 1 | 2): Promise<string | Block<T>> {
+    return await this.client.call('getblock', [blockHash, verbosity], 'number')
+  }
```
fuxingloh, line 46: You should resolve the generics for known verbosity with overloads, e.g. below

```ts
  /**
   * Get block data with particular header hash.
   * @param hash of the block
   * @param verbosity 0, 1, 2
   * @return Promise<string | Block<string> | Block<RawTx>>
   */
  async getBlock (hash: string, verbosity: 0): Promise<string>
  async getBlock (hash: string, verbosity: 1): Promise<Block<string>>
  async getBlock (hash: string, verbosity: 2): Promise<Block<RawTx>>
  async getBlock<T> (hash: string, verbosity: 0 | 1 | 2): Promise<string | Block<T>> {
    return await this.client.call('getblock', [hash, verbosity], 'number')
  }
```

packages/jellyfish-api-core/__tests__/category/blockchain.test.ts
```diff
@@ -55,26 +56,103 @@ describe('masternode', () => {
...
-    expect(info.pruned).toBe(false)
+    it('test getblock with verbo 1', async () => {
+      const data: string | Block<string> = await client.blockchain.getBlock<string>(blockHash, 1)
+      const block = data as Block<string>
+      expect(block).toHaveProperty('hash')
+      expect(block.confirmations).toStrictEqual(2)
+      expect(block.strippedsize).toStrictEqual(360)
+      expect(block.size).toStrictEqual(396)
+      expect(block.weight).toStrictEqual(1476)
+      expect(block.height).toStrictEqual(1)
```
fuxingloh, line 104: The test conditions are too strict as the masternode auto minting is time bounded to auto mint every second-ish. Currently, the way the test is written sequentially. But the test runner might not run it in that order, especially when I set to run a sub set of tests.

You can do something like this, where you test conditions 'after' and 'close to' what is supposed to be.
```ts
      expect(info.chain).toBe('regtest')
      expect(info.blocks).toBeGreaterThan(0)
      expect(info.headers).toBeGreaterThan(0)

      expect(info.bestblockhash.length).toBe(64)
      expect(info.difficulty).toBeGreaterThan(0)
      expect(info.mediantime).toBeGreaterThan(1550000000)
```

This applies for all tests, written.

fuxingloh (changes requested):

fuxingloh (approved):

PR #94, opened 2021-04-05
blockchain.getTxOut, wallet.listUnspent

canonbrother:

/kind feature

Implemented type for RPC: #48
1. blockchain.getTxOut
2. wallet.listUnspent


packages/jellyfish-json/src/remap.ts
```diff
@@ -24,14 +24,20 @@ enum MappingAction {
  * @param precision PrecisionMapping is a key value pair to allow revive value type
  */
 export function remap (text: string, precision: PrecisionMapping): any {
-  const losslessObj = parse(text)
+  const { result: losslessObj } = parse(text)
```
canonbrother, line 27: FYI, the ApiClient.call contains the `result` in resp
and gonna add the `result` in resp manually as well after remapped in ln: 40

fuxingloh (changes requested):
some minor changes

fuxingloh (approved):

PR #97, opened 2021-04-05
json reviveLosslessWithKeys

### LINKED ISSUE #93: dumb revive json-rpc value type

"smart revive" blockchain amount as bignumber, the rest numeric remains as number.

#### Why is this needed:

strict type and better view for user

canonbrother:

/kind fix

more flexible in giving type assertion on jsonrpc resp

Fixes #93


packages/jellyfish-json/__tests__/index.test.ts
```diff
@@ -40,6 +40,30 @@ describe('parse', () => {
...
+        "value": 1200000000.00000001,
+        "scriptPubKey": {
+          "asm": "OP_DUP OP_HASH160 b36814fd26190b321aa985809293a41273cfe15e OP_EQUALVERIFY OP_CHECKSIG",
+          "hex": "76a914b36814fd26190b321aa985809293a41273cfe15e88ac",
+          "reqSigs": 1,
+          "type": "pubkeyhash",
+          "addresses": [ "mwsZw8nF7pKxWH8eoKL9tPxTpaFkz7QeLU"]
+        },
+        "coinbase": true
+      }`, { value: 'bignumber', confirmations: 'number', reqSigs: 'lossless' })
```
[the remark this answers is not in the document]
fuxingloh: since reviver does not support this feature natively (parent key cannot be access in the reviver), you can parse all as `'lossless'` first.

Then remap them after parsing to lossless, something like this:

```ts
const obj = parse(text)
for (let key of obj) {
  if (precision[key]) {
    const value = obj[key]
    obj[key] = ...
  }
}
```

fuxingloh (changes requested):

fuxingloh:
/kind feature

More of a feature than fix, I re-title it as it will be used as release notes.

packages/jellyfish-json/__tests__/index.test.ts
```diff
@@ -40,6 +40,44 @@ describe('parse', () => {
...
+            "nested": 1200000000.00000001
+          }
+        },
+        "coinbase": true
+      }`, {
+        value: 'bignumber',
+        confirmations: 'lossless',
+        scriptPubKey: {
+          reqSigs: 'number',
+          custom: {
```
canonbrother, line 66: specially test this nested object after implement the recursive check on precision map

packages/jellyfish-json/__tests__/index.test.ts
```diff
@@ -40,6 +40,44 @@ describe('parse', () => {
...
+            nested: 'bignumber'
+          }
```
canonbrother, line 68: specially test this nested object after implement the recursive check on precision map

packages/jellyfish-json/src/index.ts
```diff
@@ -28,6 +35,61 @@ const reviveLosslessAs = (transformer: (string: string) => any) => {
...
+/**
+ * Remap lossless
+ * @param precision Precision - 'bignumber', 'number', 'lossless'
+ * @param losslessObj lossless json object
+ * @returns losslessObj
+ */
+const remapLosslessObj = (precision: PrecisionMapping, losslessObj: any): any => {
+  for (const k in precision) {
+    const value = losslessObj[k]
+    if (value !== undefined && value !== null) {
```
fuxingloh, line 57: I think you can rewrite it like this to reduce code complexity, (by returning/exiting early) Not sure how much complexity it will be reduced.

```ts
if (value === undefined && value === null) { 
  continue
}

const precisionType = precision[k] as Precision
if (typeof precisionType !== 'string') {
  remapLosslessObj(precisionType, losslessObj[k])
} else if (value instanceof LosslessNumber) {
  losslessObj[k] = revive(precisionType, value)
}
```

packages/jellyfish-json/__tests__/index.test.ts
fuxingloh: I think can expand and add some edge test cases, where the JSON to parse is different from the mapping.

e.g.
```ts
JSON = '{"value": {}, "nested": 1}'
mapping = { 
  value: 'bignumber',
  nested: {
    something: 'bignumber'
  }
}
```

fuxingloh (changes requested):

packages/jellyfish-json/__tests__/index.test.ts
canonbrother: yea it should throw error if providing invalid or unmatch precision mapping.
tests are added
and feel free to bring up a better approach if any.

fuxingloh (approved):

PR #101, opened 2021-04-08
remap array handling

canonbrother:

/kind fix

To cater any kinds of object structure on view

Fixes #
Add array handling on remap

### DIFF
--- a/packages/jellyfish-json/src/remap.ts
+++ b/packages/jellyfish-json/src/remap.ts
@@ -27,25 +27,17 @@ enum MappingAction {
 export function remap (text: string, precision: PrecisionMapping): any {
   const losslessObj = parse(text)
 
-  const errMessage = Array.isArray(losslessObj)
-    ? bulkValidate(losslessObj, precision)
-    : validate(losslessObj, precision)
-  if (errMessage !== undefined) {
-    throw new Error(errMessage)
+  if (Array.isArray(losslessObj)) {
+    return losslessObj.map((v: any) => {
+      const errMsg = validate(v, precision)
+      if (errMsg !== undefined) throw new Error(errMsg)
+      return remapLosslessObj(v, precision)
+    })
   }
 
-  return Array.isArray(losslessObj)
-    ? (losslessObj).map(v => remapLosslessObj(v, precision))
-    : remapLosslessObj(losslessObj, precision)
-}
-
-function bulkValidate (losslessObj: any, precision: PrecisionMapping): string | undefined {
-  for (let i = 0; i < losslessObj.length; i += 1) {
-    const errorMessage = validate(losslessObj[i], precision)
-    if (errorMessage !== undefined) {
-      return errorMessage
-    }
-  }
+  const errMsg = validate(losslessObj, precision)
+  if (errMsg !== undefined) throw new Error(errMsg)
+  return remapLosslessObj(losslessObj, precision)
 }
 
 function validate (losslessObj: any, precision: PrecisionMapping): string | undefined {
@@ -123,38 +115,23 @@ function remapLosslessObj (losslessObj: any, precision: PrecisionMapping): any {
 }
 
 function getAction (value: any, type: Precision | PrecisionMapping): MappingAction {
-  const mappingAction = getPrecisionMappingAction(value, type)
-
-  if (mappingAction === MappingAction.NONE) {
-    return getDeeplyMappingAction(value, type)
-  }
-
-  return mappingAction
-}
-
-function getPrecisionMappingAction (value: any, type: Precision | PrecisionMapping): MappingAction {
   // typed with precision
   if (typeof type === 'string' && value instanceof LosslessNumber) {
     return MappingAction.PRECISION
   }
 
-  // precision in loop
   if (typeof type === 'string' && Array.isArray(value)) {
     return MappingAction.PRECISION_LOOP
+  }
+
+  // deeply with mapping
+  if (typeof type === 'object') {
+    return MappingAction.DEEPLY_PRECISION_MAPPING
   }
 
   // number as default
   if (value instanceof LosslessNumber) {
     return MappingAction.DEFAULT_NUMBER
-  }
-
-  return MappingAction.NONE
-}
-
-function getDeeplyMappingAction (value: any, type: Precision | PrecisionMapping): MappingAction {
-  // deeply with mapping
-  if (typeof type === 'object') {
-    return MappingAction.DEEPLY_PRECISION_MAPPING
   }
 
   // deeply with unknown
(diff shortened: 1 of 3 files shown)


fuxingloh (changes requested):
This is getting painfully complex 😥

To reduce code complex and potential long term code debt, we should reduce the implementation to 'bignumber' or default to 'number'. We painfully need to keep this simple.

We should simplify to PrecisionMapping to `'bignumber' | PrecisionObject | PrecisionObject[]`. 

json|mapping
---|---
`[1,2,3]` | `'bignumber'`
`[{a:1}]` | `{a:'bignumber'} or [{a:'bignumber'}]`
`{a:1}` | `{a:'bignumber'}`
`{a: [1]}` | `{a:'bignumber'}`
`{a: [{b: 'bignumber'}]}` | `{a:{b: 'bignumber'}}`
etc | ...

```ts
export type PrecisionMapping = 'bignumber' | PrecisionObject | PrecisionObject[]

export interface PrecisionObject {
  [key: string]: 'bignumber' | PrecisionMapping
}

export function remap (text: string, precision: PrecisionMapping): any {

}
```

I wrote some test cases for you:

```ts
it('should remap object at root with key mapping', () => {
  const parsed = JellyfishJSON.parse(`{
    "big": 123.4,
    "num": 1000
  }`, {
    big: "bignumber"
  })
  expect(parsed.big instanceof BigNumber).toBe(true)
  expect(parsed.num).toBe(1000)
});

it('should remap object deeply with key mapping', () => {
  const parsed = JellyfishJSON.parse(`{
    "deeply": {
      "big": 123,
      "num": 1000
    }
  }`, {
    deeply: {
      big: "bignumber"
    }
  })
  expect(parsed.deeply.big instanceof BigNumber).toBe(true)
  expect(parsed.deeply.num).toBe(1000)
});

it('should remap all array with same precision', () => {
  const parsed = JellyfishJSON.parse(`[
    100,
    200
  ]`, 'bignumber')

  expect(parsed[0] instanceof BigNumber).toBe(true)
  expect(parsed[1] instanceof BigNumber).toBe(true)
});

it('should remap all array object with same precision', () => {
  const parsed = JellyfishJSON.parse(`[
    {
      "big": 100
    },
    {
      "big": 200.2
    }
  ]`, {
    big: 'bignumber'
  })

  expect(parsed[0].big instanceof BigNumber).toBe(true)
  expect(parsed[1].big instanceof BigNumber).toBe(true)
});

it('should remap all array with different precision and unmapped default to number', () => {
  const parsed = JellyfishJSON.parse(`[
    {
      "big": 123
    },
    {
      "big": 123.4
    },
    {
      "big": 123.45
    },
    {
      "big": 123.456
    }
  ]`, [
    {
      big: 'bignumber'
    },
    {},
    {
      big: 'bignumber'
    },
  ])

  expect(parsed[0] instanceof BigNumber).toBe(true)
  expect(parsed[1]).toBe(123.4)
  expect(parsed[2] instanceof BigNumber).toBe(true)
  expect(parsed[3]).toBe(123.456)
});

it('should remap all array object with same precision deeply', () => {
  const parsed = JellyfishJSON.parse(`[
    {
      "deeply": {
        "big": 123
      }
    },
    {
      "deeply": {
        "big": 123.4
      }
    }
  ]`, {
    deeply: {
      big: 'bignumber'
    }
  })

  expect(parsed[0].deeply.big instanceof BigNumber).toBe(true)
  expect(parsed[1].deeply.big instanceof BigNumber).toBe(true)
});
```

fuxingloh:
1 more test condition

```ts
`{
  items: [1,2,3]
}`,
{
  items: 'bignumber'
}
```

packages/jellyfish-json/src/remap.ts
```diff
@@ -26,42 +27,84 @@ enum MappingAction {
...
+    ? bulkValidate(losslessObj, precision)
+    : validate(losslessObj, precision)
   if (errMessage !== undefined) {
     throw new Error(errMessage)
   }
 
-  return remapLosslessObj(losslessObj, precision)
+  return Array.isArray(losslessObj)
+    ? (losslessObj).map(v => remapLosslessObj(v, precision))
+    : remapLosslessObj(losslessObj, precision)
```
fuxingloh, line 39: quite hard to read this, how about making it more linear:

```ts
if (Array.isArray(losslessObj)) {
  // array condition with validate
}

// non array condition with validate

```

fuxingloh: you can also so you don't need bulk method and only need to loop once.

```ts
return (losslessObj).map(v => {
  validate(v, precision)
  remapLosslessObj(v, precision)
})
```

packages/jellyfish-json/src/remap.ts
```diff
@@ -71,7 +114,7 @@ function remapLosslessObj (losslessObj: any, precision: PrecisionMapping): any {
         break
 
       case MappingAction.DEEPLY_UNKNOWN:
-        remapLosslessObj(losslessObj[k], {})
+        remapLosslessObj(losslessObj[k], precision)
```
fuxingloh, line 117: this should be `{}`

should not pass unknown mapping deeply down

Not a good idea: `{result: {a: 123} } != {a: 'bignumber'}`

Unexpected side effect of  `{foo: {a: 123}, a: 123 } != {a: 'bignumber'}` as `foo.a` should not be parsed.

packages/jellyfish-json/src/remap.ts
```diff
@@ -80,49 +123,46 @@ function remapLosslessObj (losslessObj: any, precision: PrecisionMapping): any {
...
+  const mappingAction = getPrecisionMappingAction(value, type)
+
+  if (mappingAction === MappingAction.NONE) {
+    return getDeeplyMappingAction(value, type)
+  }
+
+  return mappingAction
+}
+
+function getPrecisionMappingAction (value: any, type: Precision | PrecisionMapping): MappingAction {
```
fuxingloh, line 135: it's okay to keep the if condition long, don't care the codeclimate warning on this.

fuxingloh (changes requested):

packages/jellyfish-json/src/remap.ts
canonbrother: yea you're right

canonbrother: I remember why I put `precision` to loss the restriction 
The ApiClient.call is currently return the resp like this `{result: {jsonvalue} , error, id}`
when doing parse we just need to put

canonbrother: @fuxingloh 
here I opened issue and also for a note on current loose precision parse
hmm.. its not critical... I added the convince-able note as well 
https://github.com/DeFiCh/jellyfish/issues/104

fuxingloh:
With #104, I will merge this in once all the test passes, thanks @canonbrother

fuxingloh (approved):

PR #112, opened 2021-04-13
blockchain.getRawMempool, wallet.getNewAddress, wallet.getAddressInfo, wallet.validateAddress, wallet.listAddressGroupings, wallet.getWalletInfo, wallet.setWalletFlag, wallet.sendToAddress, wallet.createWallet

canonbrother:
#### What kind of PR is this?:

/kind feature

#### What this PR does / why we need it:
Implemented type for RPC: #48
1. blockchain.getRawMempool
2. wallet.getNewAddress
3. wallet.getAddressInfo
4. wallet.validateAddress
5. wallet.listAddressGroupings
6. wallet.getWalletInfo
7. wallet.setWalletFlag
8. wallet.sendToAddress
9. wallet.createWallet

PR #124, opened 2021-04-14
api-core to use named export instead of *

fuxingloh:

/kind refactor

To prevent conflicts due to the massive amount of data to type, export should be named.

/cc @canonbrother might cause your pending PR to break, hence this should merge after #112 and then this PR should be updated.


fuxingloh:
Closes in favour of #130

PR #125, opened 2021-04-15
new testcontainers utils for raw tx QOL testing

fuxingloh:

/kind feature

new testcontainers utils for better raw tx testing quality of life.

1. `container.waitForCondition(async () => false, 3000))`
2. `container.waitForWalletBalanceGTE(200)`
3. `container.fundAddress(address, 10)`
4. `const { address, privKey } = await container.newAddressAndPrivKey()`

This makes it easier to test condition that require balance, funded address, or priv key.

#### Additional comments:
Also added `{ v: true }` flag so that associated docker volumes get removed after the test.


canonbrother (approved):
great utils here for happy dev

PR #127, opened 2021-04-16
rawtx.createRawTransaction, rawtx.signRawTransactionWithKey, rawtx.testMempoolAccept, rawtx.sendRawTransaction

fuxingloh:
/kind feature

Added types for `rawtx.ts` #48 

1. createRawTransaction
2. signRawTransactionWithKey
3. testMempoolAccept
4. sendRawTransaction

I added `//TODO` comments for some parts of the code as it's beyond my scope to implement them. Need to work on others features now.

### DIFF
--- a/packages/jellyfish-api-core/src/category/rawtx.ts
+++ b/packages/jellyfish-api-core/src/category/rawtx.ts
@@ -49,20 +49,19 @@ export class RawTx {
    *
    * @param {string} rawTx unsigned raw transaction
    * @param {string[]} privKeys array of base58-encoded private keys for signing (WIF)
-   * @param {SignRawTxWithKeyPrevTx[]} prevTxs array of previous dependent transaction outputs
    * @param {SignRawTxWithKeyOptions} options
-   * @param {SigHashType} options.sigHashType The signature hash type to use
+   * @param {SigHashType} options.sigHashType the signature hash type to use
+   * @param {SignRawTxWithKeyPrevTx[]} options.prevTxs array of previous dependent transaction outputs
    * @return {Promise<SignRawTxWithKeyResult>}
    */
   async signRawTransactionWithKey (
     rawTx: string,
     privKeys: string[],
-    prevTxs?: SignRawTxWithKeyPrevTx[],
     options: SignRawTxWithKeyOptions = {}
   ): Promise<SignRawTxWithKeyResult> {
     const { sigHashType = SigHashType.ALL } = options
     return await this.client.call('signrawtransactionwithkey', [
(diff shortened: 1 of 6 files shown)


packages/jellyfish-api-core/src/category/rawtx.ts
```diff
@@ -0,0 +1,202 @@
...
+   * @param {SignRawTxWithKeyPrevTx[]} prevTxs array of previous dependent transaction outputs
+   * @param {SignRawTxWithKeyOptions} options
+   * @param {SigHashType} options.sigHashType The signature hash type to use
+   * @return {Promise<SignRawTxWithKeyResult>}
+   */
+  async signRawTransactionWithKey (
+    rawTx: string,
+    privKeys: string[],
+    prevTxs?: SignRawTxWithKeyPrevTx[],
+    options: SignRawTxWithKeyOptions = {}
```
canonbrother, line 61: is there any reason put the `prevTxs` out of options??
else we follow back the deno style
`Exported functions: max 2 args, put the rest into an options object.`
https://github.com/denoland/deno/blob/main/docs/contributing/style_guide.md

packages/jellyfish-api-core/src/category/rawtx.ts
```diff
@@ -0,0 +1,202 @@
...
+    rawTx: string,
+    privKeys: string[],
+    options: SignRawTxWithKeyOptions = {}
+  ): Promise<SignRawTxWithKeyResult> {
+    const { sigHashType = SigHashType.ALL } = options
+    return await this.client.call('signrawtransactionwithkey', [
+      rawTx, privKeys, options.prevTxs, sigHashType
+    ], 'number')
+  }
+
```
canonbrother, line 67: ahh.. one more..
```
const { privKeys = [], sigHashType = SigHashType.ALL } = options
  return await this.client.call('signrawtransactionwithkey', [
    rawTx, privKeys, prevTxs, sigHashType
], 'number')
```

canonbrother (commented):
The tests are super solid 👏👏

packages/jellyfish-api-core/src/category/rawtx.ts
fuxingloh: I was thinking we can require privKeys param instead of options. But now that I think about it we should standardise and enforce this requirement. It's much cleaner that way too. 👍🏻

fuxingloh: 1. 1 arg + 1 option
2. 2 arg max
3. 1 option

canonbrother (approved):

PR #128, opened 2021-04-16
fix jellyfish-json path based bignumber remapping

### LINKED ISSUE #104: strict json precision parsing
Its not critical but also a note.
Currently the precision is loose on
while:
```
parse (obj, type): any
parse({ foo: { bar: 123 }, bar: 123 }, { bar: 'bignumber' }
```

result:
```
{ foo: { bar: { BigNumber }}, bar: { BigNumber }}
```

expected:
```
{ foo: { bar: 123 }, bar: { BigNumber }}
```
note: 
Two issues will be facing while implement the strict precision
1. `ApiClient.call`'s resp object structure - `{ result, error, id }`
Ideally for rpc call is only passing the relevant precision mapping object
```
// better
getBlock(...args) {
  return await this.client.call(m, [], { number: 'bignumber' })
}

// Not so friendly
getBlock(...args) {
  return await this.client.call(m, [], { result: { number: 'bignumber' }})
}
```
2. Nested object will not be able to get parsed
Currently in order to remove the loose precision is to replace the `precision` by empty object `{}`
```
// remap.ts ln: 108-110
case MappingAction.DEEPLY_UNKNOWN:
  -- remapLosslessObj(losslessObj[k], precision)
  ++ remapLosslessObj(losslessObj[k], {})
  break
```
this causes the nested object cannot be parsed due to empty precision.

#### Why is this needed:
- technically `{foo: { bar: 123 }} !== { bar: 123 }`
- for note

But its also convince-able to remain the loose precision
1. Rare case in real scenario
```
parse({foo: {bar: 123}, bar: 123}}, {bar: 123})
```
assume `bar` is a crypto amount, so why not parsing both??

2. Its doable for but its required extra struct specification
```
parse({foo: {bar: 123}, bar: 123}}, { foo: {bar: 'number'}, bar: 'bignumber' })
// foo.bar remains number
// bar is converted to bignumber
```

fuxingloh:
/kind fix

A balancing act between BigNumber parsing and effective usage. I redesigned it to use path based mapping. It will map all within a path to the specific precision. This allow flexibility json parsing and an easy to use interface. 

For the bellow setup, only `h` and `d.f.f1` will be number, while the request get remapped to `BigNumber`.

```ts
const losslessObj = parse(`
  {
    "a": 1.2,
    "b": [1.2, 1.2],
    "c": { "c1": 1.2, "c2": [1.2] },
    "d": {
      "e": {
        "e1": 1.2,
        "e2": [1.2, 1.2],
        "e3": { "e3a": 1.2, "e3b": [1.2] }
      },
      "f": {
        "g": 1.2,
        "f1": 1.2
      }
    },
    "h": 1.2
  }
  `)
  const remapped = remap(losslessObj, {
    a: 'bignumber',
    b: 'bignumber',
    c: 'bignumber',
    d: {
      e: 'bignumber',
      f: {
        g: 'bignumber'
      }
    }
  })
```

I also changed JSON-RPC 1.0 to handle precision on result only.

Fixes #104

### DIFF
--- a/packages/jellyfish-api-jsonrpc/src/index.ts
+++ b/packages/jellyfish-api-jsonrpc/src/index.ts
@@ -8,6 +8,7 @@ import {
 import fetch from 'cross-fetch'
 import { Response } from 'cross-fetch/lib.fetch'
 import AbortController from 'abort-controller'
+import { PrecisionPath } from '@defichain/jellyfish-json'
 
 /**
  * ClientOptions for JsonRpc
@@ -58,7 +59,7 @@ export class JsonRpcClient extends ApiClient {
   /**
    * Implements JSON-RPC 1.0 specification for ApiClient
    */
-  async call<T> (method: string, params: any[], precision: Precision): Promise<T> {
+  async call<T> (method: string, params: any[], precision: Precision | PrecisionPath): Promise<T> {
     const body = JsonRpcClient.stringify(method, params)
     const response = await this.fetchTimeout(body)
     const text = await response.text()
@@ -83,8 +84,10 @@ export class JsonRpcClient extends ApiClient {
     })
   }
 
-  private static parse (text: string, precision: Precision): any {
-    const { result, error } = JellyfishJSON.parse(text, precision)
+  private static parse (text: string, precision: Precision | PrecisionPath): any {
+    const { result, error } = JellyfishJSON.parse(text, {
+      result: precision
+    })
 
     if (error !== undefined && error !== null) {
       throw new RpcApiError(error)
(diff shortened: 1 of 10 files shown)


packages/jellyfish-api-jsonrpc/src/index.ts
```diff
@@ -83,8 +84,10 @@ export class JsonRpcClient extends ApiClient {
     })
   }
 
-  private static parse (text: string, precision: Precision): any {
-    const { result, error } = JellyfishJSON.parse(text, precision)
+  private static parse (text: string, precision: Precision | PrecisionPath): any {
+    const { result, error } = JellyfishJSON.parse(text, {
+      result: precision
+    })
```
canonbrother, line 90: wow yea this is a great

canonbrother (approved):

PR #173, opened 2021-04-29
added PersistentMNRegTestContainer

fuxingloh:

/kind feature

PersistentMNRegTestContainer container is a RegTest container with MasterNode minting preconfigured.
The container configuration is persistent and can be used consistently.
If you do not stop the container, the same container can be used for all tests.
However, you must be cognizant of race conditions.

This container should not be used for finished work, it merely a dev tool to speed up test-driven development.
Once you are done with your dev work, you should swap this out for MasterNodeRegTestContainer.


canonbrother (dismissed):
Noted!!

PR #177, opened 2021-04-30
image will only try to pull if it did not already exist

fuxingloh:

/kind fix

For testing container image that are not in dockerhub, this fix will only try to pull image if an image does not exist locally.


fuxingloh:
The run order of test will create different coverage, if the docker image is pulled before running on testcontainers tests it test coverage will be lesser. There are ways to increase the coverage by setting up conditions where the image exists and do not exist. but it's beyond the testing scope.

canonbrother (approved):

PR #185, opened 2021-05-06
added defichain-testing

canonbrother:

/kind feature

To provide bundle of test fixtures which help improve efficient and effective on writing test


packages/defichain-testing/src/utils.ts
```diff
@@ -0,0 +1,236 @@
+import BigNumber from 'bignumber.js'
```
fuxingloh, line 1: I think it's a good idea to refactor and separate all these into different categories, `account.ts`, `poolpair`, `rawtxn` and ... then have them re-exported at `index.ts` this keeps the cognitive load down when we occasionally revisit this module for enhancement in the future. 

> Code climate has a inspection that limit of 250 loc per file also, going to hit soon.

packages/defichain-testing/package.json
```diff
@@ -0,0 +1,44 @@
...
+  "scripts": {
+    "build": "tsc"
+  },
+  "dependencies": {
+    "@defichain/jellyfish-network": "0.0.0",
+    "@defichain/jellyfish-crypto": "0.0.0"
+  },
+  "devDependencies": {
+    "@defichain/testcontainers": "0.0.0"
+  }
```
fuxingloh, line 43: this should be peer dependencies, peer dependencies implies it will be provided by the package user. Since the container will always be passed from the consumer, it's a good idea to put as peer deps.

packages/defichain-testing/src/utils.ts
```diff
@@ -0,0 +1,236 @@
...
+ *
+ * @param {ContainerType} container
+ * @param {number} _amount
+ * @param {UtxosToAccountOptions} [options]
+ * @param {string} [options.address]
+ * @param {Promise<void>}
+ */
+export async function utxosToAccount (
+  container: ContainerType,
+  _amount?: number,
```
fuxingloh, line 19: We can make this non-optional since it's something user have to be cognizant of when they set up the test criteria.

fuxingloh: I think we can keep certain things brief/(aka non-optional, implicit) while other explicit. If it helps with reading the testing script setup conditions might want to leave it in.

fuxingloh: At the same time if you don't see the need for us to set a certain variable like e.g. `mintable: true` (since it's always true for testing, you can keep those implicit.

fuxingloh (changes requested):
Love this, added some minors details to address for better code longevity.

/cc @jingyi2811 once merged if you got a setup that you always need, you can add into this module directly.

.github/CODEOWNERS
```diff
@@ -21,6 +21,7 @@
 /packages/jellyfish-wallet-mnemonic/            @fuxingloh
 /packages/jellyfish-wallet-whale/               @fuxingloh
 /packages/testcontainers/                       @fuxingloh
+/packages/defichain-testing/                    @fuxingloh @canonbrother
```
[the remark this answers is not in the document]
canonbrother: noted for the naming and i still think should put you at the first.. lol

packages/defichain-testing/src/utils.ts
canonbrother: yea noted i check all again

.github/CODEOWNERS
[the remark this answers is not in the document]
fuxingloh: Remember to refactor/move the directory to `packages/testing`, same for labels, governance and etc.

jingyi2811:
> Love this, added some minors details to address for better code longevity.
> 
> /cc @jingyi2811 once merged if you got a setup that you always need, you can add into this module directly.

Ok sure!

packages/testing/src/token.ts
```diff
@@ -0,0 +1,78 @@
...
+  return hashed
+}
+
+export interface MintTokensOptions {
+  address?: string
+  utxoAmount?: number
+  mintAmount?: number
+}
+
+interface CreateTokenOptions {
```
fuxingloh, line 72: I realized you didn't export everything, but should be fine anyway.

fuxingloh (approved):

PR #189, opened 2021-05-06
guess P2SH / P2PKH / P2WSH / P2WPKH type

ivan-zynesis:

/kind feature

an `Address` library to check (user) input string is a valid address type and guess the type 

Fixes #


packages/jellyfish-transaction/src/constant/Regex.ts
```diff
@@ -0,0 +1,2 @@
+export const BASE_58 = new RegExp(/[1-9A-HJ-NP-Za-km-z]/)
+export const BECH_32 = new RegExp(/[0-9ac-hj-np-z]/)
```
fuxingloh, line 2: I personally hate the use of constants, it requires unnecessary code cross-referencing when the constants are never used beyond the scope of where it's declared. And for constants that never change, it's kinda pointless to have an extra file for it.

packages/jellyfish-transaction/src/address_validator.ts
```diff
@@ -0,0 +1,183 @@
+import * as Regex from './constant/Regex'
+
+type AddressType = 'Unknown' | 'P2PKH' | 'P2SH' | 'P2WPKH' | 'P2WSH'
```
fuxingloh, line 3: enum works for this too.

fuxingloh (changes requested):

packages/jellyfish-address/src/constant/Regex.ts
```diff
@@ -0,0 +1,2 @@
+export const BASE_58 = new RegExp(/[1-9A-HJ-NP-Za-km-z]/)
+export const BECH_32 = new RegExp(/[0-9ac-hj-np-z]/)
```
fuxingloh, line 2: you could add length into your regex

ivan-zynesis: length check are done in other validator. Trying to make the logic should more or less guess the `type` when input half way.

fuxingloh (changes requested):
requesting some changes. I haven't reviewed the logic nor the test yet, will review after the refactor.

packages/jellyfish-address/src/index.ts
```diff
@@ -0,0 +1,430 @@
...
+import { bech32 } from 'bech32'
+import { SHA256 } from '@defichain/jellyfish-crypto'
+import { getNetwork, Network, NetworkName } from '@defichain/jellyfish-network'
+import { Script } from '@defichain/jellyfish-transaction'
+import { OP_CODES, OP_PUSHDATA } from '@defichain/jellyfish-transaction/src/script'
+
+export type AddressType = 'Unknown' | 'P2PKH' | 'P2SH' | 'P2WPKH' | 'P2WSH'
+
+type Validator = () => boolean
+
```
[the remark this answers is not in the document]
ivan-zynesis: Done

packages/jellyfish-network/src/index.ts
```diff
@@ -1,13 +1,9 @@
...
-/**
- * Networks available in DeFi Blockchain.
- */
-export type NetworkName = 'mainnet' | 'testnet' | 'regtest'
-
 /**
  * Network specific DeFi configuration.
  * They can be found in DeFiCh/ain project in file chainparams.cpp, under base58Prefixes
  */
+export type NetworkName = 'mainnet' | 'testnet' | 'regtest'
```
fuxingloh, line 5: to revert changes done in this file

package.json
```diff
@@ -55,4 +55,4 @@
       "limit": "1000 ms"
     }
   ]
-}
+}
```
fuxingloh, line 58: this as well

packages/jellyfish-address/src/index.ts
```diff
@@ -0,0 +1,82 @@
...
+      highestCount = val.validatorPassed
+    }
+  })
+  return (possible.get(highestKey) as T)
+}
+
+export default {
+  guess,
+  from
+}
```
fuxingloh, line 82: don't use default export, we trying to stay away from CJS coventions to use ESM

fuxingloh (changes requested):

fuxingloh:
rebase/merge main branch, I will check in abit

packages/jellyfish-address/src/index.ts
```diff
@@ -0,0 +1,82 @@
+
+import { getNetwork, NetworkName } from '@defichain/jellyfish-network'
+import { Address, AddressType, UnknownTypeAddress } from './Address'
+import { Base58Address } from './Base58Address'
+import { Bech32Address } from './Bech32Address'
+import { P2PKH } from './P2PKH'
+import { P2SH } from './P2SH'
+import { P2WSH } from './P2WSH'
+import { P2WPKH } from './P2WPKH'
```
fuxingloh, line 9: To lowercase all file name and use `_` separator. As this is the standardized approach for this repo.

https://github.com/DeFiCh/jellyfish/blob/main/CONTRIBUTING.md#use-underscores-not-dashes-in-filenames

fuxingloh (commented):
https://github.com/DeFiCh/jellyfish/blob/main/CONTRIBUTING.md

I am going to refactor myself instead based on the contributing guidelines.

fuxingloh (approved):
```diff
@@           Coverage Diff            @@
##             main     #189    +/-   ##
========================================
  Coverage   96.21%   96.22%            
========================================
  Files          66       74     +8     
  Lines        1718     1958   +240     
  Branches      235      264    +29     
========================================
+ Hits         1653     1884   +231     
- Misses         65       74     +9     
```

Mixed feelings about this PR. I would expect around `<15` code logical branches for this ticket but there are `+29` which is 10% more than the current combined mono repo code base. 

https://github.com/DeFiCh/jellyfish/pull/237 has only `+7` for comparison and it has a much more complex implementation design. Clearly, there is much to optimize, the more code there is, the more there is to maintain. Less is more.

I would take it as it is now as it functionally works, but I think I will revisit and refactor this implementation in the future. I will create a ticket to refactor this at a future date.

PR #200, opened 2021-05-10
Add OP_TRUE, OP_1 to OP_16

jingyi2811:
/kind feature

Add OP_TRUE, OP_1 to OP_16 opcodes

### DIFF
--- a/packages/jellyfish-transaction/src/script/mapping.ts
+++ b/packages/jellyfish-transaction/src/script/mapping.ts
@@ -1,25 +1,4 @@
-import {
-  OP_0,
-  OP_FALSE,
-  OP_1NEGATE,
-  OP_1,
-  OP_TRUE,
-  OP_2,
-  OP_3,
-  OP_4,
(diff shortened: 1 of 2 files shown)


packages/jellyfish-transaction/src/script/mapping.ts
```diff
@@ -1,4 +1,25 @@
...
+  OP_8,
+  OP_9,
+  OP_10,
+  OP_11,
+  OP_12,
+  OP_13,
+  OP_14,
+  OP_15,
+  OP_16
+} from './constants'
```
fuxingloh, line 22: for situations like this you might want to write it like this:

```ts
import * as constants from './constants'
```

```ts
OP_1: new constants.OP_1(),
OP_TRUE: new constants.OP_TRUE()
```

I feel like it makes the code cleaner, your call.

fuxingloh (dismissed):

fuxingloh (approved):

PR #203, opened 2021-05-10
Add OP_RIPEMD160 opcode

jingyi2811:
/kind feature

Add OP_RIPEMD160 opcode


fuxingloh (changes requested):
> I do not add any unit test to index.test.ts
> 
> The reason is I believe the unit test which was written for OP_HASH160 has included the testing of OP_RIPEMD160 as OP_HASH160 =The input is hashed twice: first with SHA-256 and then with RIPEMD-160.
> 
> Kindly advise if my decision was wrong

You need `index.test.ts` as they are different features. 
These are instructions for stack scripting, there are no magic to this. It's passed to defid to process. JS doesn't do anything with it. It merely parses it for client side rendering.

jingyi2811:
Last commit message is wrong. It should be Add OP_RIPEMD160 unit test.

Not too sure if adding this unit test is enough to close this PR. Kindly advise if you think this is insufficient.

fuxingloh (approved):

fuxingloh:
@jingyi2811 to fix conflicting files

canonbrother (dismissed):

packages/jellyfish-transaction/__tests__/script/mapping.test.ts
```diff
@@ -1,4 +1,5 @@
 import * as constants from '../../src/script'
+import { OP_CODES, OP_RIPEMD160 } from '../../src/script'
```
fuxingloh, line 2: I didn't catch this previously. do this instead:

```ts
import * as script from '../../src/script'
```

fuxingloh (changes requested):

packages/jellyfish-network/__tests__/index.test.ts
```diff
@@ -1,4 +1,4 @@
-import { Network, MainNet, RegTest, TestNet, getNetwork } from '../src'
+import { getNetwork, MainNet, Network, RegTest, TestNet } from '../src'
```
fuxingloh, line 1: revert this

fuxingloh (approved):

PR #229, opened 2021-05-16
Add new getMiningInfo for multiple masternode support

benzumbrunn:
/kind feature

`getMintingInfo` is deprecated since adding multiple masternode support from within one server:
https://github.com/DeFiCh/ain/blob/df796cb7d6dbbe0ed7d051e79db0f36630f77afb/src/rpc/mining.cpp#L220

This PR adds `getMiningInfo` which is the implementation to support multiple mining masternodes:
https://github.com/DeFiCh/ain/blob/df796cb7d6dbbe0ed7d051e79db0f36630f77afb/src/rpc/mining.cpp#L276

#### Additional comments?:
* Added docs with a deprecated comment for the old one, but please advise if this is the correct way
* `getMintingInfo` is used as an example in some places, let me know if this PR should replace the examples as well
* Let me know if the test data should include a second node now, since there's an array to test now

### DIFF
--- a/packages/jellyfish-api-core/__tests__/category/mining.test.ts
+++ b/packages/jellyfish-api-core/__tests__/category/mining.test.ts
@@ -80,4 +80,34 @@ describe('masternode', () => {
       expect(result).toBeGreaterThan(0)
     })
   })
+
+  it('should getMiningInfo', async () => {
+    await waitForExpect(async () => {
+      const info = await client.mining.getMiningInfo()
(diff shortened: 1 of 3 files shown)


fuxingloh (approved):
Thanks, I made some minor changes to how the docs are formatted. GitHub Actions is having a Major outage today, I will set this PR to auto-merge when all the checks passes.

> https://www.githubstatus.com/

canonbrother (approved):

PR #242, opened 2021-05-17
refactor DeFiAddress to export const

fuxingloh:

/kind refactor

Avoid ES6 default export at all cost, technical debt minter.


fuxingloh:
going to merge now to generate package.

PR #248, opened 2021-05-18
fix custom tx vout not deducted

ivan-zynesis:

/kind fix

total vin, total vout not tally for built transaction

Fixes #


fuxingloh (approved):
thanks :)

Issue #252, opened 2021-05-18
Filter `txtype` Parameter of `listAccountHistory()` doesn´t seem to filter AccountHistory


gamsterx:
#### What happened:
I want to get a filtered list of transactions via `listAccountHistory`. I used the `txtype` options parameter which didn´t seem to have any effect. 

I tried different strings like `PoolSwap` or `Commission` with the same unfiltered result.

As this project is currently in heavy development the docs are limited at this point in time. I managed to find the following info:
`@param {string} [options.txtype] Filter by transaction type, supported letter from 'CRTMNnpuslrUbBG`

I have no clue what the letters suppose to mean. Can somebody explain?

An example from the tests shows a `txtype: 'M'` which in my case also didn´t filter the transactions.

#### What you expected to happen:

I expect to get only history objects of the `txtype` `Rewards` or `Commission` or `PoolSwap` or `...`.

#### How to reproduce it (as minimally and precisely as possible):

` "@defichain/jellyfish": "^0.3.0"`

```ts
import { Client } from "@defichain/jellyfish";
import { AccountHistoryOptions } from "@defichain/jellyfish-api-core/dist/category/account";

const client = new Client(
    "http://USER:PW@localhost:8555"
  );
  
  const options: AccountHistoryOptions = {
    txtype: "",
  };

  const accountHistories = await client.account.listAccountHistory(
    "mine",
    options
  );
```

fuxingloh:
Those characters are the custom DeFi transactions. See https://github.com/DeFiCh/jellyfish/issues/108 for more info.

Thanks for creating this issue, appreciate it! I will assign this to be worked on, our priority is still light wallet. But this is something valid that we need to address.

/assign @jingyi2811 we need to add `export enum TxType {}` and update the documentations on `JSDoc` and `website/`
/area jellyfish-api-core
/triage accepted

siradji:
@fuxingloh , this issue was address and fixed in  `Add account.historyCount RPC` #265.

fuxingloh:
> @fuxingloh , this issue was addressed and fixed in `Add account.historyCount RPC` #265.

That RPC is for `historyCount`, this is for `listAccountHistory`. 
https://github.com/DeFiCh/jellyfish/blob/9d311db3c3f905fc05fdecdacc0ab96176c291f0/packages/jellyfish-api-core/src/category/account.ts#L110-L117

PR #254, opened 2021-05-19
added estimateSmartFee rpc

thedoublejay:

/kind feature

Implemented RPC for #48 
1. mining.estimatesmartfee

Fixes #


packages/jellyfish-api-core/src/category/mining.ts
```diff
@@ -86,3 +98,12 @@ export interface MasternodeInfo {
   mintedblocks?: number
   lastblockcreationattempt?: string
 }
+
+/**
+ * @description return type of rpc `estimatesmartfee`
+ */
```
fuxingloh, line 104: don't need to specific this, everything in this package does that.

packages/jellyfish-api-core/src/category/mining.ts
```diff
@@ -37,6 +37,16 @@ export class Mining {
...
     return await this.client.call('getmininginfo', [], 'number')
   }
+
+  /**
+   *
+   * @param {number} confirmationTarget in blocks (1 - 1008)
+   * @param {EstimateMode} [estimateMode='CONSERVATIVE'] estimateMode of fees.
+   * @returns {Promise<SmartFeeEstimation>}
+   */
+  async estimateSmartFee (confirmationTarget: number, estimateMode: wallet.Mode = wallet.Mode.CONSERVATIVE): Promise<SmartFeeEstimation> {
```
fuxingloh, line 47: don't import `wallet` even though it's applicable. This is to prevent code smell where everything starts to depend on each other, making it hard to migrate around. For abstraction, yes, but for implementation no. `wallet` is an implementation.

fuxingloh (changes requested):

fuxingloh:
don't use semantic convention either, just stick to PR purpose

packages/jellyfish-api-core/src/category/mining.ts
thedoublejay: ok will update it

packages/jellyfish-api-core/__tests__/category/mining.test.ts
```diff
@@ -40,6 +49,8 @@ describe('masternode', () => {
   beforeAll(async () => {
     await container.start()
     await container.waitForReady()
+    await container.waitForWalletCoinbaseMaturity()
+    await container.waitForWalletBalanceGTE(100)
```
fuxingloh, line 53: you can revert this change, you don't need it, you have your own setup below

packages/jellyfish-api-core/src/category/mining.ts
```diff
@@ -86,3 +96,11 @@ export interface MasternodeInfo {
...
   lastblockcreationattempt?: string
 }
+
+export interface SmartFeeEstimation {
+  feerate?: number
+  errors?: string[]
+  blocks: number
+}
+
+type EstimateMode = 'UNSET' | 'ECONOMICAL' | 'CONSERVATIVE'
```
fuxingloh, line 106: Currently, everyone put it above, we can follow that also, should use enum also. Enum we can add docs. plain types we can't

You can follow this but have it for yourself: 

https://github.com/DeFiCh/jellyfish/blob/ece0fdc6e24e2f9601fc4dfa474a9b6c924d507d/packages/jellyfish-api-core/src/category/wallet.ts#L3-L7

fuxingloh (changes requested):

packages/jellyfish-api-core/src/category/mining.ts
thedoublejay: Ok sure. I just remembered we had a discussion before about enums so I used types.

canonbrother (commented):
other than fuxing's request.. overall look nice for me

packages/jellyfish-api-core/src/category/mining.ts
fuxingloh: I think you convinced me not to use type for everything. I only use `type` selectively now.

fuxingloh: Also because of this, enum has JSDoc support but for type there isn't: https://github.com/DeFiCh/jellyfish/issues/252

website/docs/jellyfish/api/mining.md
```diff
@@ -91,3 +91,25 @@ interface MasternodeInfo {
...
 }
 ```
+
+## estimateSmartFee
+
+Estimates the approximate fee per kilobyte needed for a transaction
+
+```ts title="client.mining.estimateSmartFee()"
+interface mining {
+  estimateSmartFee (confirmationTarget: number, estimateMode: EstimateMode = 'CONSERVATIVE'): Promise<SmartFeeEstimation>
```
fuxingloh, line 101:
```suggestion
  estimateSmartFee (confirmationTarget: number, estimateMode: EstimateMode = EstimateMode.CONSERVATIVE): Promise<SmartFeeEstimation>
```

packages/jellyfish-api-core/src/category/mining.ts
```diff
@@ -37,6 +43,16 @@ export class Mining {
   async getMiningInfo (): Promise<MiningInfo> {
     return await this.client.call('getmininginfo', [], 'number')
   }
+
+  /**
+   *
+   * @param {number} confirmationTarget in blocks (1 - 1008)
+   * @param {EstimateMode} [estimateMode='CONSERVATIVE'] estimateMode of fees.
```
fuxingloh, line 50:
```suggestion
   * @param {EstimateMode} [estimateMode=EstimateMode.CONSERVATIVE] estimateMode of fees.
```

fuxingloh (changes requested):

packages/jellyfish-api-core/src/index.ts
```diff
@@ -78,10 +78,10 @@ export class ClientApiError extends ApiError {
  * API RPC error, from upstream.
  */
 export class RpcApiError extends ApiError {
-  public readonly payload: { code: number, message: string }
+  public readonly payload: { code: number, message: string, rpcMethod?: string }
 
-  constructor (error: { code: number, message: string }) {
-    super(`RpcApiError: '${error.message}', code: ${error.code}`)
+  constructor (error: { code: number, message: string, rpcMethod?: string }) {
```
fuxingloh, line 83: Remove the optional aspect for `rpcMethod?: string` and add it to https://github.com/DeFiCh/jellyfish/blob/82b6692b48492bf109f95bbe3767f03bf32e0d7a/packages/jellyfish-api-jsonrpc/src/index.ts#L92-L94
as well

fuxingloh (changes requested):

packages/jellyfish-api-core/src/index.ts
thedoublejay, line 83: @fuxingloh  @canonbrother I have added some changes on the `RpcApiError` class to make the rpcmethod visible on error logs.
image

Tests are running well locally but throwing errors on CI. Let me know if it's ok to merge, otherwise I can remove it when it's ready for review.

fuxingloh: I think varying CI conditions and CPU differences causes inconsistency in the test. 

Likely because the fees required to send increased hence the insufficient funds. To keep it consistent you can subtract fees from the amount. And with walletBalanceGTE() it will always have sufficient funds. By doing this you might need to increase the amount sent by x10.

https://github.com/DeFiCh/jellyfish/blob/82b6692b48492bf109f95bbe3767f03bf32e0d7a/packages/jellyfish-api-core/src/category/wallet.ts#L259

thedoublejay: Thanks! will try that

fuxingloh (approved):

canonbrother (approved):

Issue #108, opened 2021-04-12
DeFi CustomTx scripting


fuxingloh:
> This tracks all the DeFi scripting (AKA CustomTx) that is to be implemented in jellyfish-transaction.

```
packages/jellyfish-transaction/src/script/defi/*
```

All CustomTX has a starting signature of `DfTx` in utf8 or `0x44665478`. It is as it is, don't need to convert from BE to LE.

They are structured as a 
1. `VarUInt` + [`OP_RETURN` + `DfTx` + `TYPES[CxTxType]` + ...]
1. `VarUInt` + `6a` + `44665478` + Type + ...

## CustomTx
> It's a char, UTF8 -> HEX

- [x] createMasternode: `'C'` @ivan-zynesis #283
- [x] resignMasternode: `'R'` @ivan-zynesis #283
- [x] createToken: `'T'` @canonbrother #208
- [x] mintToken: `'M'` @canonbrother #170
- [x] updateToken: `'N'` @canonbrother #300
- [x] updateTokenAny: `'n'` @canonbrother #300
- [x] createPoolPair: `'p'` @canonbrother #233
- [x] updatePoolPair: `'u'` @canonbrother #314 
- [x] poolSwap: `'s'` @fuxingloh #143
- [x] addPoolLiquidity: `'l'` @fuxingloh #145
- [x] removePoolLiquidity: `'r'` @fuxingloh  #146
- [x] utxosToAccount: `'U'` @ivan-zynesis #175 
- [x] accountToUtxos: `'b'` @ivan-zynesis #175 
- [x] accountToAccount: `'B'` @ivan-zynesis #175 
- [x] setGovVariable: `'G'` @ivan-zynesis #298
- [x] anyAccountsToAccounts: `'a'` @ivan-zynesis #181
- [x] AutoAuthPrep: `'A'`, @ivan-zynesis #279

fuxingloh:
https://github.com/DeFiCh/ain/pull/287/files

Oracles and AuthAuth

```c++
    AutoAuthPrep       = 'A',
    // oracles
    AppointOracle         = 200,
    RemoveOracleAppoint   = 201,
    UpdateOracleAppoint   = 202,
    SetOracleData         = 203,
```

fuxingloh:
Check out #146 it's a good example on how to create DfTx

ivan-zynesis:
Start working on.
- utxosToAccount: 'U'
- accountToUtxos: 'b'
[...]

PR #265, opened 2021-05-20
Add `account.historyCount` RPC

siradji:

/kind feature

RPC implementation #48

Fixes #


packages/jellyfish-api-core/__tests__/category/account.test.ts
```diff
@@ -516,4 +516,60 @@ describe('masternode', () => {
...
+
+  describe('accountHistoryCount', () => {
+    it('should get accountHistoryCount', async () => {
+      await waitForExpect(async () => {
+        const accountHistoryCount = await client.account.accountHistoryCount()
+
+        expect(typeof accountHistoryCount).toBe('number')
+        expect(accountHistoryCount).toBeGreaterThan(0)
+      })
+    })
```
canonbrother, line 528: alignment

canonbrother (changes requested):
overall looks great
just the alignment on the test

packages/jellyfish-api-core/__tests__/category/account.test.ts
```diff
@@ -516,4 +516,60 @@ describe('masternode', () => {
       expect(data.length).toBe(64)
     })
   })
+
+  describe('accountHistoryCount', () => {
+    it('should get accountHistoryCount', async () => {
+      await waitForExpect(async () => {
+        const accountHistoryCount = await client.account.accountHistoryCount()
```
jingyi2811, line 523: The variable name accountHistoryCount should be count like below

siradji: Changed @jingyi2811

packages/jellyfish-api-core/__tests__/category/account.test.ts
```diff
@@ -516,4 +516,60 @@ describe('masternode', () => {
...
+        expect(count).toBeGreaterThan(0)
+      })
+    })
+
+    it('should get accountHistoryCount with owner as all', async () => {
+      await waitForExpect(async () => {
+        const count = await client.account.accountHistoryCount('all')
+
+        expect(typeof count).toBe('number')
+        expect(count).toBeGreaterThan(0)
```
jingyi2811, line 535: I feel like this should be toBeGreaterThanOrEqual. Don't change first. Let see what others think.

packages/jellyfish-api-core/__tests__/category/account.test.ts
siradji: @canonbrother , the alignment on line 527 and 528? 

image

canonbrother: gosh.. lol.. sorry... you're right... i gonna rest

canonbrother (dismissed):

packages/jellyfish-api-core/src/category/account.ts
```diff
@@ -235,6 +235,23 @@ export class Account {
...
+   *
+   * @param {OwnerType} [owner='mine'] single account ID (CScript or address) or reserved words 'mine' to list history count for all owned accounts or 'all' to list whole DB
+   * @param {AccountHistoryCountOptions} [options]
+   * @param {boolean} [options.no_rewards]  Filter out rewards
+   * @param {string} [options.token]  Filter by token
+   * @param {string} [options.txtype]  Filter by transaction type, supported letter from 'CRTMNnpuslrUbBG'
+   * @return {Promise<number>}  Count of account history
+   */
+  async accountHistoryCount (
+    owner: OwnerType = 'mine',
```
thedoublejay, line 250: I think we can replace `OwnerType` with an enum. Currently it's only 'mine' or 'all' but OwnerType is accepting 'mine', 'all' or any string. Maybe we can restrict it to an enum for better strict checking.

packages/jellyfish-api-core/__tests__/category/account.test.ts
```diff
@@ -516,4 +516,60 @@ describe('masternode', () => {
...
+      })
+    })
+
+    it('should get accountHistoryCount with no_rewards option', async () => {
```
thedoublejay, line 539: I think one way we can catch regression issues from node side here is to check if `no_rewards` param does it job by reducing the total count of account history. We can add an additional test case here to call an initial accountHistoryCount with rewards, then call accountHistoryCount with no_rewards. 

Should have less result with no_rewards. The only part we need to add is to make sure that account history has rewards. Maybe by creating a liquidity pool first. Just a suggestion.

thedoublejay (changes requested):

packages/jellyfish-api-core/src/category/account.ts
siradji: I saw there was an owner type already in place that i why i used that. Let me change it to an enum as you suggested.

siradji: @thedoublejay, from ain/rcp_account.cpp, `accountHistory` ownerType can accept the reserved words , `mine` and `all ` or a string  `Single account ID (CScript or address) or reserved words`. So using an enum is not ideal. What do you think?

thedoublejay: @siradji hmm yeah good point. In this case we can wrap it in an interface. Something like OwnerType | string. Then at least on other parts of code, we can use OwnerType.MINE.

siradji: Good point. Lemme do that.

siradji: @canonbrother, changing `Ownertype` to an `enum`  will after `listAccount` implementation that you worked on. How should i resolve this?

canonbrother: yea its good to go.. refactor it  =)

packages/jellyfish-api-core/__tests__/category/account.test.ts
thedoublejay: I'll let @fuxingloh decide on this. Otherwise you can resolve and merge PR if it's ok @siradji

canonbrother (dismissed):

packages/jellyfish-api-core/__tests__/category/account.test.ts
siradji: I would love to try and implement your suggestion but i don't know and i have a lot of other implementations to work on. But if @fuxingloh wants it integrated, i will do it.

fuxingloh: yeah, we should definetly do it. 

it might be complex going into the defi system but it's good to start learning now too.

packages/jellyfish-api-core/src/category/account.ts
```diff
@@ -235,6 +237,23 @@ export class Account {
...
   }
+
+  /**
+   * Returns count of account history
+   *
+   * @param {OwnerType | string} [owner= OwnerType.MINE] single account ID (CScript or address) or reserved words 'mine' to list history count for all owned accounts or 'all' to list whole DB
+   * @param {AccountHistoryCountOptions} [options]
+   * @param {boolean} [options.no_rewards]  Filter out rewards
+   * @param {string} [options.token]  Filter by token
+   * @param {string} [options.txtype]  Filter by transaction type, supported letter from 'CRTMNnpuslrUbBG'
```
fuxingloh, line 248: Since you are working on this, we need to refactor this to follow how OwnerType is done now. I have assigned you https://github.com/DeFiCh/jellyfish/issues/252#issuecomment-843766560 you can do it in this PR.

Instead of `txtype?: string` to `txtype?: TxType | string`

packages/jellyfish-api-core/src/category/account.ts
```diff
@@ -235,6 +237,23 @@ export class Account {
...
+   * @return {Promise<number>}  Count of account history
+   */
+  async accountHistoryCount (
```
fuxingloh, line 251:
```suggestion
  async historyCount (
```

Everything in this class is account, `account.accountHistoryCount` double semantic. shorten to `account.historyCount` for brevity.

packages/jellyfish-api-core/__tests__/category/account.test.ts
fuxingloh: Yeah, I am also curious why it did not fail. Likely because the MasterNode container create a MasterNode and that is an account history. Causing it to not always be 0. @canonbrother you happen to know the reason?

@siradji what is the actual number, can we do a deterministic assertion for this?

fuxingloh (changes requested):

packages/jellyfish-api-core/__tests__/category/account.test.ts
canonbrother: `because the MasterNode container create a MasterNode and that is an account history`
nope, its because of the `waitForWalletCoinbaseMaturity()`

siradji: @fuxingloh, i am getting 136.

packages/jellyfish-api-core/__tests__/category/account.test.ts
siradji: I will definitely do that.

packages/jellyfish-api-core/__tests__/category/account.test.ts
siradji: Anyways, i have changed the assertion to `toBeGreaterThanOrEqual`.

packages/jellyfish-api-core/src/category/account.ts
siradji: Implemented `TxType enum` referencing #108.

website/docs/jellyfish/api/account.md
```diff
@@ -162,3 +165,46 @@ interface UtxosToAccountUTXO {
...
+   CREATE_MASTERNODE = 'C',
+   RESIGN_MASTERNODE = 'R',
+   CREATE_TOKEN = 'T',
+   UPDATE_TOKEN = 'N',
+   UPDATE_TOKEN_ANY = 'n',
+   CREATE_POOL_PAIR = 'p',
+   UPDATE_POOL_PAIR = 'u',
+   SET_GOV_VARIABLE = 'G',
+   AUTO_AUTH_PREP = 'A'
+}
```
fuxingloh, line 204:
```suggestion
enum TxType {
  MINT_TOKEN = 'M',
  POOL_SWAP = 's',
  ADD_POOL_LIQUIDITY = 'l',
  REMOVE_POOL_LIQUIDITY = 'r',
  UTXOS_TO_ACCOUNT = 'U',
  ACCOUNT_TO_UTXOS = 'b',
  ACCOUNT_TO_ACCOUNT = 'B',
  ANY_ACCOUNTS_TO_ACCOUNTS = 'a',
  CREATE_MASTERNODE = 'C',
  RESIGN_MASTERNODE = 'R',
  CREATE_TOKEN = 'T',
  UPDATE_TOKEN = 'N',
  UPDATE_TOKEN_ANY = 'n',
  CREATE_POOL_PAIR = 'p',
  UPDATE_POOL_PAIR = 'u',
  SET_GOV_VARIABLE = 'G',
  AUTO_AUTH_PREP = 'A'
}
```

website/docs/jellyfish/api/account.md
```diff
@@ -162,3 +165,46 @@ interface UtxosToAccountUTXO {
...
+interface  AccountHistoryCountOptions {
+  token?: string
+  txtype?:TxType | string
+  no_rewards?: boolean
+}
```
fuxingloh, line 209:
```suggestion

interface AccountHistoryCountOptions {
  token?: string
  txtype?: TxType | string
  no_rewards?: boolean
}
```

packages/jellyfish-api-core/src/category/account.ts
```diff
@@ -235,6 +256,23 @@ export class Account {
...
+  /**
+   * Returns count of account history
+   *
+   * @param {OwnerType | string} [owner= OwnerType.MINE] single account ID (CScript or address) or reserved words 'mine' to list history count for all owned accounts or 'all' to list whole DB
+   * @param {AccountHistoryCountOptions} [options]
+   * @param {boolean} [options.no_rewards]  Filter out rewards
+   * @param {string} [options.token]  Filter by token
+   * @param {TxType | string} [options.txtype]  Filter by transaction type, supported letter from 'CRTMNnpuslrUbBG'
+   * @return {Promise<number>}  Count of account history
+   */
```
fuxingloh, line 269:
```suggestion
  /**
   * Returns count of account history
   *
   * @param {OwnerType | string} [owner=OwnerType.MINE] single account ID (CScript or address) or reserved words 'mine' to list history count for all owned accounts or 'all' to list whole DB
   * @param {AccountHistoryCountOptions} [options]
   * @param {boolean} [options.no_rewards] Filter out rewards
   * @param {string} [options.token] Filter by token
   * @param {TxType | string} [options.txtype] Filter by transaction type, supported letter from 'CRTMNnpuslrUbBG'
   * @return {Promise<number>} count of account history
   */
```

packages/jellyfish-api-core/src/category/account.ts
```diff
@@ -6,8 +6,29 @@ import { ApiClient } from '../.'
  * - 'mine' to list history for all owned accounts or
  * - 'all' to list the whole DB
  */
-type OwnerType = 'mine' | 'all' | string
-
+enum OwnerType {
```
fuxingloh, line 9:
```suggestion
export enum OwnerType {
```

packages/jellyfish-api-core/src/category/account.ts
```diff
@@ -6,8 +6,29 @@ import { ApiClient } from '../.'
...
+  MINE = 'mine',
+  ALL = 'all'
+}
+export enum TxType {
+  MINT_TOKEN ='M',
+  POOL_SWAP = 's',
+  ADD_POOL_LIQUIDITY = 'l',
+  REMOVE_POOL_LIQUIDITY = 'r',
+  UTXOS_TO_ACCOUNT = 'U',
+  ACCOUNT_TO_UTOXS = 'b',
```
fuxingloh, line 19:
```suggestion
  ACCOUNT_TO_UTXOS = 'b',
```

packages/jellyfish-api-core/src/category/account.ts
```diff
@@ -6,8 +6,29 @@ import { ApiClient } from '../.'
  * - 'mine' to list history for all owned accounts or
  * - 'all' to list the whole DB
  */
-type OwnerType = 'mine' | 'all' | string
-
+enum OwnerType {
+  MINE = 'mine',
+  ALL = 'all'
+}
+export enum TxType {
```
fuxingloh, line 13:
```suggestion

export enum TxType {
```

fuxingloh (changes requested):
a few changes to take note of

packages/jellyfish-api-core/__tests__/category/account.test.ts
```diff
@@ -516,4 +516,60 @@ describe('masternode', () => {
...
+      await waitForExpect(async () => {
+        const options: AccountHistoryCountOptions = {
+          token: 'DBTC'
+        }
+        const count = await client.account.historyCount('mine', options)
+
+        expect(typeof count).toBe('number')
+        expect(count).toBeGreaterThanOrEqual(0)
+      })
+    })
```
canonbrother, line 561: this test you can actually test 'DBTC', 'DETH' and 'DFI' which will output different count, other than `toBeGreaterThanOrEqual`, you can expect the result in exact count. 
you can refer to the fixture `setup` above

packages/jellyfish-api-core/__tests__/category/account.test.ts
```diff
@@ -516,4 +516,60 @@ describe('masternode', () => {
...
+
+    it('should get accountHistory with txtype option', async () => {
+      await waitForExpect(async () => {
+        const options: AccountHistoryCountOptions = {
+          txtype: TxType.MINT_TOKEN
+        }
```
canonbrother, line 567: this is same as well, you can make use of `tytype` to expect the diff exact count

packages/jellyfish-api-core/__tests__/category/account.test.ts
siradji: Did not know that. let me try doing that.

packages/jellyfish-api-core/__tests__/category/account.test.ts
siradji: Sure. Will write and additional test case for that.

packages/jellyfish-api-core/__tests__/category/account.test.ts
```diff
@@ -516,4 +516,81 @@ describe('masternode', () => {
...
+        const options2: AccountHistoryCountOptions = {
+          token: 'DETH'
+        }
+        const countWithDBTC = await client.account.historyCount('mine', options1)
+        const countWithDETH = await client.account.historyCount('mine', options2)
+
+        expect(typeof countWithDBTC).toBe('number')
+        expect(typeof countWithDETH).toBe('number')
+        expect(countWithDETH === countWithDBTC).toStrictEqual(false)
```
canonbrother, line 564: hmm... this is not my expect actually..
my previous request was actually requesting a meaningful test.. you should ady known the `count` as you can refer to the fixture set up above, so the test should be telling a story how the `accounthistorycount` work
you can straight away assert the exact number since its fixed

packages/jellyfish-api-core/__tests__/category/account.test.ts
canonbrother: this part is actually same principle with https://github.com/DeFiCh/jellyfish/pull/265/files#r638661038 comment
test with diff txtype and assert the exact number

packages/jellyfish-api-core/__tests__/category/account.test.ts
siradji: @canonbrother , you mean something like `expect(count).toStrictEqual(10)` ?

canonbrother: yea right

fuxingloh (approved):

jingyi2811 (approved):

PR #275, opened 2021-05-21
Added test failure for accounttoaccount

monstrobishi:

/kind feature

Adds a failing test case for txn-builder

Fixes #


fuxingloh:
Closed as this scenario is not relevant to the current implementation, `prevout.provider mock` implementation design is done in bare minimum sense and is bound to create that error.

PR #286, opened 2021-05-24
Add `wallet.dumpPrivKey` RPC

siradji:

/kind feature

 RPC implementation #48 

Fixes #

### DIFF
--- a/packages/jellyfish-api-core/__tests__/category/wallet.test.ts
+++ b/packages/jellyfish-api-core/__tests__/category/wallet.test.ts
@@ -222,23 +222,6 @@ describe('non masternode', () => {
       })
     })
   })
-  describe('dumpPrivKey', () => {
-    it('should reveal private key of given address', async () => {
-      await waitForExpect(async () => {
-        const address = await client.wallet.getNewAddress()
-        const privateKey = await client.wallet.dumpPrivKey(address)
-
-        expect(typeof privateKey).toStrictEqual('string')
-      })
-    })
-
-    it('should throw and error when invalid DFI address is provided', async () => {
-      await waitForExpect(async () => {
-        const invalidAddress = 'invalidAddress'
-        await expect(client.wallet.dumpPrivKey(invalidAddress)).rejects.toThrow('Invalid Defi address')
-      })
-    })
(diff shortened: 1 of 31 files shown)


packages/jellyfish-api-core/__tests__/category/wallet.test.ts
```diff
@@ -631,4 +631,21 @@ describe('masternode', () => {
...
     })
   })
+
+  describe('dumpPrivKey', () => {
+    it('should reveal private key of given address', async () => {
+      await waitForExpect(async () => {
+        const address = await client.wallet.getNewAddress()
+        const privateKey = await client.wallet.dumpPrivKey(address)
+
+        expect(typeof privateKey).toStrictEqual('string')
```
canonbrother, line 641: its recommended to expect a exact output
refer https://github.com/DeFiCh/jellyfish/blob/35a5f074c1d1789f40733360baedf5aee2aa7534/packages/testcontainers/src/testkeys.ts

canonbrother (changes requested):
need add the `dumpPrivKey` on *non masternode* test as well

packages/jellyfish-api-core/__tests__/category/wallet.test.ts
siradji: Those test keys are different from what i am getting. Is that normal?

siradji:
### Behavior 
If i run test on the `dumprivkey` suite individually, it works well and pass but fails when i run test on  the`masternode` suite  or the whole test file. 


### fails at  

```

  describe('dumpPrivKey', () => {
    it('should reveal private key of given address', async () => {
      await waitForExpect(async () => {
        const privateKey = await client.wallet.dumpPrivKey(address)

        expect(typeof privateKey).toStrictEqual('string')
      })
    })

    it('should throw and error when invalid DFI address is provided', async () => {
      await waitForExpect(async () => {
        const invalidAddress = 'invalidAddress'

        await expect(client.wallet.dumpPrivKey(invalidAddress)).rejects.toThrow('Invalid Defi address')
      })
    })
  })
```

### Error that i get

image

PR #290, opened 2021-05-25
Add mintTokens rpc

jingyi2811:
#### What kind of PR is this?:

/kind feature

#### What this PR does / why we need it:
Add mintTokens rpc

#### Which issue(s) does this PR fixes?:

Fixes part of #48

PR #323, opened 2021-05-31
`sendTokensToAddress` RPC

siradji:
What kind of PR is this?:

/kind feature

Add sendTokensToAddress RPC

Fixes part of #48
Additional comments?


siradji:
### Behavior
Test fails when source address is provided but passes when an empty object is passed. 

### Fails at

```javascript
  it('should create a transaction with source address provided', async () => {
    const to = await client.wallet.getNewAddress()
    const from = await client.wallet.getNewAddress()
    const transactionHex = await client.account.sendTokensToAddress({ [from]: '2@DFI' }, { [to]: '10@DBTC' })

    expect(typeof transactionHex).toStrictEqual('string')
  })
```

### Error 
image

website/docs/jellyfish/api/account.md
```diff
@@ -237,3 +237,26 @@ interface AccountHistoryCountOptions {
...
+type AccountRegexType = `${string}@${string}`
+
+enum SelectionModeType {
+  PIE = 'pie',
+  CRUMBS = 'crumbs',
+  FORWARD = 'forward'
+}
+
+interface Address {
+  [key: string]: AccountRegexType | AccountRegexType[]
```
jingyi2811, line 260: Should accept array only

packages/jellyfish-api-core/src/category/account.ts
```diff
@@ -294,6 +300,18 @@ export class Account {
...
+
+  /**
+   * Creates a transfer transaction from your accounts balances.
+   *
+   * @param {Address} from The source defi address is the key, the value is amount in amount amount@token format
+   * @param {Address} to The defi address is the key, the value is amount in amount amount@token format
+   * @param {SelectionModeType} [selectionMode=SelectionModeType.PIE] account selection mode. If "from" param is empty, it will auto select.
+   * @return {Promise<string>}
+   */
+  async sendTokensToAddress (from: Address, to: Address, selectionMode: SelectionModeType = SelectionModeType.PIE): Promise<string> {
```
jingyi2811, line 312: >

website/docs/jellyfish/api/account.md
```diff
@@ -237,3 +237,26 @@ interface AccountHistoryCountOptions {
...
 ```
+
+## sendTokensToAddress
+
+Creates a transfer transaction from your accounts balances.
+
+```ts title="client.account.sendTokensToAddress()"
+
+interface Account {
+  sendTokensToAddress(from: Address, to: Address, selectionMode: SelectionModeType = SelectionModeType.PIE): Promise<string>
```
jingyi2811, line 248: Again, there is a space between sendTokensToAddress and the rest. (Recommend: You can always compare this line with this line `return await this.client.call('sendtokenstoaddress', [from, to, selectionMode], 'number')` above)

packages/jellyfish-api-core/src/category/account.ts
```diff
@@ -370,3 +388,7 @@ export interface AccountHistoryCountOptions {
   txtype?: TxType | string
   no_rewards?: boolean
 }
+
+export interface Address {
+  [key: string]: AccountRegexType[]
+}
```
fuxingloh, line 394:
```suggestion
export interface AddressBalances {
  [key: string]: AccountRegexType[]
}
```

`Address` is quite misleading, how about `AddressBalances`?

fuxingloh (changes requested):

fuxingloh:
Take note of conflicts too.

packages/jellyfish-api-core/src/category/account.ts
siradji: Okay.

packages/jellyfish-api-core/__tests__/category/account/sendTokensToAddress.test.ts
```diff
@@ -0,0 +1,102 @@
...
+    await container.call('utxostoaccount', [{ [address]: '100@0' }])
+    await container.generate(1)
+
+    await container.call('minttokens', [`${amount.toString()}@${symbol}`])
+    await container.generate(1)
+  }
+
+  it('should create a transaction with auto select (empty source address)', async () => {
+    const to = await client.wallet.getNewAddress()
+    const transactionHex = await client.account.sendTokensToAddress({}, { [to]: ['2@DFI'] })
```
jingyi2811, line 50: I actually wanted to comment on this a few days ago. 
Change this to 
`const transactionHex = await client.account.sendTokensToAddress({await client.wallet.getNewAddress()}, { [to]: ['2@DFI'] })`

packages/jellyfish-api-core/__tests__/category/account/sendTokensToAddress.test.ts
```diff
@@ -0,0 +1,102 @@
...
+
+    expect(typeof transactionHex).toStrictEqual('string')
+  })
+
+  it('should create a transaction with Pie selection Mode', async () => {
+    const to = await client.wallet.getNewAddress()
```
jingyi2811, line 56: Same...

packages/jellyfish-api-core/src/category/account.ts
```diff
@@ -31,7 +31,13 @@ export enum TxType {
...
 }
 
-type AccountRegexType = `${number}@${string}`
+export enum SelectionModeType {
+  PIE = 'pie',
+  CRUMBS = 'crumbs',
+  FORWARD = 'forward'
+}
+
+type AccountRegexType = `${string}@${string}`
```
jingyi2811, line 40: If you merge with main, you will get `${number}@${string}`

packages/jellyfish-api-core/src/category/account.ts
```diff
@@ -31,7 +31,13 @@ export enum TxType {
   AUTO_AUTH_PREP = 'A'
 }
 
-type AccountRegexType = `${number}@${string}`
```
jingyi2811, line 34: You should use this

canonbrother (changes requested):
regarding pull_request template, can help follow the same??
eg:
https://github.com/DeFiCh/jellyfish/pull/290#issue-651981430

packages/jellyfish-api-core/__tests__/category/account/sendTokensToAddress.test.ts
siradji: Yeah, elimate unnecessary line of code.

packages/jellyfish-api-core/__tests__/category/account/sendTokensToAddress.test.ts
```diff
@@ -0,0 +1,95 @@
...
+
+    await container.call('utxostoaccount', [{ [address]: '100@0' }])
+    await container.generate(1)
+
+    await container.call('minttokens', [`${amount.toString()}@${symbol}`])
+    await container.generate(1)
+  }
+
+  it('should create a transaction with auto select (empty source address)', async () => {
+    const transactionHex = await client.account.sendTokensToAddress({}, { [await client.wallet.getNewAddress()]: ['2@DFI'] })
```
jingyi2811, line 49: Screenshot 2021-06-03 at 6 20 31 PM

Could you please improve your test by checking whether the recipient address receives 2 DFI or not?

Apply to all tests.

siradji: Implemented that but receiving an empty array.

jingyi2811: Did you add this line?` await container.generate(1)`
The snippet above was actually retrieved from the jellyfish codebase. You may search and take a look at it.

siradji: I have implemented that @jingyi2811.  Thanks for outlining that.

canonbrother (changes requested):
That has some missing test, can you add these?
1. test with `from` and check its balance before after
2. test with insufficient send
3. selectionMode
This is my assumptions for selectionMode behavior
```
addrA (100), addrB(50), addrC(250)
forward: addrA balance should be decreased after sent
crumbs: should select addrB
pie: should select addrC
```

jingyi2811:
Something wrong with your last commit 734aa24
You reverted the changes you made in 02fd385 which was correct

siradji:
> Something wrong with your last commit [734aa24](https://github.com/DeFiCh/jellyfish/commit/734aa244f975e6923a06db067b17b48aff7f736c)
> You reverted the changes you made in [02fd385](https://github.com/DeFiCh/jellyfish/commit/02fd385cdccd1d72a73dc65df6295680fd5e164a) which was correct

@jingyi2811 , what exactly is wrong? I have to make the changes to implement the additional test 
https://github.com/DeFiCh/jellyfish/pull/323#discussion_r644674877

packages/jellyfish-api-core/__tests__/category/account/sendTokensToAddress.test.ts
```diff
@@ -45,45 +45,72 @@ describe('SendTokenToAddress', () => {
...
+  async function sendTokensToAddress (from: AddressBalances, to: AddressBalances, selectionMode?: SelectionModeType): Promise<string> {
+    const transactionHex = await client.account.sendTokensToAddress(from, to, selectionMode)
+    await container.generate(1)
+    return transactionHex
+  }
+
   it('should create a transaction with auto select (empty source address)', async () => {
-    const transactionHex = await client.account.sendTokensToAddress({}, { [await client.wallet.getNewAddress()]: ['2@DFI'] })
+    const address = await client.wallet.getNewAddress()
+    const transactionHex = await sendTokensToAddress({}, { [address]: ['2@DFI'] })
```
jingyi2811, line 56: line 49 is correct.

jingyi2811: Screenshot 2021-06-07 at 4 16 22 PM

siradji: image

Had to change it because of line 66

jingyi2811: No need this line -> `const address = await client.wallet.getNewAddress()`

jingyi2811: Write like this:
`const transactionHex = await sendTokensToAddress({}, { [await client.wallet.getNewAddress()]: ['2@DFI'] })`

jingyi2811: Okay, my fault. Sorry.

siradji: okay

jingyi2811 (approved):

packages/jellyfish-api-core/__tests__/category/account/sendTokensToAddress.test.ts
```diff
@@ -0,0 +1,122 @@
...
+
+    await container.call('minttokens', [`${amount.toString()}@${symbol}`])
+    await container.generate(1)
+  }
+
+  async function sendTokensToAddress (from: AddressBalances, to: AddressBalances, selectionMode?: SelectionModeType): Promise<string> {
+    const transactionHex = await client.account.sendTokensToAddress(from, to, selectionMode)
+    await container.generate(1)
+    return transactionHex
+  }
```
fuxingloh, line 52: I am quite against this kind of test writing convention. This is because you are testing this function hence it should not be abstracted and be written out in detail.

While for the `createToken` that makes sense because it's part of the setup process and not really what you are testing for. You are testing `client.account.sendTokensToAddress` so I expect this to be in the test body `it('')`.

fuxingloh (changes requested):

packages/jellyfish-api-core/__tests__/category/account/sendTokensToAddress.test.ts
siradji: Understood. Actually jimmy and i talked about it. Let me refactor it.

packages/jellyfish-api-core/__tests__/category/account/sendTokensToAddress.test.ts
```diff
@@ -0,0 +1,144 @@
...
+  })
+
+  afterAll(async () => {
+    await container.stop()
+  })
+
+  let addA: string
+
+  async function setup (): Promise<void> {
+    addA = await container.call('getnewaddress')
```
jingyi2811, line 23: `add` should be `address`. The reason is your other variables use the word `address` to represent an address

packages/jellyfish-api-core/__tests__/category/account/sendTokensToAddress.test.ts
```diff
@@ -0,0 +1,144 @@
...
+
+    expect(account).toStrictEqual(['2.00000000@DBTC', '10.00000000@ETH'])
+    expect(typeof transactionHex).toStrictEqual('string')
+  })
+
+  it('should fail and throw an exception if destination address param is empty', async () => {
+    await expect(client.account.sendTokensToAddress({}, {})).rejects.toThrow('zero amounts in "to" param')
+  })
+
+  it('should throw an error with insufficient fund', async () => {
```
jingyi2811, line 139: If I run this test independently, it fails.

jingyi2811: Every `it` should pass if you run it independently. You probably need to change your code so the result is not depends on other tests

packages/jellyfish-api-core/__tests__/category/account/sendTokensToAddress.test.ts
```diff
@@ -0,0 +1,144 @@
...
+    await container.waitForReady()
+    await container.waitForWalletCoinbaseMaturity()
+    await setup()
+  })
+
+  afterAll(async () => {
+    await container.stop()
+  })
+
+  let addA: string
```
canonbrother, line 20: you might need 3 addresses to support your results of pie, crumb and forward

siradji: Yes, i have test two out of the three selection modes. However forward is, in my opinion, not testable because the results are different each time you run the test. Let me know your thoughts.

packages/jellyfish-api-core/__tests__/category/account/sendTokensToAddress.test.ts
```diff
@@ -0,0 +1,154 @@
...
+  afterAll(async () => {
+    await container.stop()
+  })
+
+  let addressA: string
+  let addressB: string
+  let addressC: string
+
+  async function setup (): Promise<void> {
+    addressB = await container.call('getnewaddress')
```
jingyi2811, line 25: Can't you make it in this order: addressA, addressB and addressC

packages/jellyfish-api-core/__tests__/category/account/sendTokensToAddress.test.ts
jingyi2811: Thanks for solving this. I have run all your test separately and all the tests are able to run independently.

website/docs/jellyfish/api/account.md
```diff
@@ -263,3 +263,34 @@ interface AccountHistoryCountOptions {
...
+
+Creates a transfer transaction from your accounts balances.
+
+```ts title="client.account.sendTokensToAddress()"
+
+interface Account {
+  sendTokensToAddress (
+      from: AddressBalances,
+      to: AddressBalances, 
+      options: SendTokensOptions ={selectionMode: SelectionModeType}
```
canonbrother, line 277:
```suggestion
      options: SendTokensOptions = {selectionMode: SelectionModeType}
```

packages/jellyfish-api-core/__tests__/category/account/sendTokensToAddress.test.ts
```diff
@@ -0,0 +1,154 @@
...
+    addressA = await container.call('getnewaddress')
+    addressC = await container.call('getnewaddress')
+
+    await createToken(addressB, 'DBTC', 200, 100)
+    await createToken(addressC, 'DETH', 200, 50)
+    await createToken(addressA, 'ETH', 200, 200)
```
canonbrother, line 31: code addrA, addrB, addrC in order for reader friendly

canonbrother: you need 3 addresses with same coin to test the selectionMode
```
await container.call('accounttoaccount', [tokenAddress, {[addressA]:`${3}@CAT`}])
await container.call('accounttoaccount', [tokenAddress, {[addressB]:`${68}@CAT`}])
await container.call('accounttoaccount', [tokenAddress, {[addressC]:`${89}@CAT`}])
```

packages/jellyfish-api-core/__tests__/category/account/sendTokensToAddress.test.ts
```diff
@@ -0,0 +1,154 @@
...
+  }
+
+  async function createToken (address: string, symbol: string, amount: number, uxtoAmount: number = 100): Promise<void> {
```
canonbrother, line 34: uxtoAmount is not needed

canonbrother (changes requested):
Some notes here:
1. How to know the emptyFrom is auto selection?
2. How to know which address is selected to send token in selectionMode test??
3. test send diff fromAmount to toAmount

the key is the set up, see https://github.com/DeFiCh/jellyfish/pull/323#discussion_r649357820
you might need re-organize the fixtures

packages/jellyfish-api-core/__tests__/category/account/sendTokensToAddress.test.ts
canonbrother: see https://github.com/DeFiCh/jellyfish/pull/323#discussion_r649357820

siradji:
@canonbrother , referring to this [comment](https://github.com/DeFiCh/jellyfish/pull/323#pullrequestreview-676953673), forward selection mode is not behaving as outlined. It is changes for each test you run making it hard to assets.

packages/jellyfish-api-core/__tests__/category/account/sendTokensToAddress.test.ts
```diff
@@ -0,0 +1,134 @@
...
+    expect(transactionHex.length).toStrictEqual(64)
+  })
+
+  it('should create a transaction with source address provided', async () => {
+    const address = await client.wallet.getNewAddress()
+    const transactionHex = await client.account.sendTokensToAddress({ [addressA]: ['10@ETH'] }, { [address]: ['10@ETH'] })
+    await container.generate(1)
+
+    expect(await client.account.getAccount(address)).toStrictEqual(['10.00000000@ETH'])
+    expect(typeof transactionHex).toStrictEqual('string')
```
jingyi2811, line 113: Your code is not consistent. Like here you need to check the length too. Like what you did in line 104.
Advice: Try to reread your code from top to bottom a few times

packages/jellyfish-api-core/__tests__/category/account/sendTokensToAddress.test.ts
```diff
@@ -0,0 +1,134 @@
...
+    addressB = await container.call('getnewaddress')
+    addressC = await container.call('getnewaddress')
+
+    await createToken(tokenAddress, 'CAT', 100)
+    await createToken(tokenAddress, 'ETH', 100)
+    await createToken(tokenAddress, 'DETH', 100)
+
+    await accountToAccount('CAT')
+    await accountToAccount('ETH')
+    await accountToAccount('DETH')
```
canonbrother, line 37: how to test with diff coins set up??
acc1 = acc2acc('bat' , 100)
acc2 = acc2acc('cat' , 60)
acc3 = acc2acc('dog', 80)

sendTokensToAddress((), [addr]: [2@cat], crumbs) // <-- its 100% select acc2, what's the point??

canonbrother (changes requested):

canonbrother:
> @canonbrother , referring to this [comment](https://github.com/DeFiCh/jellyfish/pull/323#pullrequestreview-676953673), forward selection mode is not behaving as outlined. It is changes for each test you run making it hard to assets.

i guess the first acc created is `forward`, can you double check this??

packages/jellyfish-api-core/__tests__/category/account/sendTokensToAddress.test.ts
siradji: Thanks i will do that henceforth.

packages/jellyfish-api-core/__tests__/category/account/sendTokensToAddress.test.ts
siradji:
```
await accountToAccount('CAT')
await accountToAccount('ETH')
await accountToAccount('DETH')
```
The reason for creating multiple tokens is to avoid side effects of other tests. So we can run each test with a different token to avoid side effect (balance getting reduced which will make it hard to assert in other test). 

how to test with diff coins set up??
acc1 = acc2acc('bat' , 100)
acc2 = acc2acc('cat' , 60)
acc3 = acc2acc('dog', 80)

In regards to the comment above, whenever the `accountToAccount` function is called, it does an account to account transaction to 3 accounts, `addressA, addressB, addressC` with different amount. 

```
   await container.call('accounttoaccount', [tokenAddress, { [addressA]: `${20}@${symbol}` }])
    await container.call('accounttoaccount', [tokenAddress, { [addressB]: `${30}@${symbol}` }])
    await container.call('accounttoaccount', [tokenAddress, { [addressC]: `${50}@${symbol}` }])

```
So `sendTokensToAddress((), [addr]: [2@cat], crumbs)` will be addressA because it has the lowest amount of `CAT`

canonbrother: you can set up diff coins on each test, but we only need 1 coin to test selectionMode.
else can you confirm this?
```
// 3 diff coins set up
acc1 = acc2acc('bat' , 100)
acc2 = acc2acc('cat' , 60)
acc3 = acc2acc('dog', 80)

acc2acc(null, { [addrA]: 20@bat }, { crumbs })
which account will be deducted amount?? acc2 or acc1??
```

siradji: @canonbrother , it will be deducted from acc2 of course. I wrote a test checking that.

canonbrother: can you point me to the test??

packages/jellyfish-api-core/__tests__/category/account/sendTokensToAddress.test.ts
```diff
@@ -0,0 +1,136 @@
...
+
+    const account = await client.account.getAccount(addressC)
+
+    expect(await client.account.getAccount(address)).toStrictEqual(['2.00000000@CAT'])
+    expect(typeof transactionHex).toStrictEqual('string')
+    expect(transactionHex.length).toStrictEqual(64)
+    expect(account[0]).toStrictEqual('48.00000000@CAT')
+  })
+
+  it('should create a transaction with Crumbs selection mode', async () => {
```
siradji, line 81: sendtokenstoaddress with selection mode = crumbs

packages/jellyfish-api-core/__tests__/category/account/sendTokensToAddress.test.ts
```diff
@@ -0,0 +1,136 @@
...
+    await container.generate(1)
+
+    const account = await client.account.getAccount(addressA)
+
+    expect(account[2]).toStrictEqual('10.00000000@DETH')
+    expect(await client.account.getAccount(address)).toStrictEqual(['10.00000000@DETH'])
+    expect(typeof transactionHex).toStrictEqual('string')
+    expect(transactionHex.length).toStrictEqual(64)
+  })
+
```
siradji, line 96: Test with selection mode = crumbs

packages/jellyfish-api-core/__tests__/category/account/sendTokensToAddress.test.ts
siradji: No, only test same token gets deducted. Please refer to this. https://github.com/DeFiCh/jellyfish/pull/323/files#r651480837

packages/jellyfish-api-core/__tests__/category/account/sendTokensToAddress.test.ts
canonbrother: this is correct

packages/jellyfish-api-core/__tests__/category/account/sendTokensToAddress.test.ts
```diff
@@ -0,0 +1,136 @@
...
+    await accountToAccount('ETH')
+    await accountToAccount('DETH')
+  }
+
+  async function accountToAccount (symbol: string): Promise<void> {
+    await container.call('accounttoaccount', [tokenAddress, { [addressA]: `${20}@${symbol}` }])
+    await container.call('accounttoaccount', [tokenAddress, { [addressB]: `${30}@${symbol}` }])
+    await container.call('accounttoaccount', [tokenAddress, { [addressC]: `${50}@${symbol}` }])
+    await container.generate(1)
+  }
```
canonbrother, line 45: this code messed up the test context, as mentioned, you only need 1 coin to test the selectionMode
don't add another unrelated fixtures
```
await accountToAccount('CAT')
await accountToAccount('ETH')
await accountToAccount('DETH')
```

packages/jellyfish-api-core/__tests__/category/account/sendTokensToAddress.test.ts
```diff
@@ -0,0 +1,136 @@
...
+    expect(transactionHex.length).toStrictEqual(64)
+    expect(account[0]).toStrictEqual('48.00000000@CAT')
+  })
+
+  it('should create a transaction with Crumbs selection mode', async () => {
+    const options: SendTokensOptions = {
+      selectionMode: SelectionModeType.CRUMBS
+    }
+    const address = await client.wallet.getNewAddress()
+    const transactionHex = await client.account.sendTokensToAddress({}, { [address]: ['10@DETH'] }, options)
```
canonbrother, line 86:
```suggestion
    const address = await client.wallet.getNewAddress()
    const transactionHex = await client.account.sendTokensToAddress({}, { [address]: ['10@DETH'] }, {selectionMode: SelectionModeType.CRUMBS})
```

packages/jellyfish-api-core/__tests__/category/account/sendTokensToAddress.test.ts
```diff
@@ -0,0 +1,136 @@
...
+
+    await accountToAccount('CAT')
+    await accountToAccount('ETH')
+    await accountToAccount('DETH')
+  }
+
+  async function accountToAccount (symbol: string): Promise<void> {
+    await container.call('accounttoaccount', [tokenAddress, { [addressA]: `${20}@${symbol}` }])
+    await container.call('accounttoaccount', [tokenAddress, { [addressB]: `${30}@${symbol}` }])
+    await container.call('accounttoaccount', [tokenAddress, { [addressC]: `${50}@${symbol}` }])
```
canonbrother, line 43: always `65, 43, 187` to test, prevent `100, 50, 10`

packages/jellyfish-api-core/__tests__/category/account/sendTokensToAddress.test.ts
canonbrother: add before deducted account and assert the amount

packages/jellyfish-api-core/__tests__/category/account/sendTokensToAddress.test.ts
```diff
@@ -0,0 +1,136 @@
...
+    await container.generate(1)
+
+    await container.call('utxostoaccount', [{ [address]: '100@0' }])
+    await container.generate(1)
+
+    await container.call('minttokens', [`${amount.toString()}@${symbol}`])
+    await container.generate(1)
+  }
+
+  it('should create a transaction with auto select (empty source address) Pie selection mode', async () => {
```
canonbrother, line 68: need check before deducted account amount first

canonbrother (changes requested):
can you rename the `transactionHex` to `hex`?
is the `forward` not testable?

packages/jellyfish-api-core/__tests__/category/account/sendTokensToAddress.test.ts
siradji: One coin will have side effects. I tried doing one coin, but having to run multiple test cases will make it hard to assert the exact balances before and after tokens are sent. @jingyi2811 told me make sure tests are passing independently and also the whole block. Doing one coin will make the tests fail if run independently.

siradji: For example, it i run a test to check for PIE and deduct 10 coin and then test to check crumbs , obviously the balances will be different based on whether the test is run in block (the whole suite) or independently (just the it)

canonbrother: try this
```
setup() {
    addrA = await container.call('getnewaddress')
    addrB = await container.call('getnewaddress')
    addrC1 = await container.call('getnewaddress')
    addrC2 = await container.call('getnewaddress')
    addrC3 = await container.call('getnewaddress')
    
    const tokenAddrA = await container.call('getnewaddress')
    const tokenAddrB = await container.call('getnewaddress')
    const tokenAddrC = await container.call('getnewaddress')
    await createToken(tokenAddrA, 'ANT', 200)
    await createToken(tokenAddrB, 'BAT', 200)
    await createToken(tokenAddrC, 'CAT', 200)
    
    await container.call('acc2acc', [tokenAddrA, {[addrA]:`${3}@ANT`}])
    await container.call('acc2acc', [tokenAddrB, {[addrB]:`${68}@BAT`}])
    await container.call('acc2acc', [tokenAddrC, {[addrC1]:`${89}@CAT`}])
    await container.call('acc2acc', [tokenAddrC, {[addrC2]:`${15}@CAT`}])
    await container.call('acc2acc', [tokenAddrC, {[addrC3]:`${54}@CAT`}])
    await container.generate(1)
}

Comment (siradji) [reply packages/jellyfish-api-core/__tests__/category/account/sendTokensToAddress.test.ts:45]: Thanks for the setup. It may work if we wont be testing PIE and CRUMBS. We can use CAT to test both selection modes but due to side effects it will won't work. The problem is that we can not use the same coin to test and assert balances which in what we do in PIE  and CRUMB tests

Comment (siradji) [reply packages/jellyfish-api-core/__tests__/category/account/sendTokensToAddress.test.ts:45]: A solution here is to have and addrB1, addrB2, and addrB3 to use to test one on the selection modes.

Comment (canonbrother) [reply packages/jellyfish-api-core/__tests__/category/account/sendTokensToAddress.test.ts:45]: why not?? use `CAT` to test both selectionMode
```
await container.call('acc2acc', [tokenAddrC, {[addrC1]:`${58}@CAT`}])  // forward
await container.call('acc2acc', [tokenAddrC, {[addrC2]:`${3}@CAT`}])  // crumbs
await container.call('acc2acc', [tokenAddrC, {[addrC3]:`${103}@CAT`}]) // pie
```

Comment (siradji) [reply packages/jellyfish-api-core/__tests__/category/account/sendTokensToAddress.test.ts:45]: Okay, sorry my mistake it will work since we wont be testing forward.

File: packages/jellyfish-api-core/__tests__/category/account/sendTokensToAddress.test.ts [hunk 27]
```diff
@@ -0,0 +1,163 @@
...
+    expect(typeof hex).toStrictEqual('string')
+    expect(hex.length).toStrictEqual(64)
+
+    await container.generate(1)
+
+    const accPieAfter = (await client.account.getAccount(addrCPie))[0]
+    expect(accToNum(accPieAfter)).toStrictEqual(accToNum(accPieBefore) - 4)
+
+    const accReceiver = await client.account.getAccount(addrReceiver)
+    expect(accReceiver[0]).toStrictEqual('4.00000000@CAT')
```
jingyi2811, line 88: The way you write the code is consistent. (See line 84)
`const accReceiver = (await client.account.getAccount(addrReceiver)[0])    expect(accReceiver).toStrictEqual('4.00000000@CAT')`

jingyi2811: Also, apply to all similar codes in this PR.

siradji: There is a difference between line 84 and 87. The former (line 84) is an array of balances and later (line 87) is a string.

siradji: okay

siradji: Resolved now.

canonbrother (changes requested):
one more test for sending ANT to CAT

packages/jellyfish-api-core/__tests__/category/account/sendTokensToAddress.test.ts
```diff
@@ -0,0 +1,163 @@
...
+  let addrCForward: string
+  let addrCCrumbs: string
+  let addrCPie: string
+
+  async function setup (): Promise<void> {
+    addrA = await container.call('getnewaddress')
+    addrB = await container.call('getnewaddress')
+    addrCForward = await container.call('getnewaddress')
+    addrCCrumbs = await container.call('getnewaddress')
+    addrCPie = await container.call('getnewaddress')
```
canonbrother, line 31: can you try hardcode the address?? Be noted that the `forward` must the on first order ASC
Then test the forward selection mode

packages/jellyfish-api-core/__tests__/category/account/sendTokensToAddress.test.ts
```diff
@@ -0,0 +1,163 @@
...
+
+    expect(await client.account.getAccount(addrReceiver)).toStrictEqual(['10.00000000@ANT'])
+  })
+
+  it('should fail and throw an exception when no receiver address', async () => {
+    await expect(client.account.sendTokensToAddress({}, {})).rejects.toThrow('zero amounts in "to" param')
+  })
+
+  it('should throw an error when insufficient fund', async () => {
+    const promise = client.account.sendTokensToAddress({}, { [await client.wallet.getNewAddress()]: ['500@CAT'] })
```
jingyi2811, line 159: Choose either `container.call('getnewaddress')` or `await client.wallet.getNewAddress()` and use it in your code.  I would use `container.call('getnewaddress')`

siradji:
> one more test for sending ANT to CAT

@canonbrother , i think `sendTokensToAddress` only support sending the same currency type.

canonbrother:
> > one more test for sending ANT to CAT
> 
> @canonbrother , i think `sendTokensToAddress` only support sending the same currency type.

yes i expect the error throw

packages/jellyfish-api-core/__tests__/category/account/sendTokensToAddress.test.ts
siradji: Refactored. From henceforth my code will be consistent. Thanks for getting sense into me.

packages/jellyfish-api-core/__tests__/category/account/sendTokensToAddress.test.ts
siradji: I have tried doing that with the address sorted (please refer to the block below), but it is not working. Actually, it even messed the other selectionMode tests

```
    2MydTm43kmMTmRnHgnPg91NB4jHhZraNLGR
    2N1XhV5zFqLkEeJK2VoAMEWFGuM1BjYUxga
    2N427YresEoYjn83QML2Y5SWnSDtnTLdS73
```

canonbrother (dismissed):

packages/jellyfish-api-core/__tests__/category/account/sendTokensToAddress.test.ts
```diff
@@ -0,0 +1,176 @@
...
+  it('should sendTokensToAddress with multiple receiver tokens', async () => {
+    const address = await container.call('getnewaddress')
+    const hex = await client.account.sendTokensToAddress({}, { [address]: ['2@ANT', '5@CAT', '10@BAT'] })
+
+    expect(typeof hex).toStrictEqual('string')
+    expect(hex.length).toStrictEqual(64)
+
+    await container.generate(1)
+
+    expect(await client.account.getAccount(address)).toStrictEqual(['2.00000000@ANT', '10.00000000@BAT', '5.00000000@CAT'])
```
jingyi2811, line 140: Rewrite this in this order: 2, 5, 10

packages/jellyfish-api-core/__tests__/category/account/sendTokensToAddress.test.ts
```diff
@@ -0,0 +1,176 @@
...
+
+    const accForwardAfter = (await client.account.getAccount(addrCCrumbs))[0]
+    expect(accToNum(accForwardAfter)).toStrictEqual(accToNum(accForwardBefore) - 6)
+
+    const accReceiver = (await client.account.getAccount(addrReceiver))[0]
+    expect(accReceiver).toStrictEqual('6.00000000@CAT')
+  })
+
+  it('should sendTokensToAddress with multiple receiver tokens', async () => {
+    const address = await container.call('getnewaddress')
```
jingyi2811, line 132: Shouldn't this address addrReceiver?

packages/jellyfish-api-core/src/category/account.ts
```diff
@@ -311,6 +317,23 @@ export class Account {
   ): Promise<number> {
     return await this.client.call('accounthistorycount', [owner, options], 'number')
   }
+
+  /**
+   * Creates a transfer transaction from your accounts balances.
+   *
+   * @param {AddressBalances} from The source defi address is the key, the value is amount in amount amount@token format
+   * @param {AddressBalances} to The defi address is the key, the value is amount in amount amount@token format
+   * @param {SendTokensOptions} [options]
```
jingyi2811, line 326: @param {SendTokensOptions} [options = { selectionMode: SelectionModeType.PIE }]

packages/jellyfish-api-core/__tests__/category/account/sendTokensToAddress.test.ts
```diff
@@ -0,0 +1,176 @@
...
+
+    const accReceiver = (await client.account.getAccount(addrReceiver))[0]
+    expect(accReceiver).toStrictEqual('2.00000000@CAT')
+  })
+
+  // NOTE(canonrother): "selectionMode: forward" picks the address name order by ASC, its hard to test as address is random generated
+  it.skip('should sendTokensToAddress with selectionMode forward', async () => {
+    const accForwardBefore = (await client.account.getAccount(addrCCrumbs))[0]
+
+    const addrReceiver = await client.wallet.getNewAddress()
```
jingyi2811, line 115: Choose 1: 
await container.call('getnewaddress')
await client.wallet.getNewAddress()

website/docs/jellyfish/api/account.md
```diff
@@ -263,3 +263,34 @@ interface AccountHistoryCountOptions {
...
+
+Creates a transfer transaction from your accounts balances.
+
+```ts title="client.account.sendTokensToAddress()"
+
+interface Account {
+  sendTokensToAddress (
+      from: AddressBalances,
+      to: AddressBalances, 
+      options: SendTokensOptions = {selectionMode: SelectionModeType}
```
jingyi2811, line 277: SelectionModeType.PIE

packages/jellyfish-api-core/__tests__/category/account/sendTokensToAddress.test.ts
siradji: But that is what i am getting returned.

jingyi2811 (dismissed):

canonbrother:
i'm good overall

packages/jellyfish-api-core/src/category/account.ts
```diff
@@ -311,6 +317,23 @@ export class Account {
...
+
+  /**
+   * Creates a transfer transaction from your accounts balances.
+   *
+   * @param {AddressBalances} from The source defi address is the key, the value is amount in amount amount@token format
+   * @param {AddressBalances} to The defi address is the key, the value is amount in amount amount@token format
+   * @param {SendTokensOptions} [options = { selectionMode: SelectionModeType.PIE }]
+   * @param {SelectionModeType} [options.selectionMode] Account selection mode. If "from" param is empty, it will auto select.
+   * @return {Promise<string>}
+   */
```
fuxingloh, line 329:
```suggestion
  /**
   * Creates a transfer transaction from your accounts balances.
   *
   * @param {AddressBalances} from source address as the key, the value is amount formatted as amount@token
   * @param {AddressBalances} to address as the key, the value is amount formatted as amount@token
   * @param {SendTokensOptions} [options = { selectionMode: SelectionModeType.PIE }]
   * @param {SelectionModeType} [options.selectionMode] Account selection mode. If "from" param is empty, it will auto select.
   * @return {Promise<string>}
   */
```

website/docs/jellyfish/api/account.md
```diff
@@ -263,3 +263,34 @@ interface AccountHistoryCountOptions {
...
+
+```ts title="client.account.sendTokensToAddress()"
+
+interface Account {
+  sendTokensToAddress (
+      from: AddressBalances,
+      to: AddressBalances,
+      options: SendTokensOptions = { selectionMode: SelectionModeType.PIE }
+  ): Promise<string>
+}
```
fuxingloh, line 279:
```suggestion
interface account {
  sendTokensToAddress (
    from: AddressBalances,
    to: AddressBalances,
    options: SendTokensOptions = { selectionMode: SelectionModeType.PIE }
  ): Promise<string>
}
```

packages/jellyfish-api-core/__tests__/category/account/sendTokensToAddress.test.ts
```diff
@@ -0,0 +1,176 @@
...
+
+  afterAll(async () => {
+    await container.stop()
+  })
+
+  let addrA: string
+  let addrB: string
+  let addrCForward: string
+  let addrCCrumbs: string
+  let addrCPie: string
```
fuxingloh, line 24: no global declaration, either put in within test code `it` or don't have it entirely. It's very hard to read and maintain test code like this.

If you are trying to use the same logical condition for tests, use inner `describe`.

packages/jellyfish-api-core/__tests__/category/account/sendTokensToAddress.test.ts
```diff
@@ -0,0 +1,176 @@
...
+    await container.call('minttokens', [`${amount.toString()}@${symbol}`])
+    await container.generate(1)
+  }
+
+  function accToNum (acc: string): number {
+    return Number(acc.split('@')[0])
+  }
+
+  it('should sendTokensToAddress with selectionMode pie', async () => {
+    const accPieBefore = (await client.account.getAccount(addrCPie))[0]
```
fuxingloh, line 74: don't need to further specify `pie` your entire `it` section is already about pie selection mode.

```suggestion
    const amountBefore = (await client.account.getAccount(addrCPie))[0]
```

fuxingloh: and keep a habit of using the entire full name, especially within tests. We should have to guess what is `addr` `C` `Pie`

packages/jellyfish-api-core/__tests__/category/account/sendTokensToAddress.test.ts
fuxingloh: A test should be independent, succinct, and self-sufficient. Don't make it complex by introducing global parameters when setup that could be localized.

fuxingloh (changes requested):
1. take note of small details
2. improve on code testing, don't declare and use a global variable if you don't have to, use additional describe blocks if you need to.

fuxingloh:
refactor your test

fuxingloh (approved):
thanks this looks great ❤️

canonbrother (approved):

PR #324, opened 2021-06-01
WalletHdNodeProvider signing interface

ivan-zynesis:
…to constructor

/kind feature

Allow caller (usually application layer) to attach callback triggered when any transaction signed by the provider

Fixes #

### DIFF
--- a/packages/jellyfish-wallet-mnemonic/src/mnemonic.ts
+++ b/packages/jellyfish-wallet-mnemonic/src/mnemonic.ts
@@ -1,4 +1,4 @@
-import { WalletHdNode, WalletHdNodeProvider } from '@defichain/jellyfish-wallet'
+import { SigningInterface, WalletHdNode, WalletHdNodeProvider } from '@defichain/jellyfish-wallet'
 import { DERSignature } from '@defichain/jellyfish-crypto'
 import {
   Transaction,
@@ -154,7 +154,7 @@ export class MnemonicHdNode implements WalletHdNode {
 /**
  * Provider that derive MnemonicHdNode from root. Uses a lite on demand derivation.
  */
-export class MnemonicHdNodeProvider implements WalletHdNodeProvider<MnemonicHdNode> {
+export class MnemonicHdNodeProvider extends WalletHdNodeProvider<MnemonicHdNode> {
   /**
    * @param {Buffer} seed of the hd node
    * @param {Bip32Options} options for chain agnostic generation of public/private keys
@@ -166,7 +166,8 @@ export class MnemonicHdNodeProvider implements WalletHdNodeProvider<MnemonicHdNo
 
   private readonly root: bip32.BIP32Interface
 
-  private constructor (root: bip32.BIP32Interface) {
+  private constructor (root: bip32.BIP32Interface, signingCb?: SigningInterface) {
+    super(signingCb)
     this.root = root
   }
 
--- a/packages/jellyfish-wallet-mnemonic/src/mnemonic.ts
+++ b/packages/jellyfish-wallet-mnemonic/src/mnemonic.ts
@@ -128,18 +128,20 @@ export class MnemonicHdNode implements WalletHdNode {
       return { prevout: prevout, ellipticPair: this }
     })
 
+    if (this.signingCb !== undefined) {
+      const unsignedBuffer = new SmartBuffer()
+      new CTransaction(transaction).toBuffer(unsignedBuffer)
+      await this.signingCb.beforeSign(unsignedBuffer.toBuffer(), transaction)
+    }
+
     const signed = await TransactionSigner.sign(transaction, inputs, {
       sigHashType: SIGHASH.ALL
     })
 
     if (this.signingCb !== undefined) {
-      const unsignedBuffer = new SmartBuffer()
-      new CTransaction(transaction).toBuffer(unsignedBuffer)
-      await this.signingCb.unsigned(unsignedBuffer.toBuffer(), transaction)
-
       const signedBuffer = new SmartBuffer()
       new CTransaction(signed).toBuffer(signedBuffer)
-      await this.signingCb.signed(signedBuffer.toBuffer(), signed)
+      await this.signingCb.afterSign(signedBuffer.toBuffer(), signed)
     }
 
     return signed
@@ -168,15 +170,14 @@ export class MnemonicHdNode implements WalletHdNode {
 }
 
 export interface SigningInterface {
-  unsigned: (buffer: Buffer, tx: Transaction) => Promise<void>
-  signed: (buffer: Buffer, tx: TransactionSegWit) => Promise<void>
+  beforeSign: (buffer: Buffer, tx: Transaction) => Promise<void>
+  afterSign: (buffer: Buffer, tx: TransactionSegWit) => Promise<void>
 }
 
 /**
  * Provider that derive MnemonicHdNode from root. Uses a lite on demand derivation.
  */
 export class MnemonicHdNodeProvider implements WalletHdNodeProvider<MnemonicHdNode> {
-  walletHdNode?: MnemonicHdNode
   signingCb?: SigningInterface
 
   /**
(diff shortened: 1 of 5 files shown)


packages/jellyfish-wallet/src/wallet_hd_node.ts
```diff
@@ -20,16 +21,58 @@ export interface WalletHdNode extends EllipticPair {
...
+
+  /**
+   * Derive wallet hd node and store with this provider
+   * @param {string} path
+   * @returns {T}
+   */
+  deriveAndAssign (path: string): T {
+    this.walletHdNode = this.derive(path)
+    return this.walletHdNode
+  }
```
fuxingloh, line 52: we cannot hold as `state` in WalletHdNodeProvider. this must be a provider that doesn't do anything

fuxingloh (changes requested):
this should only be done exclusively on mnemonic as the signing provider should be agnostic.

meaning if ledger hw does this signing, the interface/callback will be on ledger. (displaying on ledger)

packages/jellyfish-wallet/src/wallet_hd_node.ts
```diff
@@ -20,16 +21,58 @@ export interface WalletHdNode extends EllipticPair {
...
+
 /**
  * WalletHdNode uses the provider model to allow jellyfish-wallet to derive/provide a WalletHdNode from any sources.
  * This design keep WalletHdNode derivation agnostic of any implementation, allowing a lite
  * implementation where WalletHdNode are derived on demand.
  */
-export interface WalletHdNodeProvider<T extends WalletHdNode> {
+export abstract class WalletHdNodeProvider<T extends WalletHdNode> {
+  signingCb?: SigningInterface
+  private walletHdNode?: T
```
fuxingloh, line 36: best not to hold state also, this can easily be mem dump

fuxingloh: The current derive implementation is on demand for this reason.

packages/jellyfish-wallet-mnemonic/src/mnemonic.ts
```diff
@@ -151,26 +167,35 @@ export class MnemonicHdNode implements WalletHdNode {
...
+export interface SigningInterface {
+  unsigned: (buffer: Buffer, tx: Transaction) => Promise<void>
+  signed: (buffer: Buffer, tx: TransactionSegWit) => Promise<void>
+}
+
 /**
  * Provider that derive MnemonicHdNode from root. Uses a lite on demand derivation.
  */
 export class MnemonicHdNodeProvider implements WalletHdNodeProvider<MnemonicHdNode> {
+  walletHdNode?: MnemonicHdNode
```
fuxingloh, line 179:
```suggestion
```

should not required anymore

packages/jellyfish-wallet/src/wallet_hd_node.ts
```diff
@@ -26,7 +26,6 @@ export interface WalletHdNode extends EllipticPair {
  * implementation where WalletHdNode are derived on demand.
  */
 export interface WalletHdNodeProvider<T extends WalletHdNode> {
-
```
fuxingloh, line 29: do revert this change

packages/jellyfish-wallet-mnemonic/src/mnemonic.ts
```diff
@@ -151,26 +167,35 @@ export class MnemonicHdNode implements WalletHdNode {
   }
 }
 
+export interface SigningInterface {
+  unsigned: (buffer: Buffer, tx: Transaction) => Promise<void>
```
fuxingloh, line 171:
```suggestion
  before: (buffer: Buffer, tx: Transaction) => Promise<void>
```

packages/jellyfish-wallet-mnemonic/src/mnemonic.ts
```diff
@@ -151,26 +167,35 @@ export class MnemonicHdNode implements WalletHdNode {
...
+  signed: (buffer: Buffer, tx: TransactionSegWit) => Promise<void>
```
fuxingloh, line 172:
```suggestion
  after: (buffer: Buffer, tx: TransactionSegWit) => Promise<void>
```

Document this interface

packages/jellyfish-wallet-mnemonic/src/mnemonic.ts
```diff
@@ -124,9 +127,22 @@ export class MnemonicHdNode implements WalletHdNode {
     const inputs: SignInputOption[] = prevouts.map(prevout => {
       return { prevout: prevout, ellipticPair: this }
     })
-    return TransactionSigner.sign(transaction, inputs, {
+
+    const signed = await TransactionSigner.sign(transaction, inputs, {
       sigHashType: SIGHASH.ALL
     })
```
fuxingloh, line 133: when signingCb get passed in you should await until the promise resolve before signing.

fuxingloh (changes requested):
I think you might have some misconception of what interface means in this case. (by reading your PR description)

The `SigningInterface` is not really an UI `interface` per se. It's more of a technical term.

> By definition: a point where two systems, subjects, organizations, etc. meet and interact.

The implementation, therefore, is an Interface between signing. before and after to prevent unauthorized signing.

`Unsigned Tx` -> `Interface` -> `PrivKey` -> `Signed Tx`

packages/jellyfish-wallet-mnemonic/src/mnemonic.ts
```diff
@@ -151,26 +169,34 @@ export class MnemonicHdNode implements WalletHdNode {
...
+    return new MnemonicHdNodeProvider(node, signingCb)
   }
 
   private readonly root: bip32.BIP32Interface
 
-  private constructor (root: bip32.BIP32Interface) {
+  private constructor (root: bip32.BIP32Interface, signingCb?: SigningInterface) {
     this.root = root
+    this.signingCb = signingCb
   }
```
fuxingloh, line 197:
```suggestion
  private constructor (
    private readonly root: bip32.BIP32Interface, 
    private readonly signingCb?: SigningInterface) {
  }
```

Recently found out you could do this in TS.

packages/jellyfish-wallet-mnemonic/src/mnemonic.ts
```diff
@@ -151,26 +169,34 @@ export class MnemonicHdNode implements WalletHdNode {
...
    * @param {Bip32Options} options for chain agnostic generation of public/private keys
    */
-  static fromSeed (seed: Buffer, options: Bip32Options): MnemonicHdNodeProvider {
+  static fromSeed (seed: Buffer, options: Bip32Options, signingCb?: SigningInterface): MnemonicHdNodeProvider {
     const node = bip32.fromSeed(seed, options)
-    return new MnemonicHdNodeProvider(node)
+    return new MnemonicHdNodeProvider(node, signingCb)
   }
 
   private readonly root: bip32.BIP32Interface
```
fuxingloh, line 192:
```suggestion
```

packages/jellyfish-wallet-mnemonic/src/mnemonic.ts
```diff
@@ -151,26 +169,34 @@ export class MnemonicHdNode implements WalletHdNode {
...
+  beforeSign: (buffer: Buffer, tx: Transaction) => Promise<void>
+  afterSign: (buffer: Buffer, tx: TransactionSegWit) => Promise<void>
+}
+
 /**
  * Provider that derive MnemonicHdNode from root. Uses a lite on demand derivation.
  */
 export class MnemonicHdNodeProvider implements WalletHdNodeProvider<MnemonicHdNode> {
+  signingCb?: SigningInterface
+
```
fuxingloh, line 182:
```suggestion
```

packages/jellyfish-wallet-mnemonic/src/mnemonic.ts
```diff
@@ -79,10 +80,12 @@ export function mnemonicToSeed (mnemonic: string[]): Buffer {
...
   private readonly root: bip32.BIP32Interface
   private readonly path: string
+  private readonly signingCb?: SigningInterface
 
-  constructor (root: bip32.BIP32Interface, path: string) {
+  constructor (root: bip32.BIP32Interface, path: string, signingCb?: SigningInterface) {
     this.root = root
     this.path = path
+    this.signingCb = signingCb
   }
```
fuxingloh, line 89: I guess you could do this here as well for cleaner code.

fuxingloh (changes requested):
This seems too easy, 🤷. I guess it's good but didn't realize it's just this. 

Hmm. Might be missing something. I will review this again after some thinking.

PR #330, opened 2021-06-02
added getRpcInfo rpc

aikchun:

/kind feature

Implemented RPC for #48

1. server.getrpcinfo

Fixes #

### DIFF
--- a/packages/jellyfish-api-core/__tests__/category/server/getRpcInfo.test.ts
+++ b/packages/jellyfish-api-core/__tests__/category/server/getRpcInfo.test.ts
@@ -20,29 +20,6 @@ describe('Server on masternode', () => {
       const info = await client.server.getRpcInfo()
       expect(info.active_commands.length).toBeGreaterThan(0)
     })
-
-    // const info = await client.mining.getMiningInfo()
-    // const mn1 = info.masternodes[0]
-    //
-    // expect(info.blocks).toBeGreaterThan(0)
-    //
-    // expect(info.currentblockweight).toBeGreaterThan(0)
-    // expect(info.currentblocktx).toStrictEqual(0)
-    //
-    // expect(info.difficulty).toBeDefined()
-    // expect(info.isoperator).toStrictEqual(true)
-    //
-    // expect(mn1.masternodeid).toBeDefined()
-    // expect(mn1.masternodeoperator).toBeDefined()
-    // expect(mn1.masternodestate).toStrictEqual('ENABLED')
-    // expect(mn1.generate).toStrictEqual(true)
-    // expect(mn1.mintedblocks).toStrictEqual(0)
-    // expect(mn1.lastblockcreationattempt).not.toStrictEqual('0')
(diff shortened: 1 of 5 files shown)


packages/jellyfish-api-core/src/category/server.ts
```diff
@@ -0,0 +1,27 @@
...
+export class Server {
+  private readonly client: ApiClient
+
+  constructor (client: ApiClient) {
+    this.client = client
+  }
+
+  /**
+   * Returns details of the RPC server
+   * @return {Promise<RpcInfo>}
```
jingyi2811, line 12: Add a break line

packages/jellyfish-api-core/__tests__/category/server/getRpcInfo.test.ts
```diff
@@ -0,0 +1,45 @@
...
+  })
+
+  afterAll(async () => {
+    await container.stop()
+  })
+
+  it('should getRpcInfo', async () => {
+    await waitForExpect(async () => {
+      const info = await client.server.getRpcInfo()
+      expect(info.active_commands.length).toBeGreaterThan(0)
```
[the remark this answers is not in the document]
fuxingloh: @aikchun to assert type is a string for regression purpose, as javascript is not type safe.

packages/jellyfish-api-core/__tests__/category/server/getRpcInfo.test.ts
```diff
@@ -0,0 +1,27 @@
...
+    await container.start()
+    await container.waitForReady()
+  })
+
+  afterAll(async () => {
+    await container.stop()
+  })
+
+  it('should getRpcInfo', async () => {
+    await waitForExpect(async () => {
```
fuxingloh, line 19: Are there any reasons for using `waitForExpect`? In the past, there were flaky conditions that require that, it's mostly resolved now. You shouldn't need to use this, it creates non-determinstic tests.

aikchun: > In the past, there were flaky conditions that require that, it's mostly resolved now.

Understood. I know that `waitForExpect` to meant to be used for asynchronous api calls or e2e tests therefore it seems appropriate to be included here. I suppose regular async await will suffice.

fuxingloh (approved):

packages/jellyfish-api-core/__tests__/category/server/getRpcInfo.test.ts
fuxingloh: For some context that might be relevant for you, the MasterNodeContainer automatically generate blocks to move the ledger forward periodically. We don't use `waitForExpect` as we want to deterministically reach a block at a height. 

E.g. send balance, use `container.generate(1)` to force a block generation to move balance. Then lastly check updated balances.

canonbrother (approved):

Issue #133, opened 2021-04-18
jellyfish-wallet-mnemonic to support encrypt and decrypt


fuxingloh:
#### What would you like to be added:

1. `jellyfish-wallet-mnemonic` with `encrypt` and `decrypt` function extended above 24 word.
2. 24 word must unlock the wallet, this just encrypt the 24 words on rest.

#### Why is this needed:

The 24 words mnemonic seed should serve as a cold storage of your HD seed.
However the 24 words should not sit in your device (phone/browser/desktop) unencrypted as rest.
Encryption can be done with biometric authentication provided by device, `jellyfish-wallet-mnemonic` just need to provide the interface to do it.

Reference example: https://github.com/bitcoin/bips/blob/master/bip-0038.mediawiki

ivan-zynesis:
@fuxingloh Not very certain about the requirement, this is allow user store an the encrypted mnemonic during `first time create` or `import`? The encrypted is protected with a "simpler" (than the ori mnemonic seed) user selected passphrase?

```
interface EncryptionOptions {
  passphrase: string
  encryptedSetter: () => void
  encryptedGetter: () => string
}

static fromSeed (seed: Buffer, options: Bip32Options, encOpt?: EncryptionOptions) {}
```

fuxingloh:
1. Generate mnemonic seed for cold storage
2. Generated mnemonic seed must be the actual seed used for signing transaction
3. Mnemonic seed stored at rest is dangerous
4. Encrypt mnemonic seed while it's at rest
5. When the user wants to perform a transaction, decrypt encrypted mnemonic seed

So yeah, the encryption protects the original mnemonic seed [at rest](https://en.wikipedia.org/wiki/Data_at_rest). It is only loaded into memory when it's required for a brief period.

fuxingloh:
There is also some flow/complications to this, as the wallet should only encrypt private key. The pubKey should be accessible at rest.

PR #331, opened 2021-06-02
`jellyfish-crypto`: aes256

ivan-zynesis:

/kind feature

Dependency of #133 .
Existing lib in found in npm has no TS @types support, so build a simple one directly from `crypto`.
Simplified implementation, all arg and return are buffer only.

Fixes #

### DIFF
--- a/packages/jellyfish-crypto/__tests__/aes256.test.ts
+++ b/packages/jellyfish-crypto/__tests__/aes256.test.ts
@@ -1,0 +1,37 @@
+import { Aes256 } from '../src'
+
+const raw = 'e9873d79c6d87dc0fb6a5778633389f4e93213303da61f20bd67fc233aa33262'
+const privateKey = Buffer.from(raw, 'hex')
+const passphrase = Buffer.from('password', 'ascii')
+
+describe('Aes256', () => {
+  let encrypted: Buffer
+
+  it('', () => {
+    encrypted = Aes256.encrypt(passphrase, privateKey)
+    expect(encrypted.length).toStrictEqual(48) // [16 bytes salt, 32 bytes cipher]
+  })
+
+  it('encrypt', () => {
+    encrypted = Aes256.encrypt(passphrase, privateKey)
+    expect(encrypted.length).toStrictEqual(48) // [16 bytes salt, 32 bytes cipher]
+  })
+
+  it('decrypt - with valid passphrase', () => {
+    const decrypted = Aes256.decrypt(passphrase, encrypted)
+    expect(decrypted.toString('hex')).toStrictEqual(raw)
+  })
+
+  it('decrypt - with invalid passphrase', () => {
+    const invalid = Aes256.decrypt(passphrase.slice(1), encrypted)
+    expect(invalid.length).toStrictEqual(32)
+    expect(invalid.toString('hex')).not.toStrictEqual(raw)
+  })
+
+  it('decrypt - data too short, insufficient length to include salt', () => {
+    const invalidData = Buffer.alloc(16)
+    expect(() => {
+      Aes256.decrypt(passphrase, invalidData)
+    }).toThrow('Provided "encrypted" must decrypt to a non-empty string or buffer')
+  })
+})
--- a/packages/jellyfish-crypto/__tests__/aes256.test.ts
+++ b/packages/jellyfish-crypto/__tests__/aes256.test.ts
@@ -7,11 +7,6 @@ const passphrase = Buffer.from('password', 'ascii')
 describe('Aes256', () => {
   let encrypted: Buffer
 
-  it('', () => {
-    encrypted = Aes256.encrypt(passphrase, privateKey)
-    expect(encrypted.length).toStrictEqual(48) // [16 bytes salt, 32 bytes cipher]
-  })
-
   it('encrypt', () => {
     encrypted = Aes256.encrypt(passphrase, privateKey)
     expect(encrypted.length).toStrictEqual(48) // [16 bytes salt, 32 bytes cipher]
--- a/packages/jellyfish-crypto/__tests__/aes256.test.ts
+++ b/packages/jellyfish-crypto/__tests__/aes256.test.ts
@@ -13,13 +13,13 @@ describe('Aes256', () => {
   })
 
   it('encrypt - should be able to overwrite cipher iv for better security measure', () => {
-    const iv = Buffer.from('0102030405060708090a0b0c0d0e0f10', 'hex')
+    const iv = (): Buffer => Buffer.from('0102030405060708090a0b0c0d0e0f10', 'hex')
     encrypted = AES256.encrypt(passphrase, privateKey, iv)
     expect(encrypted.length).toStrictEqual(48) // [16 bytes salt, 32 bytes cipher]
   })
 
   it('encrypt - should reject non 16 bytes long iv', () => {
-    const iv = Buffer.from('0102030405060708090a0b0c0d0e0f', 'hex')
+    const iv = (): Buffer => Buffer.from('0102030405060708090a0b0c0d0e0f', 'hex')
     expect(() => AES256.encrypt(passphrase, privateKey, iv)).toThrow('Initialization vector must be 16 bytes long')
   })
 
--- a/packages/jellyfish-crypto/__tests__/aes256.test.ts
+++ b/packages/jellyfish-crypto/__tests__/aes256.test.ts
@@ -1,10 +1,11 @@
+import randomBytes from 'randombytes'
 import { AES256 } from '../src'
 
-const raw = 'e9873d79c6d87dc0fb6a5778633389f4e93213303da61f20bd67fc233aa33262'
-const privateKey = Buffer.from(raw, 'hex')
-const passphrase = Buffer.from('password', 'ascii')
-
 describe('Aes256', () => {
+  const raw = 'e9873d79c6d87dc0fb6a5778633389f4e93213303da61f20bd67fc233aa33262'
+  const privateKey = Buffer.from(raw, 'hex')
+  const passphrase = Buffer.from('password', 'ascii')
+
   let encrypted: Buffer
 
   it('encrypt', () => {
@@ -40,4 +41,21 @@ describe('Aes256', () => {
       AES256.decrypt(passphrase, invalidData)
     }).toThrow('Provided "encrypted" must decrypt to a non-empty string or buffer')
   })
+})
+
+it('AES256 with 1000 random vectors - should be consistent', () => {
+  const sampleData = randomBytes(1000)
+  const passphrase = randomBytes(60)
+
+  const encrypted = AES256.encrypt(passphrase, sampleData)
+  const decrypted = AES256.decrypt(passphrase, encrypted)
+
+  const encryptedAgain = AES256.encrypt(passphrase, decrypted)
+  const decryptedAgain = AES256.decrypt(passphrase, encryptedAgain)
+
+  // encrypted value are salted, it will be never the same
+  expect(encryptedAgain.toString('hex')).not.toStrictEqual(encrypted.toString('hex'))
+  // decrypted value are raw, recoverable using passphrase
+  expect(decryptedAgain.toString('hex')).toStrictEqual(decrypted.toString('hex'))
+  expect(decryptedAgain.toString('hex')).toStrictEqual(sampleData.toString('hex'))
 })
--- a/packages/jellyfish-crypto/__tests__/aes256.test.ts
+++ b/packages/jellyfish-crypto/__tests__/aes256.test.ts
@@ -12,6 +12,17 @@ describe('Aes256', () => {
     expect(encrypted.length).toStrictEqual(48) // [16 bytes salt, 32 bytes cipher]
   })
 
+  it('encrypt - should be able to overwrite cipher iv for better security measure', () => {
+    const iv = Buffer.from('0102030405060708090a0b0c0d0e0f10', 'hex')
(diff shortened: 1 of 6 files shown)


packages/jellyfish-crypto/src/aes256.ts
```diff
@@ -0,0 +1,51 @@
...
+ * @param {Buffer} encrypted The encrypted message to be decrypted.
+ * @returns {Buffer} The original plain-text message or buffer.
+ */
+function decrypt (key: Buffer, encrypted: Buffer): Buffer {
+  if (encrypted.length < 17) {
+    throw new Error('Provided "encrypted" must decrypt to a non-empty string or buffer')
+  }
+
+  const sha256 = crypto.createHash('sha256')
+  sha256.update(key)
```
fuxingloh, line 36: We should use the SHA256 within this package that is RN and Browser friendly.

fuxingloh: Lesser to manage also.

packages/jellyfish-crypto/src/aes256.ts
```diff
@@ -0,0 +1,51 @@
...
+ * Encrypt a clear-text message using AES-256 plus a random Initialization Vector.
+ * @param {Buffer} key  A passphrase of any length to used to generate a symmetric session key.
+ * @param {Buffer} data  The clear-text message or buffer to be encrypted.
+ * @returns {Buffer}
+ */
+function encrypt (key: Buffer, data: Buffer): Buffer {
+  const sha256 = crypto.createHash('sha256')
+  sha256.update(key)
+
+  const iv = crypto.randomBytes(16)
```
fuxingloh, line 18: We need the ability to provide a `randomBytes()` implementation. Using JavaScript crypto.randomBytes is too dangerous in production.

Similar to how it's done in:

https://github.com/DeFiCh/jellyfish/blob/e2d7a6faff74104bd5521d46eff155c13a4545e1/packages/jellyfish-wallet-mnemonic/src/mnemonic.ts#L57-L61

packages/jellyfish-crypto/src/aes256.ts
```diff
@@ -0,0 +1,51 @@
...
+  const cipher = crypto.createCipheriv(CIPHER_ALGORITHM, sha256.digest(), iv)
```
fuxingloh, line 19: I was looking through this list, it doesn't seem like it's supported

https://www.npmjs.com/package/react-native-crypto

Not sure if I am missing something.

packages/jellyfish-crypto/src/aes256.ts
```diff
@@ -0,0 +1,51 @@
...
+  const iv = encrypted.slice(0, 16)
+  const decipher = crypto.createDecipheriv(CIPHER_ALGORITHM, sha256.digest(), iv)
+
+  const ciphertext = encrypted.slice(16)
+  const deciphered = decipher.update(ciphertext)
+  const decipherFinal = decipher.final()
+  return Buffer.concat([deciphered, decipherFinal])
+}
+
+export const Aes256 = {
```
fuxingloh, line 48: Should call it `AES256` since AES is short for a larger word.

fuxingloh (changes requested):

packages/jellyfish-crypto/src/aes256.ts
ivan-zynesis: replace native `crypto` usage with browserify one. Not sure it will be usable for RN either.

packages/jellyfish-crypto/src/aes256.ts
ivan-zynesis: I guess, generateMnemonic rng is called multiple times. For this, I think allow custom `iv` buffer arg instead of `rng()` should be sufficient.

fuxingloh: It's called once, the reason we need to provide is that JavaScript Crypto is not using secured random bytes. You need to inject it from native.

fuxingloh (changes requested):

packages/jellyfish-crypto/src/aes256.ts
```diff
@@ -0,0 +1,58 @@
...
+const CIPHER_ALGORITHM = 'aes-256-ctr'
+
+/**
+ * Encrypt a clear-text message using AES-256 plus a random Initialization Vector.
+ * @param {Buffer} key  A passphrase of any length to used to generate a symmetric session key.
+ * @param {Buffer} data  The clear-text message or buffer to be encrypted.
+ * @param {() => Buffer} initVector Initialization vector provider, then to create AES cipher, default using `crypto` or browserify `random-bytes` package
+ * @returns {Buffer}
+ */
+function encrypt (key: Buffer, data: Buffer, initVector?: () => Buffer): Buffer {
```
fuxingloh, line 17:
```suggestion
function encrypt (key: Buffer, data: Buffer, rng?: (numOfBytes: number) => Buffer): Buffer {
```

Should be providing a rng, not providing an initVector

packages/jellyfish-crypto/src/aes256.ts
```diff
@@ -0,0 +1,58 @@
...
+  const sha256 = createHash('sha256')
+  sha256.update(key)
```
fuxingloh, line 19: please use https://github.com/DeFiCh/jellyfish/blob/3235f31ea108f7668da8aa8eeeda610429965b77/packages/jellyfish-crypto/src/hash.ts#L13

packages/jellyfish-crypto/src/aes256.ts
```diff
@@ -0,0 +1,58 @@
...
+ * Implementation reference: https://github.com/JamesMGreene/node-aes256
+ */
+import createHash from 'create-hash'
+import randomBytes from 'randombytes'
+import aes from 'browserify-aes'
+
+const CIPHER_ALGORITHM = 'aes-256-ctr'
+
+/**
+ * Encrypt a clear-text message using AES-256 plus a random Initialization Vector.
```
fuxingloh, line 11:
```suggestion
 * Encrypt a clear-text message using AES-256 plus a random Initialization Vector.
 * @see https://github.com/JamesMGreene/node-aes256
 
```

packages/jellyfish-crypto/src/aes256.ts
```diff
@@ -0,0 +1,58 @@
+/**
+ * Implementation reference: https://github.com/JamesMGreene/node-aes256
+ */
```
fuxingloh, line 3:
```suggestion
```

packages/jellyfish-crypto/src/aes256.ts
```diff
@@ -0,0 +1,58 @@
...
+    throw new Error('Initialization vector must be 16 bytes long')
+  }
+  const cipher = aes.createCipheriv(CIPHER_ALGORITHM, sha256.digest(), iv)
+  const ciphertext = cipher.update(data)
+  return Buffer.concat([iv, ciphertext, cipher.final()])
+}
+
+/**
+ * Decrypt an encrypted message back to clear-text using AES-256 plus a random Initialization Vector.
+ * @param {Buffer} key A passphrase of any length to used to generate a symmetric session key.
```
fuxingloh, line 33:
```suggestion
/**
 * Decrypt an encrypted message back to clear-text using AES-256 plus a random Initialization Vector.
 * @see https://github.com/JamesMGreene/node-aes256
 *
 * @param {Buffer} key A passphrase of any length to used to generate a symmetric session key.
```

packages/jellyfish-crypto/src/aes256.ts
```diff
@@ -0,0 +1,58 @@
...
+import randomBytes from 'randombytes'
+import aes from 'browserify-aes'
+
+const CIPHER_ALGORITHM = 'aes-256-ctr'
+
+/**
+ * Encrypt a clear-text message using AES-256 plus a random Initialization Vector.
+ * @param {Buffer} key  A passphrase of any length to used to generate a symmetric session key.
+ * @param {Buffer} data  The clear-text message or buffer to be encrypted.
+ * @param {() => Buffer} initVector Initialization vector provider, then to create AES cipher, default using `crypto` or browserify `random-bytes` package
```
fuxingloh, line 14: need to change this description also

packages/jellyfish-crypto/src/aes256.ts
fuxingloh: You need to be extremely careful on how you implement nonce, don't give consumers of the library any chance of giving you wrong and unsafe vector.

packages/jellyfish-crypto/src/aes256.ts
```diff
@@ -0,0 +1,58 @@
...
+function decrypt (key: Buffer, encrypted: Buffer): Buffer {
+  if (encrypted.length < 17) {
+    throw new Error('Provided "encrypted" must decrypt to a non-empty string or buffer')
+  }
+
+  const sha256 = createHash('sha256')
+  sha256.update(key)
+
+  // Initialization Vector
+  const iv = encrypted.slice(0, 16)
```
fuxingloh, line 46:
```suggestion
  const initVector = encrypted.slice(0, 16)
```

If you comment might be better to write out the full name.

packages/jellyfish-crypto/src/aes256.ts
```diff
@@ -0,0 +1,58 @@
...
+ * @param {Buffer} encrypted The encrypted message to be decrypted.
+ * @returns {Buffer} The original plain-text message or buffer.
+ */
+function decrypt (key: Buffer, encrypted: Buffer): Buffer {
+  if (encrypted.length < 17) {
+    throw new Error('Provided "encrypted" must decrypt to a non-empty string or buffer')
+  }
+
+  const sha256 = createHash('sha256')
+  sha256.update(key)
```
fuxingloh, line 43: use SHA256

fuxingloh (changes requested):

packages/jellyfish-crypto/__tests__/aes256.test.ts
```diff
@@ -0,0 +1,43 @@
...
+    expect(invalid.toString('hex')).not.toStrictEqual(raw)
+  })
+
+  it('decrypt - data too short, insufficient length to include salt', () => {
+    const invalidData = Buffer.alloc(16)
+    expect(() => {
+      AES256.decrypt(passphrase, invalidData)
+    }).toThrow('Provided "encrypted" must decrypt to a non-empty string or buffer')
+  })
+})
```
fuxingloh, line 43: Add 1 more test `describe('AES256 with 1000 random vectors')` where you do a round trip encrypt decrypt encrypt expecting the same results always

packages/jellyfish-crypto/__tests__/aes256.test.ts
```diff
@@ -0,0 +1,43 @@
+import { AES256 } from '../src'
+
+const raw = 'e9873d79c6d87dc0fb6a5778633389f4e93213303da61f20bd67fc233aa33262'
+const privateKey = Buffer.from(raw, 'hex')
+const passphrase = Buffer.from('password', 'ascii')
```
fuxingloh, line 5: then move this into the `Aes256` describe block

fuxingloh (changes requested):
1 more test for safety

packages/jellyfish-crypto/__tests__/aes256.test.ts
```diff
@@ -0,0 +1,61 @@
...
+  const decrypted = AES256.decrypt(passphrase, encrypted)
+
+  const encryptedAgain = AES256.encrypt(passphrase, decrypted)
+  const decryptedAgain = AES256.decrypt(passphrase, encryptedAgain)
+
+  // encrypted value are salted, it will be never the same
+  expect(encryptedAgain.toString('hex')).not.toStrictEqual(encrypted.toString('hex'))
+  // decrypted value are raw, recoverable using passphrase
+  expect(decryptedAgain.toString('hex')).toStrictEqual(decrypted.toString('hex'))
+  expect(decryptedAgain.toString('hex')).toStrictEqual(sampleData.toString('hex'))
```
fuxingloh, line 60: fori 1000

fuxingloh (changes requested):

packages/jellyfish-crypto/__tests__/aes256.test.ts
ivan-zynesis: Oops misunderstood what you meant

fuxingloh (approved):

packages/jellyfish-crypto/src/aes256.ts
```diff
@@ -0,0 +1,54 @@
...
+ * @see https://github.com/JamesMGreene/node-aes256
+ *
+ * @param {Buffer} key  A passphrase of any length to used to generate a symmetric session key.
+ * @param {Buffer} data  The clear-text message or buffer to be encrypted.
+ * @param {(lengthOfBytes: number) => Buffer} rng Initialization vector generator, default using `crypto` or browserify `random-bytes` package
+ * @returns {Buffer}
+ */
+function encrypt (key: Buffer, data: Buffer, rng?: (lengthOfBytes: number) => Buffer): Buffer {
+  const sha256 = SHA256(key)
+  const initVector = rng === undefined ? randomBytes(16) : rng(16)
```
thedoublejay, line 18: It can be `rng == null` to cover both null and undefined

thedoublejay (changes requested):

packages/jellyfish-crypto/src/aes256.ts
thedoublejay: Ok settled on the previous PR

thedoublejay (approved):

PR #332, opened 2021-06-02
Add Filter by DfTxType in AccountRPCs

### LINKED ISSUE #252: Filter `txtype` Parameter of `listAccountHistory()` doesn´t seem to filter AccountHistory

siradji:

/kind fix

Implemented `TxType` enum to fix #252

Fixes #

### DIFF
--- a/packages/jellyfish-api-core/src/category/account.ts
+++ b/packages/jellyfish-api-core/src/category/account.ts
@@ -378,12 +378,12 @@ export interface AccountHistoryOptions {
   depth?: number
   no_rewards?: boolean
   token?: string
-  txtype?: TxType | string
+  txtype?: TxType
   limit?: number
 }
 
 export interface AccountHistoryCountOptions {
   token?: string
-  txtype?: TxType | string
+  txtype?: TxType
   no_rewards?: boolean
 }
--- a/packages/jellyfish-api-core/src/category/account.ts
+++ b/packages/jellyfish-api-core/src/category/account.ts
@@ -302,7 +302,7 @@ export class Account {
    * @param {AccountHistoryCountOptions} [options]
    * @param {boolean} [options.no_rewards] Filter out rewards
    * @param {string} [options.token] Filter by token
-   * @param {TxType} [options.txtype] Filter by transaction type, supported letter from 'CRTMNnpuslrUbBG'
+   * @param {TxType} [options.txtype] Filter by transaction type. See TxType.
    * @return {Promise<number>} count of account history
    */
   async historyCount (
--- a/packages/jellyfish-api-core/src/category/account.ts
+++ b/packages/jellyfish-api-core/src/category/account.ts
@@ -31,7 +31,7 @@ export enum TxType {
   AUTO_AUTH_PREP = 'A'
 }
 
-type AccountRegexType = `${string}@${string}`
+type AccountRegexType = `${number}@${string}`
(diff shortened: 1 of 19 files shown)


packages/jellyfish-api-core/src/category/account.ts
```diff
@@ -348,7 +348,7 @@ export interface AccountHistoryOptions {
   depth?: number
   no_rewards?: boolean
   token?: string
-  txtype?: string
+  txtype?: TxType | string
```
canonbrother, line 351: i think just `TxType` because it should be restricted, don't open for `string`

canonbrother, line 351: i think just TxType because it should be restricted, we don't open for `string`

canonbrother (changes requested):

packages/jellyfish-api-core/src/category/account.ts
siradji: It's fixed.

website/docs/jellyfish/api/account.md
```diff
@@ -125,6 +125,26 @@ enum OwnerType {
   ALL = "all"
 }
 
+enum TxType {
```
ivan-zynesis, line 128:
```suggestion
enum DfTxType {
```

ivan-zynesis (changes requested):
Would like to name it specifically apply for custom tx only

packages/jellyfish-api-core/src/category/account.ts
```diff
@@ -285,7 +285,7 @@ export class Account {
    * @param {AccountHistoryCountOptions} [options]
    * @param {boolean} [options.no_rewards] Filter out rewards
    * @param {string} [options.token] Filter by token
-   * @param {TxType | string} [options.txtype] Filter by transaction type, supported letter from 'CRTMNnpuslrUbBG'
+   * @param {TxType} [options.txtype] Filter by transaction type, supported letter from 'CRTMNnpuslrUbBG'
```
fuxingloh, line 288:
```suggestion
   * @param {TxType} [options.txtype] Filter by transaction type. See TxType.
```

For consistency

fuxingloh (dismissed):

Issue #202, opened 2021-05-10
`DeFiCh/ain:1.7.x` RPCs implementation


fuxingloh:
Similar to https://github.com/DeFiCh/jellyfish/issues/48 This tracks all the RPC methods and categories being implemented in `jellyfish-api-core`. Please tag update this if you want to work on a feature.

~~As a final docker build is not produced please use the latest 1.7.x docker build for your `testcontainers`, https://github.com/DeFiCh/jellyfish/pull/184 for more information.~~

### `rpc_icxorderbook.cpp`
- [x] icx_createorder @surangap #398 
- [x] icx_makeoffer @surangap #399 
- [x] icx_closeoffer @surangap #400 
- [x] icx_submitdfchtlc @surangap #404 
- [x] icx_submitexthtlc @surangap #405 
- [x] icx_claimdfchtlc @surangap #406 
- [x] icx_closeorder @surangap #408 
- [x] icx_getorder @surangap #401 
- [x] icx_listorders @surangap #402 
- [x] icx_listhtlcs @surangap #407 

### `rpc_oracles.cpp`
- [x] appointoracle @jingyi2811 #336 
- [x] removeoracle @jingyi2811 #339 
- [x] updateoracle @jingyi2811  #341 
- [x] setoracledata @jingyi2811 #346
- [x] getoracledata @jingyi2811 #347
- [x] listoracles @jingyi2811 #345
- [x] listlatestrawprices @jingyi2811 #352 
- [x] getprice @jingyi2811 #355
- [x] listprices @jingyi2811 #356

### `spv_rpc.cpp`
- [x] spv_createhtlc @Jouzo #511
- [x] spv_claimhtlc @Jouzo #521
- [x] spv_gethtlcseed @Jouzo  #527
- [x] spv_refundhtlc @Jouzo #529
- [x] spv_listhtlcoutputs @Jouzo #581
- [x] spv_decodehtlcscript @Jouzo #513
- [x] spv_getnewaddress @Jouzo #495 
- [x] spv_getaddresspubkey @Jouzo #496
- [x] spv_listreceivedbyaddress @Jouzo #498
- [x] spv_sendtoaddress @Jouzo  #506
- [x] spv_refundhtlcall @cwkang1998 #1324

surangap:
## `src/masternodes/`
### `rpc_accounts.cpp`
- [ ] listburnhistory
- [ ] getburninfo

### `rpc_icxorderbook.cpp`
[...]

PR #341, opened 2021-06-03
Add updateOracle rpc

jingyi2811:

/kind feature

Add updateOracle rpc

Fixes part of #202

### DIFF
--- a/packages/jellyfish-api-core/src/category/oracle.ts
+++ b/packages/jellyfish-api-core/src/category/oracle.ts
@@ -1,0 +1,46 @@
+import { ApiClient } from '../.'
+
+/**
+ * Oracle RPCs for DeFi Blockchain
+ */
+export class Oracle {
+  private readonly client: ApiClient
+
+  constructor (client: ApiClient) {
+    this.client = client
+  }
+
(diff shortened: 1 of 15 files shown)


packages/jellyfish-api-core/src/category/oracle.ts
```diff
@@ -20,7 +20,7 @@ export class Oracle {
    * @param {UTXO[]} [options.utxos = []]
    * @param {string} [options.utxos.txid]
    * @param {number} [options.utxos.vout]
-   * @return {Promise<string>} oracleid
+   * @return {Promise<string>} txn id for txn created to appoint oracle
```
fuxingloh, line 23:
```suggestion
   * @return {Promise<string>} oracleid, also the txn id for txn created to appoint oracle
```

fuxingloh (changes requested):
some clarification about txn id, txn id is also used for identifying oracle data.

This allows us to identify masternode as well, that's how you masternode id get generated.

fuxingloh (approved):

canonbrother (approved):

PR #352, opened 2021-06-08
Add listLatestRawPrices rpc

jingyi2811:

/kind feature

Add listLatestRawPrices rpc

Fixes part of #202


packages/jellyfish-api-core/__tests__/category/oracle/listLatestRawPrices.test.ts
```diff
@@ -0,0 +1,64 @@
...
+        timestamp: timestamp2,
+        rawprice: 1.0,
+        state: 'expired'
+      }
+    ])
+  })
+
+  // it('should not listLatestRawPrices if token and currency do not exists', async () => {
+  //   const data = await client.oracle.listLatestRawPrices( { token: 'TESLA', currency: 'USD' })
+  //   expect(data.length).toStrictEqual(0)
```
monstrobishi, line 62: This commented out, either uncomment or delete

packages/jellyfish-api-core/__tests__/category/oracle/listLatestRawPrices.test.ts
```diff
@@ -0,0 +1,64 @@
...
+        oracleid: oracleid2,
+        weightage: 2,
+        timestamp: timestamp2,
+        rawprice: 1.0,
+        state: 'expired'
+      }
+    ])
+  })
+
+  // it('should not listLatestRawPrices if token and currency do not exists', async () => {
```
monstrobishi, line 60: do not exist

Remove the s

monstrobishi (changes requested):

jingyi2811:
Still doing. Haven't finished

packages/jellyfish-api-core/src/category/oracle.ts
```diff
@@ -84,6 +84,20 @@ export class Oracle {
...
+
+  /**
+   * Returns latest raw price updates from oracles.
+   *
+   * @param {OraclePriceFeed} priceFeed
+   * @return {Promise<OracleRawPrice[]>}
+   */
+  async listLatestRawPrices (priceFeed?: OraclePriceFeed): Promise<OracleRawPrice[]> {
+    if (priceFeed != null) {
+      return await this.client.call('listlatestrawprices', [priceFeed], 'number')
```
jingyi2811, line 96: Is there a better way to write this? 
If priceFeed value = empty object or null or undefined, this call will return an error.

packages/jellyfish-api-core/__tests__/category/oracle/listLatestRawPrices.test.ts
```diff
@@ -0,0 +1,253 @@
...
+      ]
+    )
+
+    await container.call('removeoracle', [oracleid1])
+    await container.call('removeoracle', [oracleid2])
+
+    await container.generate(1)
+  })
+
+  it('should listLatestRawPrices with various timestamps', async () => {
```
jingyi2811, line 105: This test is created to check when is an oracle price expired

packages/jellyfish-api-core/__tests__/category/oracle/listLatestRawPrices.test.ts
```diff
@@ -0,0 +1,253 @@
...
+    const result1 = data.find(element => element.oracleid === oracleid1)
+
+    expect(result1).toStrictEqual(
+      {
+        priceFeeds: priceFeed,
+        oracleid: oracleid1,
+        weightage: 1,
+        timestamp: timestamp1,
+        rawprice: 0.5,
+        state: 'live'
```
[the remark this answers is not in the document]
canonbrother: use `enum` for state

packages/jellyfish-api-core/src/category/oracle.ts
```diff
@@ -84,6 +84,20 @@ export class Oracle {
...
+
+  /**
+   * Returns latest raw price updates from oracles.
+   *
+   * @param {OraclePriceFeed} [priceFeed]
+   * @return {Promise<OracleRawPrice[]>}
+   */
+  async listLatestRawPrices (priceFeed?: OraclePriceFeed): Promise<OracleRawPrice[]> {
+    if (priceFeed != null) {
+      return await this.client.call('listlatestrawprices', [priceFeed], 'number')
```
monstrobishi, line 96: Use bignumber precision here not number

monstrobishi (changes requested):

monstrobishi (dismissed):

packages/jellyfish-api-core/src/category/oracle.ts
```diff
@@ -93,6 +98,20 @@ export class Oracle {
...
+   * @param {OraclePriceFeed} [priceFeed]
+   * @return {Promise<OracleRawPrice[]>}
+   */
+  async listLatestRawPrices (priceFeed?: OraclePriceFeed): Promise<OracleRawPrice[]> {
+    if (priceFeed != null) {
+      return await this.client.call('listlatestrawprices', [priceFeed], 'bignumber')
+    } else {
+      return await this.client.call('listlatestrawprices', [], 'bignumber')
+    }
+  }
```
canonbrother, line 114: just leave it undefined
```
async listLatestRawPrices (priceFeed?: OraclePriceFeed): Promise<OracleRawPrice[]> {
    return await this.client.call('listlatestrawprices', [priceFeed], 'bignumber')
  }
```

packages/jellyfish-api-core/__tests__/category/oracle/listLatestRawPrices.test.ts
```diff
@@ -0,0 +1,255 @@
...
+
+    // NOTE(jingyi2811): Set to 1 hour before.
+    const timestamp5 = Math.floor(new Date().getTime() / 1000) - 3600
+    const prices5 = [{ tokenAmount: '2.5@APPLE', currency: 'EUR' }]
+    await container.call('setoracledata', [oracleid5, timestamp5, prices5])
+
+    await container.generate(1)
+
+    // NOTE(jingyi2811): Pagination is not supported.
+    const data = await client.oracle.listLatestRawPrices()
```
canonbrother, line 146: better specific the type here else you won't know the type in interface is wrong
the resp has conflicts with interface type
```
weightage: new BigNumber(1),
timestamp: new BigNumber(timestamp1),
rawprice: new BigNumber(0.5),
```

canonbrother (changes requested):

packages/jellyfish-api-core/src/category/oracle.ts
jingyi2811: Not working. If priceFeed value = empty object or null or undefined, this call will return an error.

packages/jellyfish-api-core/__tests__/category/oracle/listLatestRawPrices.test.ts
jingyi2811: Ya you r right

packages/jellyfish-api-core/src/category/oracle.ts
canonbrother: hmm.. then
```
const params = priceFeed !== undefined ? [priceFeed] : []
return await this.client.call('listlatestrawprices', params, 'bignumber')
```

surangap (approved):

monstrobishi (dismissed):

packages/jellyfish-api-core/__tests__/category/oracle/listLatestRawPrices.test.ts
jingyi2811: My mistake: I have changed from number to BigNumber
https://github.com/DeFiCh/jellyfish/pull/352/files#diff-6f61e5d83c18cd13b1f2689748cc14f1301832e6d3a45bc8bff234d79e5feb44R142

I spent an hour adding the type here and check the test result:=> Even though there is number in the interface, bignumber still returned.

packages/jellyfish-api-core/src/category/oracle.ts
jingyi2811: This is working. Okay, I adopt your suggestion.

monstrobishi (approved):

packages/jellyfish-api-core/src/category/oracle.ts
```diff
@@ -119,6 +136,15 @@ export interface OracleData {
   weightage: number
 }
 
+export interface OracleRawPrice{
```
canonbrother, line 139:
```suggestion
export interface OracleRawPrice {
```

packages/jellyfish-api-core/__tests__/category/oracle/listLatestRawPrices.test.ts
```diff
@@ -0,0 +1,255 @@
...
+
+    expect(data.length).toStrictEqual(5)
+
+    const result1 = data.find(element => element.oracleid === oracleid1)
```
canonbrother, line 150: reply to https://github.com/DeFiCh/jellyfish/pull/352/files#r649039153
the reason: https://github.com/DeFiCh/jellyfish/pull/352/files#r649039153
you need specific type here as well

canonbrother (changes requested):

packages/jellyfish-api-core/__tests__/category/oracle/listLatestRawPrices.test.ts
jingyi2811: I did this yesterday but it was not working. Tested it again just now, and still not working.

packages/jellyfish-api-core/src/category/oracle.ts
jingyi2811: Fixed

packages/jellyfish-api-core/src/category/oracle.ts
```diff
@@ -93,6 +99,17 @@ export class Oracle {
...
   }
+
+  /**
+   * Returns latest raw price updates from oracles.
+   *
+   * @param {OraclePriceFeed} [priceFeed]
+   * @return {Promise<OracleRawPrice[]>}
+   */
+  async listLatestRawPrices (priceFeed?: OraclePriceFeed): Promise<OracleRawPrice[]> {
+    const params = priceFeed !== undefined ? [priceFeed] : []
```
thedoublejay, line 110: better to use priceFeed != null to cover null and undefined

thedoublejay (changes requested):

packages/jellyfish-api-core/src/category/oracle.ts
jingyi2811: Ok

canonbrother: @thedoublejay hmm.. actually `null` doesn't undefined, they are diff 
the best way is actually checking the condition by `!priceFeed` which covers price != undefined && price != null && price != ""
but ts-standard will auto convert `!priceFeed` to `priceFeed !== undefined`
i'm fine with !== null as well
will making another advance ts-lint and formatting to standardized it

thedoublejay: @canonbrother `null` covers `undefined` as long as you don't use strict equality. 
image

thedoublejay (approved):

packages/jellyfish-api-core/src/category/oracle.ts
fuxingloh: > @canonbrother `null` covers `undefined` as long as you don't use strict equality.
> image
> image

I hate this about JavaScript, everything should be strict by default. This allows mistakes to happen because you didn't check whether you used strict or not e.g. `!=` vs `!==`, everyone assumes it's `!==` ultimately to correct misconception. 

I prefer explicit over implicit if you can use `!==` or `===` always use that. I rather see it cleaner than having to guess it. Better to code defensive than openly.

https://github.com/DeFiCh/jellyfish/blob/main/CONTRIBUTING.md#explicit-over-implicit

/cc @jingyi2811 @siradji @aikchun @monstrobishi @canonbrother @izzycsy @thedoublejay @surangap @ivan-zynesis

ivan-zynesis: > @canonbrother `null` covers `undefined` as long as you don't use strict equality.
> image
> image

Generally speaking, that's not true. For eg, just one row declaration
```
let foo: Foo | null
```
Broken JS may cover `foo != null` but
```
switch (foo) should be covered for all 3 cases
```

website/docs/jellyfish/api/oracle.md
```diff
@@ -143,3 +143,32 @@ interface oracle {
...
+  rawprice: BigNumber
+  weightage: BigNumber
+  state: OracleRawPriceState
+  timestamp: BigNumber
+}
+
+interface OraclePriceFeed {
+  token: string
+  currency: string
+}
```
fuxingloh, line 173: OraclePriceFeed should go first.

packages/jellyfish-api-core/__tests__/category/oracle/listLatestRawPrices.test.ts
```diff
@@ -0,0 +1,255 @@
...
+
+    // NOTE(jingyi2811): Pagination is not supported.
+    const data = await client.oracle.listLatestRawPrices()
+
+    const result1 = data.filter(element => element.oracleid === oracleid1)
+
+    expect(result1).toStrictEqual(
+      [
+        {
+          priceFeeds: priceFeed1[0],
```
fuxingloh, line 60:
```suggestion
          priceFeeds: { token: 'APPLE', currency: 'EUR' },
```
I prefer explicitly rather than referencing.

packages/jellyfish-api-core/__tests__/category/oracle/listLatestRawPrices.test.ts
```diff
@@ -0,0 +1,255 @@
...
+      ]
+    )
+
+    await container.call('removeoracle', [oracleid1])
+    await container.call('removeoracle', [oracleid2])
+
+    await container.generate(1)
+  })
+
+  it('should listLatestRawPrices with various timestamps', async () => {
```
fuxingloh, line 107: I would prefer you write 5 tests with 1 describe block instead of putting it all together, easier to read.

packages/jellyfish-api-core/__tests__/category/oracle/listLatestRawPrices.test.ts
```diff
@@ -0,0 +1,255 @@
...
+    await container.call('removeoracle', [oracleid1])
+    await container.call('removeoracle', [oracleid2])
+    await container.call('removeoracle', [oracleid3])
+    await container.call('removeoracle', [oracleid4])
+    await container.call('removeoracle', [oracleid5])
+
+    await container.generate(1)
+  })
+
+  it('should listLatestRawPrices with empty array if there is no oracle appointed', async () => {
```
fuxingloh, line 219: you should test 1 more condition when you set 2 feed. and listLatestRawPrices with 1 feed making sure only 1 feed is returned.

fuxingloh (approved):
A few improvements you should make, won't block your progress now. Tackle in another PR, make it a standalone PR, don't mix with the others.

packages/jellyfish-api-core/__tests__/category/oracle/listLatestRawPrices.test.ts
```diff
@@ -0,0 +1,255 @@
...
+    await container.waitForReady()
+    await container.waitForWalletCoinbaseMaturity()
+  })
+
+  afterAll(async () => {
+    await container.stop()
+  })
+
+  it('should listLatestRawPrices', async () => {
+    const priceFeed1 = [
```
ivan-zynesis, line 21:
```suggestion
    const APPLE = [
```
QOL improvement, I think name with asset symbols ease reading a lot.

packages/jellyfish-api-core/__tests__/category/oracle/listLatestRawPrices.test.ts
```diff
@@ -0,0 +1,255 @@
...
+
+    // NOTE(jingyi2811): Set to 1 second after 1 hour before.
+    const timestamp4 = Math.floor(new Date().getTime() / 1000) - 3599
+    const prices4 = [{ tokenAmount: '2.0@APPLE', currency: 'EUR' }]
+    await container.call('setoracledata', [oracleid4, timestamp4, prices4])
+
+    // NOTE(jingyi2811): Set to 1 hour before.
+    const timestamp5 = Math.floor(new Date().getTime() / 1000) - 3600
+    const prices5 = [{ tokenAmount: '2.5@APPLE', currency: 'EUR' }]
+    await container.call('setoracledata', [oracleid5, timestamp5, prices5])
```
ivan-zynesis, line 141: Interesting, rpc call actually can set the price for such an "old" timestamp? @monstrobishi

surangap (approved):

packages/jellyfish-api-core/__tests__/category/oracle/listLatestRawPrices.test.ts
jingyi2811: It was like this
- test with the current timestamp
- test with 3599 seconds after the current timestamp
- test with 3600 seconds after the current timestamp
- test with 3599 seconds before the current timestamp
- test with 3600 seconds before the current timestamp

If I run all these tests 5 times, probably all would pass. However, it doesn't guarantee it will pass if I run all the tests 6th time.

Alter discussing with Ahmed a few days ago, he advised me to change the magic number 3600 to 4200.

In the new code, I have reduced the no of tests to 2.
- test with 4200 seconds after the current timestamp
- test with 4200 seconds before the current timestamp

PR #359, opened 2021-06-09
`jellyfish-transaction-signature`

ivan-zynesis:
#### What kind of PR is this?:

/kind refactor

#### What this PR does / why we need it:
To remove crypto dependencies from jellyfish-transaction
to become pure transaction object-buffer (de)composer

#### Which issue(s) does this PR fixes?:

Fixes #132 
Meant to close 132 once merged

#### Additional comments?:

PR #369, opened 2021-06-11
`jellyfish-transaction-signature`

ivan-zynesis:

/kind feature

To remove crypto dependencies from jellyfish-transaction
to become pure transaction object-buffer (de)composer

Fixes #

#### Additional comments?:
fixing build issue, to replace #359


.github/CODEOWNERS
```diff
@@ -25,6 +25,7 @@
 /packages/jellyfish-transaction/src/script/defi/            @fuxingloh @canonbrother @ivan-zynesis @monstrobishi
 
 /packages/jellyfish-transaction-builder/        @fuxingloh @ivan-zynesis @monstrobishi
+/packages/jellyfish-transaction-signature/      @ivan-zynesis
```
fuxingloh, line 28:
```suggestion
/packages/jellyfish-transaction-signature/      @fuxingloh @ivan-zynesis
```

too dangerous to go solo for this

fuxingloh (changes requested):
I will do a proper review/check after AES get merged in

.github/CODEOWNERS
ivan-zynesis: Oh I forgot it affect auto merge.

packages/jellyfish-transaction/package.json
```diff
@@ -38,7 +38,6 @@
     "bignumber.js": "^9.0.1"
   },
   "dependencies": {
-    "@defichain/jellyfish-crypto": "0.0.0",
     "smart-buffer": "^4.1.0"
   },
   "devDependencies": {
```
fuxingloh, line 43: I think we can remove all `devDependencies`, since we end up not using it anyway.

packages/jellyfish-transaction-signature/package.json
```diff
@@ -0,0 +1,50 @@
...
+  "dependencies": {
+    "@defichain/jellyfish-crypto": "0.0.0",
+    "@defichain/jellyfish-transaction": "0.0.0",
+    "smart-buffer": "^4.1.0"
+  },
+  "devDependencies": {
+    "@defichain/jellyfish-api-core": "0.0.0",
+    "@defichain/jellyfish-api-jsonrpc": "0.0.0",
+    "@defichain/testcontainers": "0.0.0"
+  }
```
fuxingloh, line 49: Same for this, I don't think we using it.

fuxingloh (changes requested):
looks good, we should remove devDeps

packages/jellyfish-transaction-signature/package.json
ivan-zynesis: They are used in unit test, wouldn't that affect?

fuxingloh: I don't think these 3 are used in the signature package. They are only used in builder tests?

ivan-zynesis: Yes, you're right. I was on false impression, as they were copied from jellyfish-transaction. Indeed, they were needed in the first place.

fuxingloh (approved):

PR #395, opened 2021-06-21
use bignumber precision on gettoken and listtokens

### LINKED ISSUE #392: TokenInfo `minted` and `limit` should be BigNumber
#### What happened:

TokenInfo in `token.getToken` minted and limited should return BigNumber as they are precision sensitive numbers. It currently returns a number now, due to jellyfish-json limitations we might need to return all as BigNumber.

/area jellyfish-api-core
/triage accepted
/priority urgent-now
/assign @canonbrother

canonbrother:

/kind fix

Token `minted` and `limit` are actually bignumber

Fixes #392

#### Additional comments?:
Due to the `gettoken` return contains randomId,  the exact precision mapping will not work correctly
so instead
`{ '0': { limit: 'bignumber', minted: 'bignumber' }`
then just
` 'bignumber' `


fuxingloh (dismissed):

packages/jellyfish-api-core/__tests__/category/token/createToken.test.ts
```diff
@@ -1,6 +1,7 @@
 import { MasterNodeRegTestContainer } from '@defichain/testcontainers'
 import { ContainerAdapterClient } from '../../container_adapter_client'
 import waitForExpect from 'wait-for-expect'
+import { BigNumber } from '@defichain/jellyfish-api-core'
```
jingyi2811, line 4: `import BigNumber from 'bignumber.js'`
Ahmed, Bella, AikChun, and I have changed to this way to import BigNumber

packages/jellyfish-api-core/__tests__/category/token/listTokens.test.ts
```diff
@@ -1,6 +1,7 @@
 import { MasterNodeRegTestContainer, RegTestContainer } from '@defichain/testcontainers'
 import { ContainerAdapterClient } from '../../container_adapter_client'
 import waitForExpect from 'wait-for-expect'
+import { BigNumber } from '@defichain/jellyfish-api-core'
```
jingyi2811, line 4: `import BigNumber from 'bignumber.js'`
Ahmed, Bella, AikChun, and I have changed to this way to import BigNumber

fuxingloh (approved):

jingyi2811 (approved):

PR #408, opened 2021-06-21
Interface`icx_closeorder` RPC 

surangap:
#### What kind of PR is this?:

/kind feature 

#### What this PR does / why we need it:
This PR will interface `icx_closeorder` RPC to jellyfish
#### Which issue(s) does this PR fixes?:

Fixes #

#### Additional comments?:
depends on #407

PR #409, opened 2021-06-21
Add complex tests for ICX

surangap:

/kind feature 

Adds some complex tests for ICX

Fixes #

#### Additional comments?:
depends on #408


surangap:
not many complex tests for this test suite at the moment. some of them have been relocated to relevant RPCs.
Also one could think of many scenarios that we can test ICX. It's better to have core functional testing on the C++ side as much as possible. But will add more complex scenarios if required along the way.

fuxingloh (approved):

Issue #292, opened 2021-05-26
test account history in `jellyfish-api-core` extensively


fuxingloh:
#### What would you like to be added:

Account history RPCs should be tested extensively.

#### Why is this needed:

Currently, the test within Account History is very limited. It only checks the soundness of the RPC does not test it end to end with different setups. We need complete/full e2e testing on account history to prevent regression.

fuxingloh:
@aikchun you need to work on this soon before moving to working on explorer related work

aikchun:
@fuxingloh What are some of the various setups and scenarios that should be tested?

And this is for `listAccountHistory` and `accountHistoryCount` rpcs?

fuxingloh:
1. Pool Rewards
2. Pool Swap
3. AccountToAccount
4. UTXO To Account
5. Account to UTXO
6. Pool Enter
7. Pool Exit

aikchun:
>     1. Pool Rewards
> 
>     2. Pool Swap
> 
>     3. AccountToAccount
> 
>     4. UTXO To Account
> 
>     5. Account to UTXO
> 
>     6. Pool Enter
> 
>     7. Pool Exit

If I infer the code correctly, for Pool Enter, this happens when user calls `addpoolliquidity` rpc right?
Pool Exit -> `removepoolliquidity` rpc
Pool Swap -> `poolswap` rpc

I'm not too sure when Pool Rewards happen.

fuxingloh:
> > ```
> > 1. Pool Rewards
> > 
> > 2. Pool Swap
> > 
> > 3. AccountToAccount
> > 
> > 4. UTXO To Account
> > 
> > 5. Account to UTXO
> > 
> > 6. Pool Enter
> > 
> > 7. Pool Exit
> > ```
> 
> If I infer the code correctly, for Pool Enter, this happens when user calls `addpoolliquidity` rpc right?
> Pool Exit -> `removepoolliquidity` rpc
> Pool Swap -> `poolswap` rpc
> 
> I'm not too sure when Pool Rewards happen.

Pool Reward happens at every block gen, you need to allocate pool reward to a pool via set governance.

PR #410, opened 2021-06-22
Added test for listAccountHistory for accountToAccount, utxosToAccount, accountToUtxos set up.

aikchun:

/kind chore

- We need to test account history for various set up to prevent regression. This PR tests for `listAccountHistory` rpc for `accountToAccount`

Fixes part of #292 

#### Additional comments?:
This PR only covers account history tests for `accountToAccount`, `utxosToAccount` and `accountToUxtos` rpcs.
Other set up and test cases will be covered in a separate PR.


packages/jellyfish-api-core/__tests__/category/account/accountToAccountHistory.test.ts
```diff
@@ -0,0 +1,64 @@
+
+import { MasterNodeRegTestContainer } from '@defichain/testcontainers'
+import { ContainerAdapterClient } from '../../container_adapter_client'
+import { BalanceTransferPayload } from '../../../src/category/account'
+
+describe('listAccountHistory', () => {
```
fuxingloh, line 6: yup this works, but instead of having multiple files, you should scope them into describe blocks as one file instead.

canonbrother (approved):

fuxingloh (approved):

PR #414, opened 2021-06-23
removed project references and internal dev deps

fuxingloh:

/kind refactor

- Removed internal `devDependencies` as it can be resolved via module mapper in jest, hence not required anymore for testing.
- Removed individual project references to just use a global reference table, easier to maintain.
- Moved `@types/lossless-json` to be part of `dependencies` in `jellyfish-json`
- Fixed invalid `@defichain/jellyfish` tsconfig.json extend reference that didn't exist


canonbrother:
ohh.. like this working?? 
hmm.. what is the `global reference table`??
-> Removed individual project references to just use a global reference table, easier to maintain.

canonbrother (approved):

ivan-zynesis (approved):

PR #437, opened 2021-07-01
Fixed bad test in oracle RPC and cleaned up docs for consistency

fuxingloh:

/kind fix

```ts
const utxos = await container.call('listunspent', [1, 9999999, [address], true])
```

When `utxos.length === 0`, it will pick a foundation auth, as you pass in an empty UTXO. And since addresses are new without having a UTXO, it will succeed silently. This is not the expected case. UTXO test must use `fundAddress` for consistent test behavior.

~~Also renamed `oracleid` to `oracleId` for the parameter to keep camelCase consistency for params. `oracleid` is still used for data objects since ain uses that but we should use `oracleId` on params.~~


canonbrother:
ohh... i think i gonna rename those `poolpair` to `poolPair` on another PR later

canonbrother (dismissed):

fuxingloh:
> ohh... i think i gonna rename those `poolpair` to `poolPair` on another PR later

Actually, maybe it's fine.

fuxingloh:
> ohh... i think i gonna rename those `poolpair` to `poolPair` on another PR later

I will revert the rename since it's consistently inconsistent

canonbrother (approved):

jingyi2811 (approved):

PR #444, opened 2021-07-06
Added listBurnHistory RPC

Jouzo:
/kind feature

Adds listBurnHistory RPC

Fixes missing RPC implementation in #48


packages/jellyfish-api-core/__tests__/category/account/listBurnHistory.test.ts
```diff
@@ -0,0 +1,195 @@
...
+    expect(typeof history[0].blockHash).toStrictEqual('string')
+    expect(history[0].blockHash.length).toStrictEqual(64)
+    expect(history[0].type).toStrictEqual('CreateMasternode')
+    expect(history[0].txn).toStrictEqual(2)
+    expect(typeof history[0].txid).toStrictEqual('string')
+    expect(history[0].txid.length).toStrictEqual(64)
+    expect(typeof history[0].amounts).toStrictEqual('object')
+    expect(history[0].amounts.length).toStrictEqual(1)
+    expect(history[0].amounts[0]).toStrictEqual('1.00000000@DFI')
+  })
```
siradji, line 42: Tests are failing when i run them independently. Each test most be independent and must not rely on other tests  to pass. Make sure your tests are passing independently.

packages/jellyfish-api-core/src/category/account.ts
```diff
@@ -343,6 +344,25 @@ export class Account {
   async listCommunityBalances (): Promise<CommunityBalanceData> {
     return await this.client.call('listcommunitybalances', [], 'bignumber')
   }
+
+  /**
+   * Returns information about burn history
+   *
+   * @param {BurnHistoryOptions} [options]
+   * @param {number} [options.maxBlockHeight] Optional height to iterate from (down to genesis block), (default = chaintip).
```
siradji, line 352:
```suggestion
   * @param {number} [options.maxBlockHeight]  The block height to iterate from.
```

siradji (changes requested):

packages/jellyfish-api-core/__tests__/category/account/listBurnHistory.test.ts
fuxingloh: @Jouzo rmb to address this

Jouzo: Done, let me know if this new version is coherent with the rest

packages/jellyfish-api-core/src/category/account.ts
```diff
@@ -343,6 +344,25 @@ export class Account {
...
+
+  /**
+   * Returns information about burn history
+   *
+   * @param {BurnHistoryOptions} [options]
+   * @param {number} [options.maxBlockHeight]  The block height to iterate from.
+   * @param {number} [options.depth] Maximum depth, from the genesis block is the default
+   * @param {string} [options.token] Filter by token
+   * @param {DfTxType} [options.txtype] Filter by transaction type. See DfTxType.
+   * @param {number} [options.limit=100] Maximum number of records to return, 100 by default
```
canonbrother, line 356: can you add test for `maxBlockHeight`, `depth` and `limit`??

Jouzo: Sure, let me add that

packages/jellyfish-api-core/__tests__/category/account/listBurnHistory.test.ts
```diff
@@ -0,0 +1,261 @@
...
+
+  it('should listBurnHistory with limit 5', async () => {
+    const history = await client.account.listBurnHistory()
+    const historyWithLimit = await client.account.listBurnHistory({
+      limit: 5
+    })
+    expect(historyWithLimit.length).toStrictEqual(5)
+    expect(historyWithLimit.slice(0, 5)).toStrictEqual(history.slice(0, 5))
+  })
```
canonbrother, line 260: duplicated.. just need one

packages/jellyfish-api-core/__tests__/category/account/listBurnHistory.test.ts
```diff
@@ -0,0 +1,261 @@
...
+    expect(history.every(({ blockHeight }) => blockHeight <= maxBlockHeight)).toBeTruthy()
+  })
+
+  it('should listBurnHistory with maxBlockHeight 500', async () => {
+    const maxBlockHeight = 500
+    const history = await client.account.listBurnHistory({
+      maxBlockHeight
+    })
+    expect(history.every(({ blockHeight }) => blockHeight <= maxBlockHeight)).toBeTruthy()
+  })
```
canonbrother, line 222: test is not clear.. for this you need to assert the before after to compare the diff
eg:
```
historyBefore = await listBurnHistory()
assert(maxBlockHeight(historyBefore)).toStrictEqual(15)

historyAfter = await listBurnHistory({maxBlockHeight: 3})
assert(maxBlockHeight(historyAfter)).toStrictEqual(3)
```

canonbrother (changes requested):
we just need one example for the test on each filter
eg:
filter by
txtype - 1
token - 1
maxBlockHeight - 1
depth - 1
limit - 1

and here are BDD standard, trying to be story-like code which is helpful a lot

packages/jellyfish-api-core/__tests__/category/account/listBurnHistory.test.ts
```diff
@@ -0,0 +1,261 @@
...
+      expect(typeof burnHistory.blockHeight).toStrictEqual('number')
+      expect(typeof burnHistory.blockHash).toStrictEqual('string')
+      expect(typeof burnHistory.blockTime).toStrictEqual('number')
+      expect(typeof burnHistory.type).toStrictEqual('string')
+      expect(typeof burnHistory.txn).toStrictEqual('number')
+      expect(typeof burnHistory.txid).toStrictEqual('string')
+      expect(burnHistory.amounts.length).toBeGreaterThan(0)
+      expect(typeof burnHistory.amounts[0]).toStrictEqual('string')
+    }
+  })
```
canonbrother, line 91: no `let` so move to here
and just need one for general `list` test
```
const history = listBurnHistory()
for (...) {
   // general assertion here (eg: type)
   // blockhash, blocktime, txid, txn

   // distinct assertion here
   // type, owner, amounts
}
```

packages/jellyfish-api-core/__tests__/category/account/listBurnHistory.test.ts
```diff
@@ -0,0 +1,261 @@
...
+
+  it('should listBurnHistory with depth 0', async () => {
+    const depth = 0
+    const history = await client.account.listBurnHistory()
+    const depthHistory = await client.account.listBurnHistory({
+      depth
+    })
+    const maxBlockHeight = Math.max(...history.map(el => el.blockHeight)) // Get maxBlockHeight from history
+    expect(depthHistory.every(({ blockHeight }) => blockHeight >= maxBlockHeight - depth)).toBeTruthy()
+  })
```
canonbrother, line 232: hmm... TBH.. i don't quite understand the `depth` filter here

Jouzo: Ok let me make that clearer

packages/jellyfish-api-core/__tests__/category/account/listBurnHistory.test.ts
Jouzo: Got it, let me remove duplicate test and assertions

packages/jellyfish-api-core/__tests__/category/account/listBurnHistory.test.ts
```diff
@@ -0,0 +1,261 @@
...
+    expect(typeof history[6].amounts).toStrictEqual('object')
+    expect(history[6].amounts.length).toStrictEqual(1)
+    expect(history[6].amounts[0]).toStrictEqual('10.00000000@DFI')
+  })
+
+  it('should listBurnHistory with filter on txtype None', async () => {
+    const history = await client.account.listBurnHistory({
+      txtype: DfTxType.NONE
+    })
+    expect(history.every(({ type }) => type === 'None')).toBeTruthy()
```
jingyi2811, line 163: You need to assert the length of the history

packages/jellyfish-api-core/__tests__/category/account/listBurnHistory.test.ts
```diff
@@ -0,0 +1,261 @@
...
+  })
+
+  it('should listBurnHistory with filter on txtype UtxosToAccount', async () => {
+    const history = await client.account.listBurnHistory({
+      txtype: DfTxType.UTXOS_TO_ACCOUNT
+    })
+    expect(history.every(({ type }) => type === 'UtxosToAccount')).toBeTruthy()
```
jingyi2811, line 170: For those scenarios where the length of the history is 1, write something like this. You do not need to use `every`
```

    expect(history.length).toStrictEqual(1)
    expect(history[0].type).toStrictEqual('UtxosToAccount')
```

packages/jellyfish-api-core/__tests__/category/account/listBurnHistory.test.ts
```diff
@@ -0,0 +1,261 @@
...
+  it('should listBurnHistory with limit 1', async () => {
+    const history = await client.account.listBurnHistory()
+    const historyWithLimit = await client.account.listBurnHistory({
+      limit: 1
+    })
+    expect(historyWithLimit.length).toStrictEqual(1)
+    expect(historyWithLimit[0]).toStrictEqual(history[0])
+  })
+
+  it('should listBurnHistory with limit 5', async () => {
```
jingyi2811, line 253: You wrote 2 `it` for` limit 1 `and` limit 5`
Unless you predict they return a different **pattern** of results, you should keep 1 only. 
I would recommend you keep `limit 5`

packages/jellyfish-api-core/__tests__/category/account/listBurnHistory.test.ts
Jouzo: Done on 30ae17d, tell me if you find it clearer

packages/jellyfish-api-core/__tests__/category/account/listBurnHistory.test.ts
```diff
@@ -0,0 +1,222 @@
...
+    })
+    expect(history.every(({ amounts }) => amounts[0].includes('DFI'))).toBeTruthy()
+  })
+
+  it('should listBurnHistory with maxBlockHeight 110', async () => {
+    const history = await client.account.listBurnHistory()
+    const historyWithMaxBlockHeight = await client.account.listBurnHistory({
+      maxBlockHeight: 110
+    })
+    expect(Math.max(...history.map(el => el.blockHeight))).toStrictEqual(112)
```
ivan-zynesis, line 200: I wonder where is this 112 came from.
`waitForWalletCoinbaseMaturity` - 101
setupHistory, `generate(1)` x8
`createToken` has `generate(1)` x2

website/docs/jellyfish/api/account.md
```diff
@@ -216,7 +216,8 @@ enum DfTxType {
   CREATE_POOL_PAIR = 'p',
   UPDATE_POOL_PAIR = 'u',
   SET_GOV_VARIABLE = 'G',
-  AUTO_AUTH_PREP = 'A'
+  AUTO_AUTH_PREP = 'A',
+  NONE = '0'
```
ivan-zynesis, line 220: I am not sure if this is explicitly for burn history only. Is it a legit filter option for other account history listing since it is actually not a custom DfTx in the first place.

Jouzo: I don't fully understand the implication of your question. 
It's a valid value for filtering by txtype though

packages/jellyfish-api-core/__tests__/category/account/listBurnHistory.test.ts
Jouzo: It is not clear this way. I reverted back to a more logical test

packages/jellyfish-api-core/__tests__/category/account/listBurnHistory.test.ts
```diff
@@ -0,0 +1,224 @@
...
+    expect(history[2].owner).toStrictEqual(burnAddress)
+    expect(history[2].type).toStrictEqual('AccountToAccount')
+    expect(history[2].amounts.length).toStrictEqual(1)
+    expect(history[2].amounts[0]).toStrictEqual('1.00000000@DFI')
+  })
+
+  it('sixth transaction should be send to burn address with accounttoutxos', async () => {
+    const history = await client.account.listBurnHistory()
+    expect(history[1].owner).toStrictEqual(burnAddress)
+    expect(history[1].type).toStrictEqual('None')
```
jingyi2811, line 133: I was thinking why this is `None`. Should be `AccountToUtxos`

packages/jellyfish-api-core/__tests__/category/account/listBurnHistory.test.ts
```diff
@@ -0,0 +1,224 @@
...
+
+  const createHistory = async (): Promise<void> => {
+    // test masternode creation fee burn
+    const newAddress = await container.getNewAddress('', 'legacy')
+    await client.masternode.createMasternode(newAddress)
+    await container.generate(1)
+
+    // creates funded account address
+    fundedAddress = await container.getNewAddress()
+    await client.wallet.sendToAddress(fundedAddress, 1)
```
jingyi2811, line 31: I would add `await container.generate(1)` here too

packages/jellyfish-api-core/__tests__/category/account/listBurnHistory.test.ts
```diff
@@ -0,0 +1,224 @@
...
+    await container.call('utxostoaccount', [{ [fundedAddress]: '3@0' }])
+    await container.generate(1)
+
+    // burn token
+    await createToken(container, 'GOLD', { collateralAddress: fundedAddress })
```
jingyi2811, line 36: I would add `await container.generate(1)` after this line too

packages/jellyfish-api-core/__tests__/category/account/listBurnHistory.test.ts
```diff
@@ -0,0 +1,224 @@
...
+
+    // mint token and send token to burn address
+    await container.call('minttokens', ['100@GOLD'])
+    await container.generate(1)
+
+    await accountToAccount(container, 'GOLD', 100, {
+      from: fundedAddress, to: burnAddress
+    })
```
jingyi2811, line 44: I would add `await container.generate(1)` after this line too

packages/jellyfish-api-core/__tests__/category/account/listBurnHistory.test.ts
```diff
@@ -0,0 +1,224 @@
...
+
+    // utxostoaccount burn
+    await container.call('utxostoaccount', [{ [burnAddress]: '1@0' }])
+    await container.generate(1)
+
+    // send to burn address with accountToAccount
+    await accountToAccount(container, '0', 1, {
+      from: fundedAddress, to: burnAddress
+    })
```
jingyi2811, line 53: I would add `await container.generate(1)` after this line too

packages/jellyfish-api-core/__tests__/category/account/listBurnHistory.test.ts
```diff
@@ -0,0 +1,224 @@
...
+    expect(history.length).toStrictEqual(1)
+    expect(history[0].type).toStrictEqual('CreateToken')
+  })
+
+  it('should listBurnHistory with filter on token DFI', async () => {
+    const history = await client.account.listBurnHistory({
+      token: 'DFI'
+    })
+    expect(history.every(({ amounts }) => amounts[0].includes('DFI'))).toBeTruthy()
```
jingyi2811, line 190: Need to check history length too

website/docs/jellyfish/api/account.md
Jouzo: It is a valid DfTxType on the masternode side as can be seen in `ain` `src/masternodes/mn_checks.h`

packages/jellyfish-api-core/__tests__/category/account/listBurnHistory.test.ts
canonbrother: guess something wrong from the `ain`

canonbrother (commented):
that still have the duplicated tests but since its added i just ignore it
overall i'm good

Jouzo:
Are you talking about the filter by token type @canonbrother ?

canonbrother:
> Are you talking about the filter by token type @canonbrother ?

yes the `txtype` filter, both `txtype` filter tests are considered same condition

Jouzo:
you're right, let me fix it

packages/jellyfish-api-core/__tests__/category/account/listBurnHistory.test.ts
jingyi2811: I thought so too yesterday

packages/jellyfish-api-core/__tests__/category/account/listBurnHistory.test.ts
```diff
@@ -0,0 +1,222 @@
...
+    expect(history[4].type).toStrictEqual('AccountToAccount')
+    expect(history[4].amounts.length).toStrictEqual(1)
+    expect(history[4].amounts[0]).toStrictEqual('100.00000000@GOLD')
+  })
+
+  it('fourth transaction should be utxostoaccount burn', async () => {
+    const history = await client.account.listBurnHistory()
+    expect(history[3].owner).toStrictEqual(burnAddress)
+    expect(history[3].type).toStrictEqual('UtxosToAccount')
+    expect(typeof history[3].amounts).toStrictEqual('object')
```
jingyi2811, line 117: I think you may remove this line. These 2 is good enough.

```
expect(history[3].amounts.length).toStrictEqual(1)
expect(history[3].amounts[0]).toStrictEqual('1.00000000@DFI') 
```

packages/jellyfish-api-core/__tests__/category/account/listBurnHistory.test.ts
```diff
@@ -0,0 +1,222 @@
...
+    expect(history[6].txn).toStrictEqual(2)
+    expect(history[6].amounts.length).toStrictEqual(1)
+    expect(history[6].amounts[0]).toStrictEqual('1.00000000@DFI')
+  })
+
+  it('second transaction should be after burn token', async () => {
+    const history = await client.account.listBurnHistory()
+    expect(history[5].owner).toStrictEqual('6a19446654785404474f4c4404474f4c4408000000000000000007')
+    expect(history[5].type).toStrictEqual('CreateToken')
+    expect(history[5].txn).toStrictEqual(1)
```
jingyi2811, line 100: You can add this line to history[4], [3] .... [0]

packages/jellyfish-api-core/__tests__/category/account/listBurnHistory.test.ts
```diff
@@ -0,0 +1,222 @@
...
+    await container.start()
+    await container.waitForReady()
+    await container.waitForWalletCoinbaseMaturity()
+
+    await client.wallet.importPrivKey(burnAddressPrivateKey)
+
+    await createHistory()
+  })
+
+  const createHistory = async (): Promise<void> => {
```
jingyi2811, line 23: I would consider this as top level function.  See
https://github.com/DeFiCh/jellyfish/blob/main/CONTRIBUTING.md#top-level-functions-should-not-use-arrow-syntax

You may write like this
`async function createHistory(): Promise<void> {  `

which is the common way written by other developer

website/docs/jellyfish/api/account.md
```diff
@@ -334,3 +336,55 @@ interface CommunityBalanceData {
...
 }
 ```
+
+## listBurnHistory
+
+Returns information about burn history
+
+```ts title="client.account.listBurnHistory()"
+interface account {
+  listBurnHistory(
```
jingyi2811, line 346: If you look at other RPCs in the same file, there is always a space between` function name` and `(`
And, you wrote like this too in `account.ts`. See
Screenshot 2021-07-19 at 10 22 22 PM


Hence, change `listBurnHistory(` to `listBurnHistory (`

aikchun (commented):
For assertions that uses `.every`, since the return type is a boolean, we should probably use `.toBe(true)` or `.toStrictEqual(true)`, instead of `toBeTruthy` to be more definitive on what we're testing for.

jingyi2811 (approved):

packages/jellyfish-api-core/__tests__/category/account/listBurnHistory.test.ts
```diff
@@ -0,0 +1,197 @@
...
+    expect(history[5].amounts[0]).toStrictEqual('1.00000000@DFI')
+  })
+
+  it('third transaction should be send token to burn address', async () => {
+    const history = await client.account.listBurnHistory()
+    expect(history[4].owner).toStrictEqual(burnAddress)
+    expect(history[4].type).toStrictEqual('AccountToAccount')
+    expect(history[4].txn).toStrictEqual(1)
+    expect(history[4].amounts.length).toStrictEqual(1)
+    expect(history[4].amounts[0]).toStrictEqual('100.00000000@GOLD')
```
fuxingloh, line 111:
```suggestion
    expect(history[4]).toStrictEqual({
      owner: burnAddress,
      type: 'AccountToAccount',
      txn: 1,
      amounts: ['100.00000000@GOLD'],
    })
```

Might be better to write it like this

fuxingloh (approved):

Issue #294, opened 2021-05-26
`MasterNodeRegTestContainer` block gen feature should be controlled by testcontainers


fuxingloh:
#### What would you like to be added:

`MasterNodeRegTestContainer` block generation should be automated by testcontainers. `@defichain/testcontainers` should be created such that block generation can only be done by code within testcontainers. `gen=1` was set to automated the generation process on ain side.

#### Why is this needed:

Having automated block gen on the ain side is unpredictable and creates too many conditions for "side-effect" when testing on jellyfish. This effect is cascaded down to whale, playground and potentially wallet implementation.

#### Additional comments:

https://github.com/DeFiCh/playground uses its own block generation interval, we can follow that implementation reference.

https://github.com/DeFiCh/playground/blob/09b29fe1d47f48ac75283ecdda85c234c4bf94c2/src/module.playground/playground.block.ts#L13-L17

PR #450, opened 2021-07-08
Bump DeFiCh/ain to `v1.8.0`

fuxingloh:

/kind dependencies

Also, fixed https://github.com/DeFiCh/jellyfish/issues/294


fuxingloh:
@surangap is there an activation height for this?

If so, we need to add it here:
https://github.com/DeFiCh/jellyfish/blob/19fa41ed4c85b0ff6959ca29ee74d773f7257d88/packages/testcontainers/src/chains/reg_test_container/index.ts#L23

PR #451, opened 2021-07-09
Make prevout optional in elliptic pair

monstrobishi:

/kind chore

Most practical implementations don't want to pass in a prevout to obtain an elliptic pair, the existing interface makes it awkward for end users. E.g. have to pass in an empty prevout, etc 

Fixes #

### DIFF
--- a/packages/jellyfish-transaction-builder/__tests__/provider.mock.ts
+++ b/packages/jellyfish-transaction-builder/__tests__/provider.mock.ts
@@ -86,7 +86,7 @@ export class MockEllipticPairProvider implements EllipticPairProvider {
     }
   }
 
-  get (prevout: Prevout): EllipticPair {
+  get (prevout?: Prevout): EllipticPair {
     return this.ellipticPair
   }
 }
--- a/packages/jellyfish-transaction-builder/src/provider.ts
+++ b/packages/jellyfish-transaction-builder/src/provider.ts
@@ -40,5 +40,5 @@ export interface EllipticPairProvider {
    * @param {Prevout} prevout for the EllipticPair
    * @return {EllipticPair}
    */
-  get: (prevout: Prevout) => EllipticPair
+  get: (prevout?: Prevout) => EllipticPair
 }


monstrobishi:
Some more context, due to the existing interface the supposed "convenience enhancements" have left me with the following awkward code in Salmon after refactor

```
public async getChangeScript (): Promise<Script> {
    const ellipticPair = 
      this.builder.ellipticPairProvider.get({ 
        txid: '', 
        vout: 0,
        tokenId: 0,
        script: { stack: [] },
        value: new BigNumber(0)
      })

    return {
      stack: [
        OP_CODES.OP_0,
        OP_CODES.OP_PUSHDATA(HASH160(await ellipticPair.publicKey()), 'little')
      ]
    }
  }
```

Open to suggestions.

Passing elliptic pair separately from `P2WPKHTransactionBuilder` is not a valid suggestion as this opens up the possiblity of passing in different elliptic pairs, and without a convenient getter there's no way to validate this.

fuxingloh:
> Some more context, due to the existing interface the supposed "convenience enhancements" have left me with the following awkward code in Salmon after refactor
> 
> ```
> public async getChangeScript (): Promise<Script> {
>     const ellipticPair = 
>       this.builder.ellipticPairProvider.get({ 
>         txid: '', 
>         vout: 0,
>         tokenId: 0,
>         script: { stack: [] },
>         value: new BigNumber(0)
>       })
> 
>     return {
>       stack: [
>         OP_CODES.OP_0,
>         OP_CODES.OP_PUSHDATA(HASH160(await ellipticPair.publicKey()), 'little')
>       ]
>     }
>   }
> ```
> 
> Open to suggestions.
> 
> Passing elliptic pair separately from `P2WPKHTransactionBuilder` is not a valid suggestion as this opens up the possiblity of passing in different elliptic pairs, and without a convenient getter there's no way to validate this.

If you pass in a `WalletAccount` you can get change script via `account.getChangeScript()`

fuxingloh:
> Some more context, due to the existing interface the supposed "convenience enhancements" have left me with the following awkward code in Salmon after refactor
> 
> ```
> public async getChangeScript (): Promise<Script> {
>     const ellipticPair = 
>       this.builder.ellipticPairProvider.get({ 
>         txid: '', 
>         vout: 0,
>         tokenId: 0,
>         script: { stack: [] },
>         value: new BigNumber(0)
>       })
> 
>     return {
>       stack: [
>         OP_CODES.OP_0,
>         OP_CODES.OP_PUSHDATA(HASH160(await ellipticPair.publicKey()), 'little')
>       ]
>     }
>   }
> ```
> 
> Open to suggestions.
> 
> Passing elliptic pair separately from `P2WPKHTransactionBuilder` is not a valid suggestion as this opens up the possiblity of passing in different elliptic pairs, and without a convenient getter there's no way to validate this.

Take a look at this:

https://github.com/DeFiCh/wallet/blob/0c699bc5d44dd22b758c13f1637b2ccd05bd2d38/app/screens/AppNavigator/screens/Balances/ConvertScreen.tsx#L233-L259

monstrobishi:
Ok that does help my specific situation but I still think requiring prevout in `get` is not necessary

fuxingloh:
> Ok that does help my specific situation but I still think requiring prevout in `get` is not necessary

it's necessary if you are creating a transaction with utxo from different address, prevout allow you to locate the required utxo for spend

PR #453, opened 2021-07-09
`jellyfish-wallet` should provide non-hd `WalletNode` by default

fuxingloh:

/kind refactor

To allow non-HdWallet use cases and implementations as not all wallet implementations are HD, we should not enforce HD Nodes.

### DIFF
--- a/packages/jellyfish-wallet/src/wallet_account.ts
+++ b/packages/jellyfish-wallet/src/wallet_account.ts
@@ -1,5 +1,5 @@
-import { Script, OP_CODES, Transaction, Vout, TransactionSegWit } from '@defichain/jellyfish-transaction'
-import { WalletHdNode } from './wallet_hd_node'
+import { OP_CODES, Script, Transaction, TransactionSegWit, Vout } from '@defichain/jellyfish-transaction'
+import { WalletNode } from './wallet_node'
 import { Bech32, HASH160 } from '@defichain/jellyfish-crypto'
 import { Network } from '@defichain/jellyfish-network'
 import { DeFiAddress } from '@defichain/jellyfish-address'
@@ -11,9 +11,9 @@ import { DeFiAddress } from '@defichain/jellyfish-address'
  *
  * WalletAccount implementation uses NATIVE SEGWIT redeem script exclusively.
  */
-export abstract class WalletAccount implements WalletHdNode {
+export abstract class WalletAccount implements WalletNode {
   protected constructor (
-    private readonly hdNode: WalletHdNode,
+    private readonly hdNode: WalletNode,
     private readonly network: Network
   ) {
   }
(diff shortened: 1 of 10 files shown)


monstrobishi:
Since we are working on naming here, I'm not sure if I understand the term non-hd wallet node, as the term HD node is very specific to HD wallets?

packages/jellyfish-wallet/src/wallet_account.ts
```diff
@@ -85,8 +85,8 @@ export abstract class WalletAccount implements WalletHdNode {
 export interface WalletAccountProvider<T extends WalletAccount> {
 
   /**
-   * @param {WalletHdNode} hdNode of this wallet account
+   * @param {WalletNode} hdNode of this wallet account
    * @return WalletAccount
    */
-  provide: (hdNode: WalletHdNode) => T
+  provide: (hdNode: WalletNode) => T
```
ivan-zynesis, line 91:
```suggestion
  provide: (walletNode: WalletNode) => T
```

packages/jellyfish-wallet/src/wallet_account.ts
```diff
@@ -11,9 +11,9 @@ import { DeFiAddress } from '@defichain/jellyfish-address'
  *
  * WalletAccount implementation uses NATIVE SEGWIT redeem script exclusively.
  */
-export abstract class WalletAccount implements WalletHdNode {
+export abstract class WalletAccount implements WalletNode {
   protected constructor (
-    private readonly hdNode: WalletHdNode,
+    private readonly hdNode: WalletNode,
```
ivan-zynesis, line 16:
```suggestion
    private readonly walletNode: WalletNode,
```

monstrobishi:
https://arshbot.medium.com/hd-wallets-explained-from-high-level-to-nuts-and-bolts-9a41545f5b0

Quote:
Technically speaking, Hierarchical Deterministic Wallets are a tree structure where each **node** has an extended private and public key. Any **node** can have any number of children. Meaning a master key could regenerate 10 accounts in 10 different currencies each with a very high number of addresses. I chose 10 as an arbitrary number, the number could be as large as you’d like.

Imo it's either an HD Node or a private key pair, a node without the tree is no longer a node

monstrobishi:
Perhaps call it `WalletEllipticPair` or `ExtendedEllipticPair`

fuxingloh:
> Perhaps call it `WalletEllipticPair` or `ExtendedEllipticPair`

Make sense, let's call it `WalletEllipticPair`? can you make the change?

monstrobishi:
Sure

ivan-zynesis (changes requested):
Looking at the implementation again. `WalletHdNode `  do not really make sense, as how it can perform direct `getScript()` and `getAddress()`

multiple node derivation will be inside `HdWalletNodeProvider<WalletNode>`, maybe can just remove `WalletHdNode` entirely.

fuxingloh:
looks good, let's get this merged and we can create a release to sync whale

ivan-zynesis (approved):

PR #454, opened 2021-07-09
`jellyfish-wallet-classic`

### LINKED ISSUE #452: `jellyfish-wallet-classic` single keypair provider

A single keypair provider similar to the mock here: 

https://github.com/DeFiCh/jellyfish/blob/main/packages/jellyfish-wallet/__tests__/node.mock.ts

#### Why is this needed:

Useful for light implementations and e2e tests

monstrobishi:

/kind feature

WalletClassic implements a simple, single elliptic pair wallet.

Fixes #452

### DIFF
--- a/packages/jellyfish-wallet-classic/src/classic.ts
+++ b/packages/jellyfish-wallet-classic/src/classic.ts
@@ -1,0 +1,65 @@
+import { WalletEllipticPair } from '@defichain/jellyfish-wallet'
+import { EllipticPair } from '@defichain/jellyfish-crypto'
+import {
+  Transaction,
+  Vout,
+  TransactionSegWit
+} from '@defichain/jellyfish-transaction'
+import { SignInputOption, TransactionSigner } from '@defichain/jellyfish-transaction-signature'
+
+/**
+ * WalletClassic extends WalletEllipticPair with a simple classic implementation.
+ *
+ * Single elliptic pair wallet.
+ */
+export class WalletClassic implements WalletEllipticPair {
+  constructor (public readonly ellipticPair: EllipticPair) {
+  }
+
+  /**
+   * @return {Promise<Buffer>} compressed public key
+   */
+  async publicKey (): Promise<Buffer> {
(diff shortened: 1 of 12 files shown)


ivan-zynesis (commented):
Overall looks good. But I am hesitant the import of `jellyfish-wallet` for the WalletEllipticPair interface, this package suppose to much lighter.

I think this classic package user are usually backend microservice which directly accessing their corresponding private key, I am thinking if we can just implement `EllipticPair` instead.

monstrobishi:
@ivan-zynesis I agree but this would require some refactoring of Whale, as Whale requires a WalletEllipticPair interface for it's providers

fuxingloh:
> @ivan-zynesis I agree but this would require some refactoring of Whale, as Whale requires a WalletEllipticPair interface for it's providers

Yeah, it requires `jellyfish-wallet` as `jellyfish-wallet-classic` is a jellyfish wallet provider.

packages/jellyfish-wallet-classic/src/classic.ts
```diff
@@ -0,0 +1,65 @@
...
+  TransactionSegWit
+} from '@defichain/jellyfish-transaction'
+import { SignInputOption, TransactionSigner } from '@defichain/jellyfish-transaction-signature'
+
+/**
+ * WalletClassic extends WalletEllipticPair with a simple classic implementation.
+ *
+ * Single elliptic pair wallet.
+ */
+export class WalletClassic implements WalletEllipticPair {
```
fuxingloh, line 15: Possibly, in the future we can include these static helper method to make it easier to create Classic Instance

1. `WalletClassic.fromWIF(string)`
2. `WalletClassic.fromPrivkey(Buffer)`

fuxingloh (approved):

ivan-zynesis (approved):

fuxingloh:
it just tick over to 97.5, we have 98% coverage now on the badge :)

PR #476, opened 2021-07-21
`MasterNodeRegTestContainer` block gen feature should be controlled by testcontainers

### LINKED ISSUE #294: `MasterNodeRegTestContainer` block gen feature should be controlled by testcontainers

fuxingloh:

/kind feature

Fixes #294 

MasterNodeRegTestContainer block generation should be automated by testcontainers. @defichain/testcontainers should be created such that block generation can only be done by code within testcontainers. gen=1 was set to automated the generation process on ain side.

#### Why is this needed:
Having automated block gen on the ain side is unpredictable and creates too many conditions for "side-effect" when testing on jellyfish. This effect is cascaded down to whale, playground and potentially wallet implementation.


packages/testcontainers/src/chains/reg_test_container/masternode.ts
```diff
@@ -119,7 +122,7 @@ export class MasterNodeRegTestContainer extends RegTestContainer {
       }
       await this.generate(1)
       return false
-    }, timeout, 1)
+    }, timeout, 100)
```
ivan-zynesis, line 125: Possible to be lower? So our test/ci can run faster.

fuxingloh: Not possible, because only 1 block can be mined per second. 100ms loop allowance allow it to mine forward if within consensus time limit.

ivan-zynesis (approved):

siradji (approved):

jingyi2811 (approved):

surangap (approved):

PR #507, opened 2021-07-27
EncryptedProviderData encrypt mnemonic words instead of privKey

### LINKED ISSUE #503: ability to get a copy of mnemonic seed on encrypted wallet
Ability to get a copy of mnemonic seed on encrypted wallet. 
#### Why is this needed:
Linked to this issue https://github.com/DeFiCh/wallet/issues/301

fuxingloh:

/kind fix

- `HdNodeProvider.data` and `HdNodeProvider.options` should be `public readonly` for external read
- EncryptedProviderData should encode encrypted words instead of private key for mnemonic recovery

Fixes #503

#### Additional comments?:

- [ ] @ivan-zynesis how do to encrypt even buffer length properly?

### DIFF
--- a/packages/jellyfish-wallet-encrypted/__tests__/bip32.test.ts
+++ b/packages/jellyfish-wallet-encrypted/__tests__/bip32.test.ts
@@ -44,6 +44,22 @@ const prevout: Vout = {
   tokenId: 0x00
 }
 
+it('should be able to access provider.data and provider.options', async () => {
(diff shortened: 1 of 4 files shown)


fuxingloh:
> EncryptedProvider should be simple, we should only decrypt `xpriv` keeping the implementation clean. As this is how it's supposed to be. Closed as suggested by @ivan-zynesis

Issue #107, opened 2021-04-12
DeFi OP_CODES implementation mapping


fuxingloh:
> This tracks all the OP_CODES class and their categories to be implemented in jellyfish-transaction. DeFi blockchain custom transaction will be tracked in another issue.

```
packages/jellyfish-transaction/src/script/*
```

## `data.ts`
- [x] 0x01-0x4b @fuxingloh #106
- [x] OP_PUSHDATA1 @fuxingloh #106
- [x] OP_PUSHDATA2 @fuxingloh #106
- [x] OP_PUSHDATA4 @fuxingloh #106

## `constants.ts`
- [x] OP_0 @fuxingloh #106
- [x] OP_FALSE @fuxingloh #106
- [x] OP_1NEGATE @jingyi2811 #198
- [x] OP_RESERVED @jingyi2811 #217 
- [x] OP_1 @jingyi2811 #200
- [x] OP_TRUE @jingyi2811 #200
- [x] OP_2 @jingyi2811 #200
- [x] OP_3 @jingyi2811 #200
- [x] OP_4 @jingyi2811 #200
- [x] OP_5 @jingyi2811 #200
- [x] OP_6 @jingyi2811 #200
- [x] OP_7 @jingyi2811 #200
- [x] OP_8 @jingyi2811 #200
- [x] OP_9 @jingyi2811 #200
- [x] OP_10 @jingyi2811 #200
- [x] OP_11 @jingyi2811 #200
- [x] OP_12 @jingyi2811 #200
- [x] OP_13 @jingyi2811 #200
- [x] OP_14 @jingyi2811 #200
- [x] OP_15 @jingyi2811 #200
- [x] OP_16 @jingyi2811 #200

## `control.ts` 
- [x] OP_NOP = 0x61 @siradji #333 
- [x] OP_VER = 0x62 @siradji #333 
- [x] OP_IF = 0x63 @siradji #333 
- [x] OP_NOTIF = 0x64 @siradji #333 
- [x] OP_VERIF = 0x65 @siradji #333 
- [x] OP_VERNOTIF = 0x66 @siradji #333 
- [x] OP_ELSE = 0x67 @siradji #333 
- [x] OP_ENDIF = 0x68 @siradji #333 
- [x] OP_VERIFY = 0x69 @siradji #333 
- [x] OP_RETURN @fuxingloh #106

## `stack.ts`
- [x] OP_TOALTSTACK = 0x6b @aikchun #389 
[...]

PR #512, opened 2021-07-28
Add arithmetic opcodes

jingyi2811:
What kind of PR is this?:
/kind feature

Adds arithmetic opcodes mentioned in #107

Fixes missing opcodes in #107

Additional comments?:


packages/jellyfish-transaction/src/script/index.ts
```diff
@@ -1,11 +1,12 @@
-export * from './bitwise'
```
jingyi2811, line 1: Reorganize this file like order in https://github.com/DeFiCh/jellyfish/issues/107

packages/jellyfish-transaction/src/script/mapping.ts
```diff
@@ -6,10 +6,11 @@ import { OP_PUSHDATA } from './data'
 import { OP_DEFI_TX } from './dftx'
```
jingyi2811, line 6: Reorganize imports like order in https://github.com/DeFiCh/jellyfish/issues/107

packages/jellyfish-transaction/src/script/index.ts
jingyi2811: It was good. Now it is not in correct order, probably is the result of GIT merge

fuxingloh (approved):

canonbrother (approved):

PR #523, opened 2021-07-30
Added `restart` to MNRegContainer to fix `generatetoaddress`

surangap:

/kind fix

restart functionality to MasterNodeRegTestContainer. This allows to restart containers in the middle of the test with additional defid options. This allows fixing the problem with generating blocks to newly created masternodes.

Fixes #

#### Additional comments?:
depends https://github.com/DeFiCh/ain/pull/646

### DIFF
--- a/packages/testcontainers/__tests__/containerRestart.test.ts
+++ b/packages/testcontainers/__tests__/containerRestart.test.ts
@@ -1,0 +1,44 @@
+import { MasterNodeRegTestContainer, StartOptions } from '../src'
+
+class CustomMasterNodeRegTestContainer extends MasterNodeRegTestContainer {
+  // NOTE(surangap): defid command line args has priority over args from defi.conf
+  protected getCmd (opts: StartOptions): string[] {
+    const cmd = super.getCmd(opts).filter(cmd => cmd !== '-dummypos=1') // remove -dummypos=1
+    return [
+      ...cmd,
+      '-dummypos=0'
+    ]
+  }
+}
+
+describe('container restart', () => {
+  const container = new CustomMasterNodeRegTestContainer()
+
+  beforeEach(async () => {
+    await container.start()
+    await container.waitForReady()
+    await container.waitForWalletCoinbaseMaturity()
+  })
+
+  afterEach(async () => {
+    await container.stop()
+  })
+
+  it('should create a masternode and mint some blocks', async () => {
+    const masternodeCount1 = await container.call('getactivemasternodecount', [])
+    expect(masternodeCount1).toBeLessThan(2)
+
+    const address = await container.getNewAddress('', 'legacy')
+    await container.call('createmasternode', [address])
+    await container.generate(1)
+
+    await container.restart(['masternode_operator=' + address])
+    await container.generate(20)
+
+    // generate blocks for new masternode address
+    await container.generate(20, address)
+
+    const masternodeCount2 = await container.call('getactivemasternodecount', [])
+    expect(masternodeCount2).toBeGreaterThan(1)
+  })
+})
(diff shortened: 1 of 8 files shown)


packages/testcontainers/src/chains/defid_container.ts
```diff
@@ -83,6 +83,13 @@ export abstract class DeFiDContainer extends DockerContainer {
     await this.container.start()
   }
 
+  async restart (additionalStartOptions: string[]): Promise<void> {
+    // NOTE(surangap)
+    // implement this.container?.restart() here.
+    // for some reason i could not work it out. even though the container is restrted, new command is not passed.
+    // it was the old command running again. If we could make this out, args can be directly passed with restart.
```
canonbrother, line 90: seems restart will remain the inherited filesystem state, perhaps just stop and start and name it as `stopStart` function??

packages/testcontainers/src/chains/reg_test_container/masternode.ts
```diff
@@ -75,6 +75,23 @@ export class MasterNodeRegTestContainer extends RegTestContainer {
...
+    fileContents += '\n'
+
+    await this.exec({
+      Cmd: ['bash', '-c', `echo "${fileContents}" > ~/.defi/defi.conf`]
+    })
+
+    await this.container?.stop()
+    await this.container?.start()
+    await super.waitForReady(25000)
+  }
```
canonbrother, line 93: hmm.. perhaps
```
async stopStart (additional) {
  super.stopStart(additional)
  super.waitForReady()
}

async stopStart (additional) {
  let fileContents = additionalStartOptions.join('\n')
  fileContents += '\n'

  await this.exec({
    Cmd: ['bash', '-c', `echo "${fileContents}" > ~/.defi/defi.conf`]
  })
  
  this.container.stop()
  this.container.start()
}
```

packages/testcontainers/__tests__/containerRestart.test.ts
```diff
@@ -0,0 +1,44 @@
...
+    await container.restart(['masternode_operator=' + address])
+    await container.generate(20)
+
+    // generate blocks for new masternode address
+    await container.generate(20, address)
+
+    const masternodeCount2 = await container.call('getactivemasternodecount', [])
+    expect(masternodeCount2).toBeGreaterThan(1)
+  })
+})
```
Jouzo, line 44: Can you test that masternode1.mintedBlocks > 0 and masternode2.mintedBlocks > 0 ?

packages/testcontainers/src/chains/defid_container.ts
surangap: yes, it will keep the data.
the requirement is to restart the container(old data intact) with a new CMD. but even `start()` does work with a non-empty `options`. so the same behaviour for `restart()`. the issue is either in `Dockerode` or related to docker APIs
tried a few workarounds without success. needs more debugging.
I like `stopStart `. kinda gives the idea of old data intact.

packages/testcontainers/src/chains/reg_test_container/masternode.ts
surangap: yes. the same could move up for now.

Jouzo (approved):

fuxingloh:
- instead of `stopStart`, rename to `setDeFiConf` since that's what it does
- added `restart()` instead of due to above refactor
- `start()` will now automatically `waitForReady()` since it's always called after `start()`
- Increase default timeout to `20000ms`
- skipping `container.generate(2000)` test in HTLC, it takes forever

fuxingloh (approved):

Issue #488, opened 2021-07-23
`DeFiCh/ain:2.x.x` Loan RPCs implementation


fuxingloh:
Similar to https://github.com/DeFiCh/jellyfish/issues/48, this tracks all the RPC methods and categories being implemented in `jellyfish-api-core`. Please tag and update this if you are working on a feature.

As a final docker build is not produced please use the latest `1.9.x` docker builds for your `testcontainers`, https://github.com/DeFiCh/jellyfish/pull/184 for more information on how to set custom image doing development.

### `rpc_loan.cpp`
- [x] createLoanScheme @jingyi2811 #520 
- [x] updateLoanScheme @jingyi2811 #530 
- [x] destroyLoanScheme @jingyi2811  #526 
- [x] setDefaultLoanScheme @jingyi2811  #554 
- [x] listLoanSchemes @jingyi2811  #524 
- [x] getLoanScheme @jingyi2811 #638 
- [x] setCollateralToken @jingyi2811  #537
- [x] getCollateralToken @jingyi2811  #546
- [x] listCollateralTokens @jingyi2811  #551
- [x] setLoanToken @jingyi2811  #568 
- [x] updateLoanToken @jingyi2811  #570 
- [x] getloantoken @surangap #758 
- [x] listLoanTokens @jingyi2811  #571 
- [x] takeLoan @canonbrother #661 
- [x] getLoanInfo @ivan-zynesis ~~#654~~ update: DeFiChain almost reworked the RPC, opened PR is invalid
- [x] loanPayBack @canonbrother #699 
- [x] getInterest @canonbrother #678
- [x] ~~getCollateralPrice~~

### `rpc_vault.cpp`
- [x] createVault @surangap  #553 
- [x] getVault  @surangap #569 
- [x] listVaults  @surangap #552 
- [x] updateVault @jingyi2811 #698 
- [x] depositToVault @canonbrother #661 
- [x] withdrawFromVault @ivan-zynesis #741  
- [x] closevault @jingyi2811 #730 
- [x] auctionBid @canonbrother #677 
- [x] listAuctions @jingyi2811 #695
- [x] listAuctionHistory @jingyi2811 #787 
- [ ] listVaultHistory @jingyi2811 #834
- [x] estimateloan @cwkang1998 #1347
- [x] estimatecollateral @cwkang1998 #1350 
- [x] estimatevault @cwkang1998 #1354 

### `rpc_oracles.cpp`
[...]

PR #526, opened 2021-07-30
Add destroyLoanScheme rpc

jingyi2811:

/kind feature

Add destroyLoanScheme rpc

Fixes part of #488


packages/jellyfish-api-core/__tests__/category/loan/destroyLoanScheme.test.ts
```diff
@@ -0,0 +1,163 @@
...
+
+describe('Loan', () => {
+  const container = new LoanMasterNodeRegTestContainer()
+  const testing = Testing.create(container)
+
+  beforeAll(async () => {
+    await testing.container.start()
+    await testing.container.waitForWalletCoinbaseMaturity()
+
+    // NOTE(jingyi2811): Default scheme
```
canonbrother, line 14: `NOTE(name)` is quite abuse for me.. the use case of *NOTE* should be something require us to NOTE
```suggestion
    // Default scheme
```

canonbrother: i add on the practice of **NOTE**
the use case:
i . abnormal approach: note the reason to reader why doing the logic on this way
ii. reference: something is required to understand before read this
...

canonbrother (commented):
overall good, just the `annotation`

packages/jellyfish-api-core/src/category/loan.ts
```diff
@@ -0,0 +1,37 @@
...
+
+  /**
+   * Destroys a loan scheme.
+   *
+   * @param {string} id Unique identifier of the loan scheme, max 8 chars
+   * @param {number} [activateAfterBlock] Block height at which new changes take effect
+   * @param {DeleteLoanSchemeOptions} [options]
+   * @param {UTXO[]} [options.utxos = []] Specific UTXOs to spend
+   * @param {string} [options.utxos.txid] Transaction Id
+   * @param {number} [options.utxos.vout] Output number
```
surangap, line 21:
```suggestion
   * @param {string} options.utxos.txid Transaction Id
   * @param {number} options.utxos.vout Output number
```
not optional within `UTXO`

packages/jellyfish-api-core/src/category/loan.ts
```diff
@@ -0,0 +1,37 @@
...
+   * @return {Promise<string>} Hex string of the transaction
+   */
+  async destroyLoanScheme (id: string, activateAfterBlock?: number, options: DeleteLoanSchemeOptions = {}): Promise<string> {
```
surangap, line 24: why not use `utxos: UTXO[]` directly  instead of `options: DeleteLoanSchemeOptions` here. can avoid additional  interface.
```suggestion
  async destroyLoanScheme (id: string, activateAfterBlock?: number, utxos: UTXO[] = []): Promise<string> {
```

surangap (changes requested):
request for some refactor.

packages/jellyfish-api-core/src/category/loan.ts
Jouzo: See guideline at : https://github.com/DeFiCh/jellyfish/blob/main/CONTRIBUTING.md#exported-functions-max-2-args-put-the-rest-into-an-options-object

Jouzo: Agree that creating an interface just for utxos might create more confusion

jingyi2811: I wrote what you suggested here a few months ago, But Fuxing said it is better to stick with  third parameter is always an option

fuxingloh (approved):
looks good!

PR #527, opened 2021-07-30
Added SPV getHtlcSeed RPC

Jouzo:

/kind feature

Adds SPV getHtlcSeed RPC

Fixes part of #202


surangap (approved):
try to add a multinode setup if that feature is working. It Will give more clarity and would be closer to the real world scenario. anw we should add some more real-world scenario tests(separately) combining both ICX and SPV RPCs

Jouzo:
> looks good.
> try to add a multinode setup if that feature is working. It Will give more clarity and would be closer to the real world scenario. anw we should add some more real-world scenario tests(separately) combining both ICX and SPV RPCs

I was thinking about adding some complex test after finishing all those PRs like https://github.com/DeFiCh/jellyfish/blob/main/packages/jellyfish-api-core/__tests__/category/icxorderbook/complexTests.test.ts

Jouzo:
Multinode setup got merged in. I will add the tests to this PR

packages/jellyfish-api-core/__tests__/category/spv/getHtlcSeed.test.ts
```diff
@@ -0,0 +1,79 @@
...
+    await group.get(0).call('spv_fundaddress', [await group.get(0).call('spv_getnewaddress')]) // Funds 1 BTC
+  })
+
+  afterAll(async () => {
+    await group.stop()
+  })
+
+  it('should not get HTLC secret from another node', async () => {
+    const pubKeyA = await group.get(0).call('spv_getaddresspubkey', [await group.get(0).call('spv_getnewaddress')])
+    const pubKeyB = await group.get(0).call('spv_getaddresspubkey', [await group.get(0).call('spv_getnewaddress')])
```
canonbrother, line 67:
```suggestion
    const pubKeyB = await group.get(1).call('spv_getaddresspubkey', [await group.get(0).call('spv_getnewaddress')])
```

canonbrother, line 67:
```suggestion
    const pubKeyB = await group.get(1).call('spv_getaddresspubkey', [await group.get(1).call('spv_getnewaddress')])
```

packages/jellyfish-api-core/__tests__/category/spv/getHtlcSeed.test.ts
```diff
@@ -0,0 +1,79 @@
...
+    const htlc = await group.get(0).call('spv_createhtlc', [pubKeyA, pubKeyB, '10'])
+
+    await group.get(0).call('spv_sendtoaddress', [htlc.address, 0.1]) // Funds HTLC address
+    await group.get(0).call('spv_claimhtlc', [htlc.address, await group.get(0).call('spv_getnewaddress'), htlc.seed]) // claim HTLC
```
canonbrother, line 71:
```suggestion
    await group.get(0).call('spv_sendtoaddress', [htlc.address, 0.1]) // Funds HTLC address
    await group.waitForSync()
    await group.get(0).call('spv_claimhtlc', [htlc.address, await group.get(0).call('spv_getnewaddress'), htlc.seed]) // claim HTLC
```

fuxingloh:
@surangap need your review on this too

packages/jellyfish-api-core/__tests__/category/spv/getHtlcSeed.test.ts
```diff
@@ -0,0 +1,78 @@
...
+  it('should not get HTLC secret from another node', async () => {
+    const pubKeyA = await group.get(0).call('spv_getaddresspubkey', [await group.get(0).call('spv_getnewaddress')])
+    const pubKeyB = await group.get(0).call('spv_getaddresspubkey', [await group.get(0).call('spv_getnewaddress')])
+    const htlc = await group.get(0).call('spv_createhtlc', [pubKeyA, pubKeyB, '10'])
+
+    await group.get(0).call('spv_sendtoaddress', [htlc.address, 0.1]) // Funds HTLC address
+    await group.waitForSync()
+    await group.get(0).call('spv_claimhtlc', [htlc.address, await group.get(0).call('spv_getnewaddress'), htlc.seed]) // claim HTLC
+
+    const emptySecret = await client1.spv.getHtlcSeed(htlc.address)
```
surangap, line 72: Hmm. This is incorrect as per my knowledge. once the secret is revealed, it is public. anyone can retrieve it.
consider the same scenario described here. other party, client1 should be able to retrieve the secret.

But i doubt that this `await group.waitForSync()` won't be enough for this. in real world, clients has to be synced in BTC network. means 10 mins for one block. Here in `regtest`, we don't even connect to BTC network. we just fake it. so a solution would be to do the faking in all nodes inside `group.waitForSync()`. But not sure it's an overkill. 
But the idea is anyone can retrive the secret once it's revealed.
@fuxingloh

Jouzo: Ok, that's not what I observed while testing. As you pointed out, it must be because it is not synced to BTC network.
How could I test real multi node situation instead ?

surangap: the answer is in my last comment. there are some fake/test spv RPCs that can be used to implement something like that. then it would be like we are faking and testing the same.

at this point, it does not make sense to use a multinode setup to test spv. even C++ side regtest tests are doing single node.

Jouzo: Should we put on hold multi node testing on jellyfish side until C++ regtests do it ?

fuxingloh: Yeah, let's put multi-node testing on hold (`it.skip`). Since there isn't a proper way to peer connection in the BTC network.

Jouzo: it.skip or just remove it ? As the logic is wrong anyway

surangap (approved):
looks good

fuxingloh (approved):

PR #529, opened 2021-07-30
Added SPV refundHtlc RPC

Jouzo:

/kind feature

Adds SPV refundHtlc RPC

Fixes part of #202

### DIFF
--- a/packages/jellyfish-api-core/__tests__/category/spv/refundHtlc.test.ts
+++ b/packages/jellyfish-api-core/__tests__/category/spv/refundHtlc.test.ts
@@ -25,7 +25,7 @@ describe('Spv', () => {
     await container.call('spv_sendtoaddress', [htlc.address, 0.1]) // Funds HTLC address
 
     const destinationAddress = await container.call('spv_getnewaddress')
-    const result = await client.spv.refundHtlc(htlc.address, destinationAddress)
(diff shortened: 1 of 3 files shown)


packages/jellyfish-api-core/__tests__/category/spv/refundHtlc.test.ts
```diff
@@ -0,0 +1,104 @@
...
+
+  it('should refundHtlc', async () => {
+    const pubKeyA = await container.call('spv_getaddresspubkey', [await container.call('spv_getnewaddress')])
+    const pubKeyB = await container.call('spv_getaddresspubkey', [await container.call('spv_getnewaddress')])
+    const htlc = await container.call('spv_createhtlc', [pubKeyA, pubKeyB, '10'])
+
+    await container.call('spv_sendtoaddress', [htlc.address, 0.1]) // Funds HTLC address
+
+    const destinationAddress = await container.call('spv_getnewaddress')
+    const result = await client.spv.refundHtlc(htlc.address, destinationAddress)
```
surangap, line 28: This seems incorrect with the real-life scenario. The Buyer has to wait until a timeout to call a  refund.
refer https://en.bitcoin.it/wiki/BIP_0199
But we experience this behavior because in `regtest` we do not connect to the Bitcoin network, rather fake it. either these time constraints should be added to C++ side spv fake wrapper. But for now please add a comment, otherwise a reader could get the wrong idea.

surangap (changes requested):
Requesting some comments/explanatory notes update

surangap (approved):
seems ok.

fuxingloh (approved):

PR #534, opened 2021-08-01
`txn-builder`: masternode-create and masternode-resign

### LINKED ISSUE #504: `jellyfish-txn-builder` - createMasternode / resignMasternode
builder logic for creating
- `CTransactionSegwit<CreateMasterNode>`
- `CTransactionSegwit<ResignMasterNode>`

#### Why is this needed:
Allow light wallet to create/resign masternode

canonbrother:

/kind feature
/kind fix

Add masternode create and resign to txn builder
Update rpc createmn test
Some fixes on createmn transaction

Fixes #504


packages/jellyfish-transaction/src/script/dftx/dftx_masternode.ts
```diff
@@ -7,9 +7,8 @@ import { BufferComposer, ComposableBuffer } from '../../buffer/buffer_composer'
  * CreateMasterNode DeFi Transaction
  */
 export interface CreateMasterNode {
-  type: number // ----------------------------------| 1 byte, if 0x01 p2pkh, otherwise = p2wpkh
-  collateralPubKeyHash: string // ------------------| VarUInt{20 bytes}
-  operatorPubKeyHash?: string // -------------------| VarUInt{20 bytes}, p2pkh only, optional, use collateral pkh as default
```
canonbrother, line 12: https://github.com/DeFiCh/ain/blob/85360ad432ae8c5ecbfbfc7d63dd5bc6fe41e875/src/masternodes/mn_checks.h#L157-L159
only one address for operating

packages/jellyfish-transaction/src/script/dftx/dftx_masternode.ts
```diff
@@ -7,9 +9,9 @@ import { BufferComposer, ComposableBuffer } from '../../buffer/buffer_composer'
...
+  operatorType: number // --------------------------| 1 byte, 0x01 = p2pkh, 0x04 = p2wpkh
+  operatorAuthAddress: string // -------------------| VarUInt{20 bytes}
+  timeLock?: number // -----------------------------| 4 bytes, only available after EunosPaya
```
surangap, line 14: from c++ side
uint16_t timelock{0};
```suggestion
  timeLock?: number // -----------------------------| 2 bytes, only available after EunosPaya
```

packages/jellyfish-transaction-builder/__tests__/txn/txn_builder_masternode_create.test.ts
```diff
@@ -0,0 +1,230 @@
...
+    }
+  })
+
+  const promise = sendTransaction(container, txn)
+  await expect(promise).rejects.toThrow(DeFiDRpcError)
+  await expect(promise).rejects.toThrow('CreateMasternodeTx: bad owner and|or operator address (should be P2PKH or P2WPKH only) or node with those addresses exists')
+})
+
+// ISSUE(canonbrother): createmasternode via sendrawtransaction will bypass balance checker??
+it.skip('should be failed as min 20k DFI is needed', async () => {
```
canonbrother, line 184: @fuxingloh  @monstrobishi @surangap 
i realized i missed the balance checker test so i added it then i found this issue
i ady checked the ain, seems it doesn't have balance checker while createmasternode via sendrawtransaction
but can you guys help confirm again? :pray:

packages/jellyfish-transaction/src/tx.ts
```diff
@@ -56,6 +56,11 @@ export interface Vout {
   value: BigNumber // ------------------| 8 bytes
   script: Script // --------------------| n = VarUInt{1-9 bytes}, + n bytes
   tokenId: number // -------------------| 1 byte (Although it is VarUInt but disabled hence always 0x00)
+
+  // ISSUE(canonbrother): nValue same as value, nTokenId same as tokenId, its inconsistent vout struct issue
+  // https://github.com/DeFiCh/ain/blob/c812f0283a52840996659121a755a9f723be2392/src/masternodes/mn_checks.cpp#L441-L442
+  nValue?: BigNumber // ----------------| 8 bytes
```
surangap, line 62: duplication here. c++ side only have `nValue, nTokenId` -> https://github.com/DeFiCh/ain/blob/948100ccc6d849c1e3d81e21d539ad9bb9d6f231/src/primitives/transaction.h#L140
we can either use `value,tokenId` or `nValue, nTokenId` since the variable name does not matter for serialized data. But latter would add more refactoring to jellyfish side existing code. suggesting to stick with `value,tokenId`

canonbrother: hmm.. if stick back with `value, tokenId` it will never pass because this checker
https://github.com/DeFiCh/ain/blob/c812f0283a52840996659121a755a9f723be2392/src/masternodes/mn_checks.cpp#L441-L442

surangap: should not be the case right? i mean in c++ side it constructs vout structure from serialized data. so it can use any name. should be ok as long as c++ side sticks to the same variables. correct me if i'm missing anything.

packages/jellyfish-transaction-builder/__tests__/txn/txn_builder_masternode_create.test.ts
surangap: test description says it all? `regtest` does not have `20K` limit

packages/jellyfish-transaction/src/script/dftx/dftx_masternode.ts
```diff
@@ -7,9 +9,9 @@ import { BufferComposer, ComposableBuffer } from '../../buffer/buffer_composer'
  * CreateMasterNode DeFi Transaction
  */
 export interface CreateMasterNode {
-  type: number // ----------------------------------| 1 byte, if 0x01 p2pkh, otherwise = p2wpkh
-  collateralPubKeyHash: string // ------------------| VarUInt{20 bytes}
-  operatorPubKeyHash?: string // -------------------| VarUInt{20 bytes}, p2pkh only, optional, use collateral pkh as default
+  operatorType: number // --------------------------| 1 byte, 0x01 = p2pkh, 0x04 = p2wpkh
+  operatorAuthAddress: string // -------------------| VarUInt{20 bytes}
+  timeLock?: number // -----------------------------| 2 bytes, only available after EunosPaya
```
surangap, line 14: add dftx test to check this one also?

surangap (changes requested):
requesting some refactoring and additional tests

packages/jellyfish-transaction/src/tx.ts
canonbrother: it just a value, should not affected by diff property

packages/jellyfish-transaction-builder/__tests__/txn/txn_builder_masternode_create.test.ts
canonbrother: ohh.. checked.. just need 2 for collateral

packages/jellyfish-transaction-builder/src/txn/txn_builder_masternode.ts
```diff
@@ -0,0 +1,20 @@
+import { BigNumber } from '@defichain/jellyfish-json'
+import { OP_CODES, Script, TransactionSegWit, CreateMasterNode, ResignMasterNode } from '@defichain/jellyfish-transaction'
+import { P2WPKHTxnBuilder } from './txn_builder'
+
+export class TxnBuilderMasternode extends P2WPKHTxnBuilder {
+  async create (createMasternode: CreateMasterNode, changeScript: Script): Promise<TransactionSegWit> {
```
surangap, line 6: can add some js docs

surangap (changes requested):
one docs update. then good.

surangap (dismissed):
Looks good to me

packages/jellyfish-transaction-builder/__tests__/txn/txn_builder_masternode_create.test.ts
```diff
@@ -0,0 +1,186 @@
...
+  const script = await providers.elliptic.script()
+
+  const txn: TransactionSegWit = await builder.masternode.create(createMasternode, script)
+
+  try {
+    await sendTransaction(container, txn)
+  } catch (err) {
+    expect(err.message).toStrictEqual('DeFiDRpcError: \'CreateMasternodeTx: malformed tx vouts (wrong creation fee or collateral amount) (code 16)\', code: -26')
+  }
+})
```
fuxingloh, line 186: need to add timelock tests

packages/jellyfish-transaction-builder/__tests__/txn/txn_builder_masternode_resign.test.ts
```diff
@@ -0,0 +1,105 @@
...
+    nodeId: txid
+  }
+
+  const script = await providers.elliptic.script()
+  const txn = await builder.masternode.resign(resignMasterNode, script)
+
+  const promise = sendTransaction(container, txn)
+  await expect(promise).rejects.toThrow(DeFiDRpcError)
+  await expect(promise).rejects.toThrow('ResignMasternodeTx: tx must have at least one input from the owner')
+})
```
fuxingloh, line 105: need to add timelock failed test

packages/jellyfish-transaction-builder/src/txn/txn_builder_masternode.ts
```diff
@@ -0,0 +1,34 @@
...
+   *
+   * @param {CreateMasterNode} createMasternode transaction to create
+   * @param {Script} changeScript to send unspent to after deducting the (converted + fees)
+   * @return {Promise<TransactionSegWit>}
+   */
+  async create (createMasternode: CreateMasterNode, changeScript: Script): Promise<TransactionSegWit> {
+    return await this.createDeFiTx(
+      OP_CODES.OP_DEFI_TX_CREATE_MASTER_NODE(createMasternode),
+      changeScript,
+      new BigNumber('1')
```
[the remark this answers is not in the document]
fuxingloh: how do collaterals get locked?

which address lock the collaterals?

fuxingloh (changes requested):

packages/jellyfish-transaction-builder/__tests__/txn/txn_builder_masternode_create.test.ts
canonbrother: currently build doesn't support `timelock`

https://github.com/DeFiCh/ain/blob/e6caba0c0b220182b69e6043502a51c3f567c599/src/masternodes/mn_checks.h#L167-L170

for this i might need build another docker image with eunospaya release, right??

packages/jellyfish-transaction-builder/src/txn/txn_builder_masternode.ts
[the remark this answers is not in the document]
canonbrother: yes the `1` is for creation fee, actually the `note` was added
https://github.com/DeFiCh/jellyfish/blob/bb6ce39d68eafe2ff628599eb6bd5b95279a6f56/packages/jellyfish-transaction-builder/__tests__/txn/txn_builder_masternode_create.test.ts#L36-L38

sure add a simple desc here

packages/jellyfish-transaction-builder/__tests__/txn/txn_builder_masternode_create.test.ts
canonbrother: rebase the `main` branch, it supports mn.timeLock

packages/jellyfish-transaction/__tests__/script/dftx/dftx_masternode/CreateMasterNode.test.ts
```diff
@@ -20,85 +22,35 @@ it('should bi-directional buffer-object-buffer', () => {
...
-
-    it('should compose from composable to buffer', () => {
-      const composable = new CCreateMasterNode(createMasterNode)
-      const buffer = new SmartBuffer()
-      composable.toBuffer(buffer)
+const header = '6a1a4466547843' // OP_RETURN(0x1a) (length 26 = 0x1a) CDfTx.SIGNATURE(0x44665478) CCreateMasterNode.OP_CODE(0x43)
+const data = '01742b337e0f40d5f229a89d3a26d53ae1093b6cff' // CreateMasterNode.operatorType(0x01) CreateMasterNode.operatorAuthAddress(0x742b337e0f40d5f229a89d3a26d53ae1093b6cff)
+const createMasterNode: CreateMasterNode = {
+  operatorType: 0x01,
+  operatorAuthAddress: '742b337e0f40d5f229a89d3a26d53ae1093b6cff'
```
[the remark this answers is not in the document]
canonbrother: hmm... for txn test, instead of adding another describe block..
its sufficient to extend the hex on test as serialization doesn't have vary business logics (testing the same condition only)

https://github.com/DeFiCh/jellyfish/blob/703d059a2ab08f3b52fc7dc29996661f717dd243/packages/jellyfish-transaction/__tests__/script/dftx/dftx_masternode/CreateMasterNode.test.ts#L12-L13

https://github.com/DeFiCh/jellyfish/blob/703d059a2ab08f3b52fc7dc29996661f717dd243/packages/jellyfish-transaction/__tests__/script/dftx/dftx_masternode/CreateMasterNode.test.ts#L26-L32

packages/jellyfish-transaction/__tests__/script/dftx/dftx_masternode/CreateMasternode.test.ts
```diff
@@ -20,12 +23,12 @@ it('should bi-directional buffer-object-buffer', () => {
...
 })
 
-describe('without `operatorPubKeyHash`', () => {
-  const header = '6a1a4466547843' // OP_RETURN (length 26 = 0x1a), PUSH_DATA(44665478, 43)
-  const data = '01742b337e0f40d5f229a89d3a26d53ae1093b6cff'
-  const createMasterNode: CreateMasterNode = {
-    type: 0x01,
-    collateralPubKeyHash: '742b337e0f40d5f229a89d3a26d53ae1093b6cff'
+describe('CreateMasternode', () => {
+  const header = '6a1a4466547843' // OP_RETURN(0x1a) (length 28 = 0x1c) CDfTx.SIGNATURE(0x44665478) CCreateMasternode.OP_CODE(0x43)
```
surangap, line 27:
```suggestion
  const header = '6a1a4466547843' // OP_RETURN(0x6a) (length 28 = 0x1c) CDfTx.SIGNATURE(0x44665478) CCreateMasternode.OP_CODE(0x43)
```

packages/jellyfish-transaction/src/script/dftx/dftx_masternode.ts
```diff
@@ -1,30 +1,42 @@
...
+      ComposableBuffer.uInt8(() => cmn.operatorType, v => cmn.operatorType = v),
+      ComposableBuffer.hex(20, () => cmn.operatorAuthAddress, v => cmn.operatorAuthAddress = v),
+      {
+        fromBuffer: (buffer: SmartBuffer): void => {
+          if (buffer.remaining() > 0) {
+            cmn.timelock = buffer.readUInt16LE()
+          }
+        },
+        toBuffer: (buffer: SmartBuffer): void => {
+          if (cmn.timelock !== undefined) {
```
surangap, line 35: After eunosPaya, timelock value(0x0000) will be added when serializing if not specified. this will change the total length to 28(0x1c). nodes might still parse it to default value 0 even if we don't send it. But better to implement accordingly with defid. 
so have a default value 0 for timelock and write that when serializing.
This is what is get for scriptPubkey without timelock and with it.
```
6a 1c 44665478 43 04 c93c4da5a164f96c18f5abc1c7534bd2d7036199 0000
6a 1c 44665478 43 04 dd8537bb5d1f069f8d4c66770a9fc705373ba98c 0401
```

surangap (changes requested):
Requesting some changes after eunosPaya, 
also refactor the test fixtures, if you need to test the behaviour before eunosPaya, you might need to use another container without eunosPaya flag set.

canonbrother:
> Requesting some changes after eunosPaya,
> also refactor the test fixtures, if you need to test the behaviour before eunosPaya, you might need to use another container without eunosPaya flag set.

@surangap 
I couldn't reason this as personally i think this is not that necessary unless that has critical i need to fix
first, the `txn` fixture only be involved in buffer composer and serialization logic, it doesn't really care whether the fixture is correct, eg: putting invalid in the fixture, it will still pass, but for better document wise, we will ensure the fixture is correct, some we pick those from mainnet, if have. I added the default timelock 0000 fixture with desc, i think that is clear enough.

second, the before/after eunospaya logic can still work fine, so if create another old fork container to test the old fork logic, seems like really not necessary, unless this
eg: the before/after params order different, then need another build to test
before { ownerAddr: string, opeAddr: string }
after { ownerAddr: string, timelock?: string, opeAddr?: string }

feel free to advice if i'm miss something

surangap:
> > Requesting some changes after eunosPaya,
> > also refactor the test fixtures, if you need to test the behaviour before eunosPaya, you might need to use another container without eunosPaya flag set.
> 
> @surangap
> I couldn't reason this as personally i think this is not that necessary unless that has critical i need to fix
> first, the `txn` fixture only be involved in buffer composer and serialization logic, it doesn't really care whether the fixture is correct, eg: putting invalid in the fixture, it will still pass, but for better document wise, we will ensure the fixture is correct, some we pick those from mainnet, if have. I added the default timelock 0000 fixture with desc, i think that is clear enough.
> 
> second, the before/after eunospaya logic can still work fine, so if create another old fork container to test the old fork logic, seems like really not necessary, unless this
> eg: the before/after params order different, then need another build to test
> before { ownerAddr: string, opeAddr: string }
> after { ownerAddr: string, timelock?: string, opeAddr?: string }
> 
> feel free to advice if i'm miss something

I think you missed the main point from my comments. https://github.com/DeFiCh/jellyfish/pull/534/files#r687649278
For the fixtures, yes you can use anything as long as it accords with the serialization logic. eunosPaya flag is additional information I mentioned, which might not need for this test. ignore that.

let me explain more about the main point, https://github.com/DeFiCh/jellyfish/pull/534/files#r687649278
let's say with the current jellyfish implementation, after eunosPaya, if someone craft a `createMasternode` tx without a timelock, the serialized scriptPubkey will be `6a 1a 44665478 43 01 b684e2095bcdfb63f1265f35775da23ca8e8aaff`. But if you create the the same tx using `createmasternode` defid RPC, it will give you `6a 1c 44665478 43 01 b684e2095bcdfb63f1265f35775da23ca8e8aaff 0000`
notice the difference in length. also refer https://github.com/DeFiCh/ain/blob/d04337ff3e884b4fb240ca65218c2c13d923f2ad/src/masternodes/rpc_masternodes.cpp#L167 for more info.

As i mentioned in my previous comment, `6a 1a 44665478 43 01 b684e2095bcdfb63f1265f35775da23ca8e8aaff` might still be parsed by the nodes and populate the timelock as `0000` since it is the default value in c++ side. refer https://github.com/DeFiCh/ain/blob/d04337ff3e884b4fb240ca65218c2c13d923f2ad/src/masternodes/mn_checks.h#L159
But I would prefer to have the same implementation as the c++ side.

canonbrother:
> > > Requesting some changes after eunosPaya,
> > > also refactor the test fixtures, if you need to test the behaviour before eunosPaya, you might need to use another container without eunosPaya flag set.
> > 
> > 
> > @surangap
> > I couldn't reason this as personally i think this is not that necessary unless that has critical i need to fix
> > first, the `txn` fixture only be involved in buffer composer and serialization logic, it doesn't really care whether the fixture is correct, eg: putting invalid in the fixture, it will still pass, but for better document wise, we will ensure the fixture is correct, some we pick those from mainnet, if have. I added the default timelock 0000 fixture with desc, i think that is clear enough.
> > second, the before/after eunospaya logic can still work fine, so if create another old fork container to test the old fork logic, seems like really not necessary, unless this
> > eg: the before/after params order different, then need another build to test
> > before { ownerAddr: string, opeAddr: string }
> > after { ownerAddr: string, timelock?: string, opeAddr?: string }
> > feel free to advice if i'm miss something
> 
> I think you missed the main point from my comments. [#534 (comment)](https://github.com/DeFiCh/jellyfish/pull/534#discussion_r687649278)
> For the fixtures, yes you can use anything as long as it accords with the serialization logic. eunosPaya flag is additional information I mentioned, which might not need for this test. ignore that.
> 
> let me explain more about the main point, [#534 (comment)](https://github.com/DeFiCh/jellyfish/pull/534#discussion_r687649278)
> let's say with the current jellyfish implementation, after eunosPaya, if someone craft a `createMasternode` tx without a timelock, the serialized scriptPubkey will be `6a 1a 44665478 43 01 b684e2095bcdfb63f1265f35775da23ca8e8aaff`. But if you create the the same tx using `createmasternode` defid RPC, it will give you `6a 1c 44665478 43 01 b684e2095bcdfb63f1265f35775da23ca8e8aaff 0000`
> notice the difference in length. also refer https://github.com/DeFiCh/ain/blob/d04337ff3e884b4fb240ca65218c2c13d923f2ad/src/masternodes/rpc_masternodes.cpp#L167 for more info.
> 
> As i mentioned in my previous comment, `6a 1a 44665478 43 01 b684e2095bcdfb63f1265f35775da23ca8e8aaff` might still be parsed by the nodes and populate the timelock as `0000` since it is the default value in c++ side. refer https://github.com/DeFiCh/ain/blob/d04337ff3e884b4fb240ca65218c2c13d923f2ad/src/masternodes/mn_checks.h#L159
> But I would prefer to have the same implementation as the c++ side.

yes i understand the changes before/after the eunospaya....
but i don't quite get the request for the test...
hmm.. can you pseudo how does the changes should be?
it's great if you can straight away contribute the changes

fuxingloh:
I think by that effect, making `timelock` a non-optional value where it always defaults to`0x0000`. 
   
```diff
- timelock?: number // -----------------------------| 2 bytes, 0x0401 = 5 years, 0x0802 = 10 years
+ timelock: number // -----------------------------| 2 bytes, 0x0401 = 5 years, 0x0802 = 10 years
```

fuxingloh:
> I think by that effect, making `timelock` a non-optional value where it always defaults to`0x0000`.
> 
> ```diff
> - timelock?: number // -----------------------------| 2 bytes, 0x0401 = 5 years, 0x0802 = 10 years
> + timelock: number // -----------------------------| 2 bytes, 0x0401 = 5 years, 0x0802 = 10 years
> ```

Quite annoying we can't get backward compatibility..., how do we know which format to use without knowing block height?

canonbrother:
> I think by that effect, making `timelock` a non-optional value where it always defaults to`0x0000`.
> 
> ```diff
> - timelock?: number // -----------------------------| 2 bytes, 0x0401 = 5 years, 0x0802 = 10 years
> + timelock: number // -----------------------------| 2 bytes, 0x0401 = 5 years, 0x0802 = 10 years
> ```

i see.. is this what @surangap you mean??

monstrobishi:
@fuxingloh if we make it default to 0000, won't it produce a different OP_RETURN? then the serialization/deserialization won't line up

fuxingloh:
> @fuxingloh if we make it default to 0000, won't it produce a different OP_RETURN? then the serialization/deserialization won't line up

Damn it...

monstrobishi:
When ain deserializes, it checks for end of stream and omits timelock if it's not there

When it serializes, it checks block height

monstrobishi:
Perhaps if timelock was optional, then when we deserialize we know that this was a pre Eunos tx, and when we serialize again we omit timelock.

fuxingloh:
> Perhaps if timelock was optional, then when we deserialize we know that this was a pre Eunos Paya tx, and when we serialize again we omit timelock.

I really hope we can omit to carrying the `height` flag around when parsing `DfTx`. More and more reasons where that's not the case...

Another e.g. where there just lack of consistency: https://github.com/DeFiCh/jellyfish/pull/556

ivan-zynesis:
> @fuxingloh if we make it default to 0000, won't it produce a different OP_RETURN? then the serialization/deserialization won't line up

Exactly, thus we may endedup with `PreEunosCreateMasterNode` and `PostEunosCreateMasterNode`, upto user (eg whale indexer) to decide what to do.

surangap:
> When ain deserializes, it checks for end of stream and omits timelock if it's not there
> 
> When it serializes, it checks block height

This is not exactly correct I think. 

- when create masternode via RPC it checks for height and write timelock or not when writing tx metadata
- But when deserialize, if EOS, then it will be the default value 0 for timelock. even inside defid, it expects pre EunosPaya to have timelock value of 0. refer https://github.com/DeFiCh/ain/blob/d04337ff3e884b4fb240ca65218c2c13d923f2ad/src/masternodes/mn_checks.cpp#L699
- need to do another check/test from c++ side for serialized output for pre eunosPaya without timelock. `getrawtransaction` might be returning from the mempool. need to check otherwise.

canonbrother:
understand, even on preEunosPaya, timelock must equal to 0

packages/jellyfish-transaction/__tests__/script/dftx/dftx_masternode/CreateMasternode.test.ts
```diff
@@ -0,0 +1,59 @@
...
+
+    it('should compose from composable to buffer', () => {
+      const composable = new CCreateMasternode(createMasterNode)
+      const buffer = new SmartBuffer()
+      composable.toBuffer(buffer)
+
+      expect(buffer.toBuffer().toString('hex')).toStrictEqual(data)
+    })
+  })
+})
```
monstrobishi, line 59: Need tests here for sending create masternode tx, also test with timelock, no timelock etc

monstrobishi (changes requested):

packages/jellyfish-transaction/__tests__/script/dftx/dftx_masternode/CreateMasternode.test.ts
canonbrother: huh?? timelock is mandatory now in rawtx section 
sending createmn tx test is here
https://github.com/DeFiCh/jellyfish/blob/e224ab4626b7e4ee186f6e5188789c6f3888bd7e/packages/jellyfish-transaction-builder/__tests__/txn/txn_builder_masternode_create.test.ts

but timelock in rpc call is optional

monstrobishi: thanks I see that file now! I looked for ages and couldn't find it at first, I was looking at the composer and the txn-builder

for no timelock, I request a test to decompose pre Eunos Paya txns

canonbrother: hmm.. sounds like my previous commit should be correct.. the preEunosPaya doesn't have timelock impl.. the timelock should not be mandatory in rawtx section...

surangap:
> > When ain deserializes, it checks for end of stream and omits timelock if it's not there
> > When it serializes, it checks block height
> 
> This is not exactly correct I think.
> 
> * when create masternode via RPC it checks for height and write timelock or not when writing tx metadata
> * But when deserialize, if EOS, then it will be the default value 0 for timelock. even inside defid, it expects pre EunosPaya to have timelock value of 0. refer https://github.com/DeFiCh/ain/blob/d04337ff3e884b4fb240ca65218c2c13d923f2ad/src/masternodes/mn_checks.cpp#L699
> * need to do another check/test from c++ side for serialized output for pre eunosPaya without timelock. `getrawtransaction` might be returning from the mempool. need to check otherwise.

pre eunosPaya timelock 0 at decerialization inside `defid`, can be seen with this PR https://github.com/DeFiCh/ain/pull/679

canonbrother:
> pre eunosPaya timelock 0 at decerialization inside `defid`, can be seen with this PR [DeFiCh/ain#679](https://github.com/DeFiCh/ain/pull/679)

hmm.. which docker image i should refer for this update huh??

packages/jellyfish-transaction/src/script/dftx/dftx_masternode.ts
```diff
@@ -1,30 +1,42 @@
...
+            buffer.writeUInt16LE(cmn.timelock)
+          }
+        }
```
surangap, line 38: Suggesting this here. not sure if i'm being short sighted here. if then please mention.
```suggestion
        fromBuffer: (buffer: SmartBuffer): void => {
          if (buffer.remaining() > 0) {
            cmn.timelock = buffer.readUInt16LE()
          }
        },
        toBuffer: (buffer: SmartBuffer): void => {
          if (cmn.timelock == undefined) {
              cmn.timelock = 0x0000
           }
           buffer.writeUInt16LE(cmn.timelock)
        }
```
Few  points.

- for pre or post eunosPaya `defid` can work with either `0x1c` or `0x1a` lengths.
- so on the jellyfish side, we can always serialize with `0x1c`, i.e with timelock.
- When deserializing at jellyfish side, we can support both `0x1c` or `0x1a` lengths with the above implementation. 
- But note that if we deserialize a pre eunosPaya tx and then serialize the same, we will get `0x1c` version with timelock `0x0000`

canonbrother: i think only the last point will cause issue but not critical

> But note that if we deserialize a pre eunosPaya tx and then serialize the same, we will get 0x1c version with timelock 0x0000

so jellyfish user might need to take note while they're composing preEnuosPaya mn txid
i mean they need to manually fix the txid to get the data, seems not so good :thinking: 
```ts
const hex = '6a1a4466547843047b9275d2116bd2cd96a5f5b598a1470b99fc5c35' // <-- wrong
const hex = '6a1c4466547843047b9275d2116bd2cd96a5f5b598a1470b99fc5c350000' // <-- correct
const stack = toOPCodes(
      SmartBuffer.fromBuffer(Buffer.from(hex, 'hex'))
)
const buffer = toBuffer(stack)
```

surangap: > i think only the last point will cause issue but not critical
> 
> > But note that if we deserialize a pre eunosPaya tx and then serialize the same, we will get 0x1c version with timelock 0x0000
> 
> so jellyfish user might need to take note while they're composing preEnuosPaya mn txid
> i mean they need to manually fix the txid to get the data, seems not so user-friendly 🤔
> 
> ```ts
> const hex = '6a1a4466547843047b9275d2116bd2cd96a5f5b598a1470b99fc5c35' // <-- wrong
> const hex = '6a1c4466547843047b9275d2116bd2cd96a5f5b598a1470b99fc5c350000' // <-- correct
> const stack = toOPCodes(
>       SmartBuffer.fromBuffer(Buffer.from(hex, 'hex'))
> )
> const buffer = toBuffer(stack)
> ```

do you mean the `hex` not `txid`? `hex` means hex string of the tx. in that case the above would be fine for both pre/post eunosPaya deserialization.
since you pointed out, the `txid` will definitely be changed if we deserialize a pre eunosPaya tx and serialize the same again in jellyfish side. if we have such usage, it might be a problem 🤔

canonbrother: yea.. example use case in whale indexing.. user is required to handle the preEunosPaya mn hex... see whether we want opt to this approach

surangap: hmm. then the ugly solution would be to incorporate `height` in dftx crafting layer or something similar to https://github.com/DeFiCh/jellyfish/pull/534#issuecomment-898264787 @fuxingloh

packages/jellyfish-transaction/src/script/dftx/dftx_masternode.ts
```diff
@@ -1,30 +1,42 @@
...
- * CreateMasterNode DeFi Transaction
+ * CreateMasternode DeFi Transaction
  */
-export interface CreateMasterNode {
-  type: number // ----------------------------------| 1 byte, if 0x01 p2pkh, otherwise = p2wpkh
-  collateralPubKeyHash: string // ------------------| VarUInt{20 bytes}
-  operatorPubKeyHash?: string // -------------------| VarUInt{20 bytes}, p2pkh only, optional, use collateral pkh as default
+export interface CreateMasternode {
+  operatorType: number // --------------------------| 1 byte, 0x01 = p2pkh, 0x04 = p2wpkh
+  operatorAuthAddress: string // -------------------| VarUInt{20 bytes}
```
monstrobishi, line 12: Maybe we should keep the name `operatorPubKeyHash` for this one since it's technically the pubkey not the address (I just confirmed)

packages/jellyfish-transaction-builder/src/txn/txn_builder_masternode.ts
```diff
@@ -0,0 +1,34 @@
...
+   *
+   * @param {CreateMasternode} createMasternode transaction to create
+   * @param {Script} changeScript to send unspent to after deducting the (converted + fees)
+   * @return {Promise<TransactionSegWit>}
+   */
+  async create (createMasternode: CreateMasternode, changeScript: Script): Promise<TransactionSegWit> {
+    return await this.createDeFiTx(
+      OP_CODES.OP_DEFI_TX_CREATE_MASTER_NODE(createMasternode),
+      changeScript,
+      new BigNumber('1') // for creation fee (regtest - 1, other than regtest - 10)
```
ivan-zynesis, line 17: Missing env handling? In test/prod, it might be insufficient balance in collect prevout and fail right?

packages/jellyfish-transaction/src/script/dftx/dftx_masternode.ts
ivan-zynesis: Situation: post eunos (which will be all the way in the future), timelock undefined
still mandatory to write 0x0000 right?

canonbrother: @surangap  @ivan-zynesis 
this is the issue
https://github.com/DeFiCh/whale/blob/8a050fc06bfc6cded44761952ab99b15f58d724e/src/module.indexer/model/dftx.indexer.ts#L60

ivan-zynesis: If we do not include any height info in dftx crafting layer, there will be no solution for actually able to create masternode buffer correctly both pre and post Eunos for timelock undefined case. Practically, it no longer requires to do pre Eunos, but forcing default value with have issue with the "bi-directional" test.

packages/jellyfish-transaction-builder/src/txn/txn_builder_masternode.ts
canonbrother: yea you got the point.. hmm.. thinking about it

packages/jellyfish-transaction/src/script/dftx/dftx_masternode.ts
canonbrother: > Practically, it no longer requires to do pre Eunos, but forcing default value with have issue with the "bi-directional" test.

I don't think this is a good practice. The better approach should be able to handle pre an post cases.

packages/jellyfish-transaction-builder/src/txn/txn_builder_masternode.ts
canonbrother: the easiest approach for now
https://github.com/DeFiCh/jellyfish/blob/92c3ebc4ae14424ea69a85759af9ae54c8847fc5/packages/jellyfish-transaction-builder/src/txn/txn_builder_masternode.ts#L13-L20

packages/jellyfish-transaction/src/script/dftx/dftx_masternode.ts
canonbrother: it's actually not that critical, it's just a "ux", no point to having long discussion but dangling the PR also blocking the project
either `force default` or `optional`
i follow majority

fuxingloh: Let's go with optional now

Although it's inconsistent, it's something bearable on the jellyfish side. To properly look into this, we need to explore this in Q4 with an indexer rebuild.

It's not a huge issue for crafting transactions because `timelock` can be enforced on `jellyfish-transaction-builder` making it non optional.

packages/jellyfish-transaction-builder/src/txn/txn_builder_masternode.ts
```diff
@@ -0,0 +1,35 @@
...
+
+export class TxnBuilderMasternode extends P2WPKHTxnBuilder {
+  /**
+   * Build create masternode transaction
+   *
+   * @param {CreateMasternode} createMasternode transaction to create
+   * @param {Script} changeScript to send unspent to after deducting the (converted + fees)
+   * @return {Promise<TransactionSegWit>}
+   */
+  async create (createMasternode: CreateMasternode, changeScript: Script, network = 'regtest'): Promise<TransactionSegWit> {
```
surangap, line 13: I think with this https://github.com/DeFiCh/jellyfish/pull/534/files#r690875965,
we need to handle `force default` logic here either based on `height` or user input(creator of the `createMasternode` msg). and add some safety checks based on that.

canonbrother: i don't get it.. don't we handle it `optional`?? why do `force default` logic here??
can you submit a suggestion code here??

monstrobishi (approved):

packages/jellyfish-transaction-builder/src/txn/txn_builder_masternode.ts
monstrobishi: I think we can safely omit this, for me it makes sense to prioritise accurate deserialization of historic (pre Eunos Paya) tx, for indexing, but why add so much complexity to enforce sending of historic tx? the practical use case is non existant

surangap: This is actually to ensure the safety of the future, not the past. consider this scenario. someone tries to create a createMasternode tx after post eunosPaya without timelock. with the current dftx layer implementation, it would not write `0x0000` when serializing. two options

1. we let the creator of the `createMasternode` message take full responsibility, i.e. always mention timelock 0 when creating post eunosPaya. if he does not, too bad it would be the incorrect format. 
2. or else we can ask for another parameter from the user, `forceDefaultTimelock` so that he is informed, it's going to be post eunosPaya format.

I would prefer the second option.
@canonbrother if you still need code suggestions, I could provide one if we could agree that my point is valid and should be addressed.

monstrobishi: I see your point. I was referring to the extra logic like adding a height check. If we default to enforcing new behaviour as you suggested i.e. #2 then I'm ok

canonbrother: @surangap  hmm.. understand.. i asked the suggestion to allow me better understand the request
the suggestion:
```ts
createMasternode.timelock = createMasternode.timelock ?? 0x0000
```

monstrobishi: @canonbrother not quite, he's suggesting adding a default value `forceDefaultTimelock` (defaults to true) to enable this behaviour

canonbrother: @monstrobishi  @surangap I think this is good for me.. better ux at least... :sweat_smile: 
i will add this
```ts
createMasternode.timelock = createMasternode.timelock ?? 0x0000
```

surangap: ok, this addresses the problem I think. we can agree that in real life nobody is going to use this to create masternodes for pre eunosPaya unless the chain is rolled back. for regtest, it's fine IMO

surangap (approved):
looks ok now👏

packages/jellyfish-transaction-builder/src/txn/txn_builder_masternode.ts
canonbrother: out of topic.. never see a blockchain rollback its structure.. but switching network is yes (eg: eth2)

packages/jellyfish-transaction/src/script/dftx/dftx_masternode.ts
ivan-zynesis: My question was actually if post Eunos defid will reject timelock undefined (missing 0x0000 byte), as long as that do not happen, this looks good.

ivan-zynesis (approved):

monstrobishi (approved):

packages/jellyfish-transaction-builder/__tests__/txn/txn_builder_masternode_resign.test.ts
```diff
@@ -0,0 +1,155 @@
...
+
+it('should be failed as tx must have at least one input from owner', async () => {
+  const collateralAddress = await container.getNewAddress('', 'legacy')
+
+  const txid = await jsonRpc.masternode.createMasternode(collateralAddress)
+  await container.generate(1)
+
+  const masternodes = await jsonRpc.masternode.listMasternodes()
+  const createdMasternode = Object.values(masternodes).find(mn => mn.ownerAuthAddress === collateralAddress)
+  expect(createdMasternode).not.toBeUndefined()
```
fuxingloh, line 91:
```suggestion
  expect(createdMasternode).toBeDefined()
```

fuxingloh (approved):
Ignore the requested change, I need to think about something for the network flag.

Overall, this is a great PR. Very detailed and explicit.

Something that I am thinking about when we implement on Wallet Side.

Currently, we derive our wallet node as such for all wallet activity.
```
1129/{account}/0/0
```

Masternode locks UTXO making that address unusable, this requires us to extend and use other paths of the HD wallet for masternode creation.

```
1129/1/0/0 -> MN 0
1129/1/0/1 -> MN 1
1129/1/0/2 -> MN 1
```

I will make some changes in `jellyfish-wallet` to allow for this kind of configuration and reservation to ease DeFiCh/wallet integration.

PR #536, opened 2021-08-02
removed template interface in `jellyfish-api-core`

fuxingloh:

/kind fix

```diff
- type TokenRegexType = `${number}@${string}`
```

- This is a new feature that is added very recently, not all typescript projects support it. E.g. `DeFiCh/wallet`
- Doesn't work well with dynamic variable construction

### DIFF
--- a/packages/jellyfish-api-core/src/category/token.ts
+++ b/packages/jellyfish-api-core/src/category/token.ts
@@ -1,8 +1,6 @@
 import BigNumber from 'bignumber.js'
 import { ApiClient } from '../.'
 
-type TokenRegexType = `${number}@${string}`
-
 /**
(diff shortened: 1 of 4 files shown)


packages/jellyfish-api-core/src/category/token.ts
```diff
@@ -91,14 +89,14 @@ export class Token {
   /**
    * Creates a transaction to mint tokens.
    *
-   * @param {TokenRegexType} payload
+   * @param {`${number}@${string}`} amountToken
```
jingyi2811, line 92: Isn't that this should be `string`?

jingyi2811 (approved):

canonbrother (approved):

PR #547, opened 2021-08-03
Add listCollateralTokens rpc

jingyi2811:

/kind feature

Add listCollateralTokens rpc

Fixes part of #488


packages/jellyfish-api-core/__tests__/category/loan/listCollateralTokens.test.ts
```diff
@@ -0,0 +1,71 @@
...
+
+  afterAll(async () => {
+    await container.stop()
+  })
+
+  it('should listCollateralTokens with empty object if there is no collateral tokens available', async () => {
+    const data = await client.loan.listCollateralTokens()
+    expect(Object.keys(data).length).toStrictEqual(0)
+  })
+
```
siradji, line 40: I think this test should come after 'it should list token'. From the simplest to more complex tests.

siradji (changes requested):

packages/jellyfish-api-core/__tests__/category/loan/listCollateralTokens.test.ts
jingyi2811: The problem is so far there is no way to delete collateral token once it was created.

PR #551, opened 2021-08-03
Add listCollateralTokens rpc

jingyi2811:

/kind feature

Add listCollateralTokens rpc

Fixes part of #488


packages/jellyfish-api-core/__tests__/category/loan/listCollateralTokens.test.ts
```diff
@@ -0,0 +1,68 @@
...
+      token: 'TSLA',
+      factor: new BigNumber(1),
+      priceFeedId: priceFeedId2,
+      activateAfterBlock: 170
+    }])
+    await testing.generate(1)
+
+    const data = await testing.rpc.loan.listCollateralTokens()
+    expect(data).toStrictEqual({
+      [collateralTokenId1]: { activateAfterBlock: 160, factor: 0.5, priceFeedId: priceFeedId1, token: 'AAPL' },
```
Jouzo, line 58: `factor` should be BigNumber.
I don't know how this should be done as it is nested and not handled by `{ factor: 'bignumber' }`
cc @fuxingloh

fuxingloh: 2 Problems

1. `{ factor: 'bignumber' }` only works for list and object, not for a map with keys. Difficult to do a full mapping, we should just resort to mapping everything as BigNumber
2. `C++` shouldn't return a `Map` for a pagination/list query, counter-intuitive. It losses the natural ordering and is semantically not a sound way to deal with list/cursor operations. They should return an array.

packages/jellyfish-api-core/src/category/loan.ts
```diff
@@ -131,6 +140,17 @@ export interface SetCollateralToken {
...
 
+export interface CollateralTokensData {
+  [key: string]: CollateralTokenDetail
+}
+
+export interface CollateralTokenDetail {
+  token: string
+  factor: BigNumber
+  priceFeedId: string
+  activateAfterBlock: BigNumber
```
surangap, line 151: optional. https://github.com/DeFiCh/ain/blob/a43fda7d0842ff1c90ab39880df38077ba36bd91/src/masternodes/rpc_loan.cpp#L15

packages/jellyfish-api-core/__tests__/category/loan/listCollateralTokens.test.ts
```diff
@@ -0,0 +1,62 @@
...
+      token: 'TSLA',
+      factor: new BigNumber(1),
+      priceFeedId: priceFeedId2,
+      activateAfterBlock: 130
+    }])
+    await testing.generate(1)
+
+    const data = await testing.rpc.loan.listCollateralTokens()
+    expect(data).toStrictEqual({
+      [collateralTokenId1]: { activateAfterBlock: new BigNumber(120), factor: new BigNumber(0.5), priceFeedId: priceFeedId1, token: 'AAPL' },
```
surangap, line 58: Also, add a collateral token without `activateAfterBlock` to the test.

packages/jellyfish-api-core/src/category/loan.ts
jingyi2811: For setCollateralToken, yes, activateAfterBlock is optional.  However, listCollateralTokens always return activateAfterBlock value. If you don't pass in activateAfterBlock in the setCollateralToken, you will retrieve the current block when the setCollateralToken executed.

surangap: yeah true. https://github.com/DeFiCh/ain/blob/13cc9108141fdf4ed0bf93adf8b4b61fd516ebd0/src/masternodes/mn_checks.cpp#L1830
ignore.

surangap (dismissed):

packages/jellyfish-api-core/src/category/loan.ts
jingyi2811: Anyway I improve the test I wrote earlier

Jouzo (approved):

surangap (approved):

packages/jellyfish-api-core/src/category/loan.ts
```diff
@@ -131,6 +140,17 @@ export interface SetCollateralToken {
   activateAfterBlock?: number
 }
 
+export interface CollateralTokensData {
+  [key: string]: CollateralTokenDetail
```
fuxingloh, line 144: node side should probably change this to array

cc @surangap

fuxingloh (approved):

packages/jellyfish-api-core/src/category/loan.ts
Jouzo: Diego is currently working on changing list* to return arrays

PR #554, opened 2021-08-03
Add setDefaultLoanScheme rpc

jingyi2811:

/kind feature

Add setDefaultLoanScheme rpc

Fixes part of #488


packages/jellyfish-api-core/__tests__/category/loan/setDefaultLoanScheme.test.ts
```diff
@@ -0,0 +1,132 @@
...
+      }
+    }
+
+    const utxos = await testing.container.call('listunspent', [1, 9999999, [masternodes[masternodeId].ownerAuthAddress]])
+    const inputs: UTXO[] = utxos.map((utxo: UTXO) => {
+      return {
+        txid: utxo.txid,
+        vout: utxo.vout
+      }
+    })
```
Jouzo, line 107: Less complex with 
```
const utxo = await testing.container.fundAddress(await container.call('getnewaddress'), 10)
```

packages/jellyfish-api-core/src/category/loan.ts
```diff
@@ -0,0 +1,30 @@
...
+    this.client = client
+  }
+
+  /**
+   * Sets the default loan scheme.
+   *
+   * @param {string} id Unique identifier of the loan scheme, max 8 chars
+   * @param {UTXO[]} [utxos = []] Specific UTXOs to spend
+   * @param {string} [utxos.txid] Transaction Id
+   * @param {number} [utxos.vout] Output number
```
surangap, line 19:
```suggestion
   * @param {string} utxos.txid Transaction Id
   * @param {number} utxos.vout Output number
```

surangap (commented):
minor docs update

packages/jellyfish-api-core/__tests__/category/loan/setDefaultLoanScheme.test.ts
```diff
@@ -0,0 +1,115 @@
...
+    const loanSchemeId = await testing.container.call('destroyloanscheme', ['scheme3', 120])
+    expect(typeof loanSchemeId).toStrictEqual('string')
+    expect(loanSchemeId.length).toStrictEqual(64)
+    await testing.generate(1)
+
+    const promise = testing.rpc.loan.setDefaultLoanScheme('scheme3')
+    await expect(promise).rejects.toThrow('RpcApiError: \'Test DefaultLoanSchemeTx execution failed:\nCannot set scheme3 as default, set to destroyed on block 120\', code: -32600, method: setdefaultloanscheme')
+
+    // Delete at block 120
+    await testing.container.waitForBlockHeight(120)
```
surangap, line 87: redundant.

surangap (commented):
minor. rest looks good.

packages/jellyfish-api-core/__tests__/category/loan/setDefaultLoanScheme.test.ts
jingyi2811: If I don't do so, the code written in this `it` maybe is going to be executed in the next `it` depending on how many block generated. It is a clean up code, to make sure everything set in this `it` executed here

surangap: I see. ideally, tests(`it()`) should not affect each other. but since we are using the same container, this might be hard to achieve fully. but we try to minimize the influence on each other and to write more independent tests. 
In your case, the cleanup code should be inside an `afterEach` block. but here the cleanup depends on the height. not sure if we can do something like this https://github.com/DeFiCh/jellyfish/blob/a980b3b2aa15c6f13f8380aa366333b5d11b1133/packages/jellyfish-api-core/__tests__/category/icxorderbook/icx_setup.ts#L119

If not, you can keep it since changing the test execution order does not affect other tests.
or else you can change the comparison logic(`expects`) in other tests.

jingyi2811: Ya Let's see what Fuxing think

Jouzo (approved):

packages/jellyfish-api-core/__tests__/category/loan/setDefaultLoanScheme.test.ts
jingyi2811: I will apply the same changes to all my rpcs and dftxs

fuxingloh: I agree with @surangap best case scenario is always separating them, I would just create another container for this test. Writing it like this works for now but is difficult to maintain in the long run. Culturally if we take an extreme stance of writing independent tests it makes it an authoritative norm that all tests must be independent. A win-win-win for everyone that maintains this codebase.

But I am confused about the utility of destroying a loan scheme, wondering what purpose does it serve. Can you even destroy a loan scheme with an active loan?

fuxingloh: Found the reasoning in https://github.com/DeFiCh/pinkpaper/tree/main/loan

I will merge these, for @surangap point should tackle in your next PRs

fuxingloh (approved):

packages/jellyfish-api-core/__tests__/category/loan/setDefaultLoanScheme.test.ts
jingyi2811: Key sentence: Any loans under the scheme will be moved to default loan scheme after destruction.

PR #555, opened 2021-08-03
`@defichain/jellyfish-wallet` derivation refactoring

fuxingloh:

 
/kind feature

`@defichain/jellyfish-wallet-mnemonic` to use a unique seed of `"@defichain/jellyfish-wallet-mnemonic"` different from the default of `"Bitcoin seed"`.
- This will ultimately lock our implementation to be only usable with our library. This means the `24 phrases` will only work with the jellyfish libraries. This prevents users from attempting to export/import phrases from other libraries and creating a compatibility nightmare for us due to incompatible derivation techniques. We could intervene in UX but better if done and dealt with at the implementation core.
- With our own unique seed, a stolen 24 phrase will not be as easy to crack since the stealer will need to know what type of wallet this mnemonic phrase belongs to. (Not the main purpose of this refactor)

`@defichain/jellyfish-wallet` to include `COINTYPE` via `1129/{account}/0/0`
- To allow `jellyfish-wallet` to be agnostic from coin-type for ICX adoption.

### DIFF
--- a/packages/jellyfish-wallet/src/wallet.ts
+++ b/packages/jellyfish-wallet/src/wallet.ts
@@ -24,7 +24,7 @@ export class JellyfishWallet<Account extends WalletAccount, HdNode extends Walle
    * @return Promise<WalletAccount>
    */
   get (account: number): Account {
-    const path = `${account}/0/0`
+    const path = `1129/${account}/0/0`
     const node = this.nodeProvider.derive(path)
     return this.accountProvider.provide(node)
   }
--- a/package-lock.json
+++ b/package-lock.json
@@ -13290,6 +13290,15 @@
         "@types/node": "*"
       }
     },
+    "node_modules/@types/create-hmac": {
+      "version": "1.1.0",
+      "resolved": "https://registry.npmjs.org/@types/create-hmac/-/create-hmac-1.1.0.tgz",
+      "integrity": "sha512-BNYNdzdhOZZQWCOpwvIll3FSvgo3e55Y2M6s/jOY6TuOCwqt3cLmQsK4tSmJ5fayDot8EG4k3+hcZagfww9JlQ==",
+      "dev": true,
+      "dependencies": {
+        "@types/node": "*"
+      }
+    },
     "node_modules/@types/docker-modem": {
       "version": "3.0.1",
       "resolved": "https://registry.npmjs.org/@types/docker-modem/-/docker-modem-3.0.1.tgz",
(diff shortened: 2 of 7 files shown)


packages/jellyfish-wallet/src/wallet.ts
```diff
@@ -24,7 +24,7 @@ export class JellyfishWallet<Account extends WalletAccount, HdNode extends Walle
    * @return Promise<WalletAccount>
    */
   get (account: number): Account {
-    const path = `${account}/0/0`
+    const path = `1129/${account}/0/0`
```
canonbrother, line 27: hmm.. not quite follow wallet stuff
wondering is this notation correct?
```{purpose}/{account}/{coint}/{index}```

ivan-zynesis: Irrelevant, this is bitcoin hdnode specific

ivan-zynesis (approved):
Need manual test on wallet, can't tell if any dep can be missing is nodeify

packages/jellyfish-wallet/src/wallet.ts
fuxingloh: 1. `1129` is from https://github.com/satoshilabs/slips/blob/master/slip-0044.md
2. account denoting account
3. 0 donating `internal/external` (change address)
4. address within account

But yeah, we are off track from default to make it easier to implement encrypted wallet.

canonbrother (approved):

fuxingloh:
There were some discussions that when into this which resulted in the refactoring. BIP-32 and BIP-44 are written for full node implementation, when used in a full implementation where hardened keys can easily be managed it makes sense. 

However, the hardened node makes it rather insecure in light implementations where `xprivkey` is required for Script resolution. This forces you to hold on to a private key on REST which is less than ideal. This can easily be solved with wallet.dat but that just makes it less sane to manage. This was necessitated by the wallet as just in time decryption was the only secure implementation and refactoring was the only sane move. Security above anything.

As for `bitcoind` and `defid`, they don't follow the BIP-44 specification since it's not been an approved proposal since 2014. Therefore maintaining compatibility with BIP-44 is opinionated as no one follows it either.

As for the BIP-32 specification, the above justify the change and we decided to a small adjustment for the jellyfish ecosystem's sake. 

Jellyfish project is a 30+ package large project, the philosophy was to always strive to implement a better JS library for modern crypto development. Attempting to manage compatibility with older libraries and specifications while innovating is just not going to work for us.

PR #556, opened 2021-08-03
inconsistent `maxPrice` should parse from buffer into consistent variant regardless

fuxingloh:

/kind fix

Because maxPrice consistency structure is not enforced at the transaction level, `jellyfish-transaction` must be able to parse without erroring out. This PR will convert inconsistent maxPrice representation into the consistent variant.


fuxingloh:
@Jouzo @canonbrother @ivan-zynesis FYI, I am going to merge this is in now as I need it on `DeFiCh/whale`.

Issue #431, opened 2021-06-29
`@defichain/testing` QOL utilities and setups


fuxingloh:
On the point of not using existing `client.xx.xx` where they are already `typed`, we don't need to be so strict about it. The best-case scenario would total isolation of concerns in each test case (file in this case) but that would just be overkill. If there is an existing typed method from the client. You can just use it. But for some obvious scenarios where we need isolation, e.g. it's not yet typed but we need assert it.

TLDR: it's okay to use `client.account.getAccount` if it's already merged instead of `container.call('getaccount', [])`.

> We need to look into this for greater testing QOL, but right now we don't have resources for it.

/area testing
/kind refactor

_Originally posted by @fuxingloh in https://github.com/DeFiCh/jellyfish/issues/399#issuecomment-870596525_

Issue #274, opened 2021-05-21
`jellyfish-transaction-builder` mock providers should use a light implementation


fuxingloh:
#### What would you like to be added:

Currently, it's attached to a full node via `setupMocks` which imports the priv key, this is a bad design.

https://github.com/DeFiCh/jellyfish/blob/63fd7700ef9959cc50bb2dadfc3cff202771aaab/packages/jellyfish-transaction-builder/__tests__/provider.mock.ts#L45-L48

Extremely bad design on my end, done for speed. We need to refactor this soon.

PR #566, opened 2021-08-06
Adds utility function to get txid's fixture

Jouzo:

/kind feature

Adds utility function to get txid's fixture in @defichain/testing

### DIFF
--- a/packages/testing/src/rawtx.ts
+++ b/packages/testing/src/rawtx.ts
@@ -40,6 +40,18 @@ export async function createSignedTxnHex (
   return signed.hex
 }
 
+/**
+ * Get fixture for a txid
+ *
+ * @param {MasterNodeRegTestContainer} container
+ * @return {Promise<string>} Fixture as hex string
+ */
+export async function getFixture (container: MasterNodeRegTestContainer, txid: string): Promise<string> {
+  const rawtx = await container.call('getrawtransaction', [txid])
+  const decoded = await container.call('decoderawtransaction', [rawtx])
+  return decoded.vout[0].scriptPubKey.hex
+}
+
 export interface CreateSignedTxnHexOptions {
   aEllipticPair: EllipticPair
   bEllipticPair: EllipticPair
--- a/packages/testing/src/rawtx.ts
+++ b/packages/testing/src/rawtx.ts
@@ -44,6 +44,7 @@ export async function createSignedTxnHex (
  * Get fixture for a txid
  *
  * @param {MasterNodeRegTestContainer} container
+ * @param {string} txid
  * @return {Promise<string>} Fixture as hex string
  */
 export async function getFixture (container: MasterNodeRegTestContainer, txid: string): Promise<string> {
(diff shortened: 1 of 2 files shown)


packages/testing/src/rawtx.ts
```diff
@@ -40,6 +40,18 @@ export async function createSignedTxnHex (
   return signed.hex
 }
 
+/**
+ * Get fixture for a txid
+ *
+ * @param {MasterNodeRegTestContainer} container
```
jingyi2811, line 46: Should be 2 params, right?

jingyi2811:
Great, next time I can call this function

jingyi2811 (dismissed):

packages/testing/src/rawtx.ts
```diff
@@ -40,6 +40,19 @@ export async function createSignedTxnHex (
...
+ * @param {string} txid
+ * @return {Promise<string>} Fixture as hex string
+ */
+export async function getFixture (container: MasterNodeRegTestContainer, txid: string): Promise<string> {
```
fuxingloh, line 50: Too ambiguous, should rename to exactly what it does

fuxingloh (changes requested):

packages/testing/src/rawtx.ts
jingyi2811: Agree

packages/testing/src/rawtx.ts
```diff
@@ -40,6 +40,19 @@ export async function createSignedTxnHex (
...
+ * Get fixture as hex string for a txid
+ *
+ * @param {MasterNodeRegTestContainer} container
+ * @param {string} txid
+ * @return {Promise<string>} Fixture as hex string
+ */
+export async function getFixtureAsHexString (container: MasterNodeRegTestContainer, txid: string): Promise<string> {
+  const rawtx = await container.call('getrawtransaction', [txid])
+  const decoded = await container.call('decoderawtransaction', [rawtx])
+  return decoded.vout[0].scriptPubKey.hex
```
jingyi2811, line 53: Could you do a quick testing. Seems like 1 rpc is enough to get the same result

```
  const rawtx = await container.call('getrawtransaction', [result, true])
  console.log(rawtx.vout[0].scriptPubKey.hex)
```

fuxingloh:
I was talking to @surangap about this today. I will create and refactor a centralized testing utility in the coming day. We have too many tools and utilities all over the place now. Highly repetitive and not DRY enough. I will not merge this in, for now, you just create it locally within the package you need it in, but keep it drafted. I will see your use case and create the relevant tools for you when I start on it.

For reference tickets that span multiple repos:
- https://github.com/DeFiCh/jellyfish/issues/431
- https://github.com/DeFiCh/jellyfish/issues/274
- https://github.com/DeFiCh/whale/issues/97
- https://github.com/DeFiCh/whale/issues/89

fuxingloh:
Closed, implemented in https://github.com/DeFiCh/jellyfish/pull/576

PR #576, opened 2021-08-10
`@defichain/jellyfish-testing` new centralized testing framework for jellyfish ecosystem

### LINKED ISSUE #431: `@defichain/testing` QOL utilities and setups

fuxingloh:

/kind feature

This package introduces a centralized testing framework for the jellyfish replacing the existing `@defichain/testing`.
All methods within `@defichain/testing` have thus been deprecated.

Testing is an essential part of any serious quality software developer work. This package provides many abstractions for
various commonly used setup patterns for the DeFi blockchain. This keeps your testing setup DRY for repeated tests.

`@defichain/jellyfish-testing` relies on other jellyfish dependencies, they are cyclic dependant. However, within this
mono-repo, we don't need to declare `"devDependencies"` as they are mapped in `tsconfig.base.json`. This configuration
allows any packages within `jellyfish-*` to rely on `jellyfish-testing` without causing cyclic dependencies errors.

This new implementation warps a container on init via `Testing.create(container)` extending on top of rpc and
container. The expectations are other more complex fixtures will be added into `testing.fixture`. 

```ts
const container = new MasterNodeRegTestContainer()
const testing = Testing.create(container)

testing.rpc         // to access JSON RPC
testing.fixture     // for complex fixture setup e.g. poolpair
testing.rawtx       // for rawtx setup with reasonable defaults 
testing.token       // for token setup with reasonable defaults 
testing.poolpair    // for poolpair setup with reasonable defaults 
```

On the point of not using existing JSON RPC within tests, this is optional moving forward, please use them to keep it
DRY.

Fixes #431
Fixes #566 with `testing.rawtx.getDfTx(txid)`

#### Additional comments

In this PR, I only replaced certain setups for brevity, we will deprecate and slowly move towards using `jellyfish-testing` reclusively in days to come.


packages/jellyfish-testing/src/fixture.ts
```diff
@@ -0,0 +1,54 @@
...
+ * TestingFixture setup complex fixtures for jellyfish testing.
+ */
+export class TestingFixture {
+  constructor (
+    private readonly testing: Testing
+  ) {
+  }
+
+  async createPoolPair (options: TestingPoolPairCreateRequest): Promise<poolpair.PoolPairInfo> {
+    if (options.a.symbol !== 'DFI') {
```
jingyi2811, line 14: Is that better if you have `await this.testing.generate(1)` for every calls?

packages/jellyfish-testing/src/rawtx.ts
```diff
@@ -0,0 +1,122 @@
...
+  async sign (options: TestingRawTxSign): Promise<string> {
+    const { hex, privKeys } = options
+
+    const mapped = await asPrivKey(privKeys)
+    const signed = await this.rpc.rawtx.signRawTransactionWithKey(hex, mapped)
+    return signed.hex
+  }
+
+  /**
+   * @param {string} txid to of a DfTx
```
jingyi2811, line 45: Maybe you can remove the comment section like other functions in this file.
If not, the return type should be `Promise<TestingRawTxDfTx>`

packages/jellyfish-testing/src/fixture.ts
canonbrother: hmm.. seems its intention, is tit better manageable for multi-nodes??

fuxingloh: Yeah, it is intentional as block generation is quite expensive time-wise. (You can't generate too far into the future)

By design `fixture.ts` should only generate blocks they need, so we don't waste expensive testing cycle. As this is a shared fixture for the entirety of the jellyfish testing ecosystem.

canonbrother (approved):

jingyi2811 (approved):

surangap (approved):
nice :clap:

PR #590, opened 2021-08-11
Adds ICX to jellyfish-testing

Jouzo:

/kind feature

Adds ICX setup and testing functionnalities to jellyfish-testing

### DIFF
--- a/packages/jellyfish-testing/__tests__/icxorderbook.test.ts
+++ b/packages/jellyfish-testing/__tests__/icxorderbook.test.ts
@@ -1,0 +1,206 @@
+import { MasterNodeRegTestContainer } from '@defichain/testcontainers'
+import { Testing } from '@defichain/jellyfish-testing'
+import BigNumber from 'bignumber.js'
+import { ICXOrderStatus, ICXOrderType, ICXHTLCType } from '../src/icxorderbook'
+
+const testing = Testing.create(new MasterNodeRegTestContainer())
+
+describe('ICX setup', () => {
+  beforeAll(async () => {
+    await testing.container.start()
+    await testing.container.waitForWalletCoinbaseMaturity()
+  })
+
+  afterAll(async () => {
+    await testing.container.stop()
+  })
+
+  it('should setup icx and init all public fields', async () => {
+    await testing.icxorderbook.setup()
+
+    expect(testing.icxorderbook.idDFI).toStrictEqual('0')
+    expect(testing.icxorderbook.idBTC).toStrictEqual('1')
+    expect(testing.icxorderbook.ICX_TAKERFEE_PER_BTC).toStrictEqual(0.001)
+    expect(testing.icxorderbook.DEX_DFI_PER_BTC_RATE).toStrictEqual(100)
+    expect(testing.icxorderbook.symbolDFI).toStrictEqual('DFI')
+    expect(testing.icxorderbook.symbolBTC).toStrictEqual('BTC')
+    expect(typeof testing.icxorderbook.accountDFI).toStrictEqual('string')
+    expect(testing.icxorderbook.accountDFI.length).toBeGreaterThan(0)
+    expect(typeof testing.icxorderbook.accountBTC).toStrictEqual('string')
+    expect(testing.icxorderbook.accountBTC.length).toBeGreaterThan(0)
+    expect(typeof testing.icxorderbook.poolOwner).toStrictEqual('string')
+    expect(testing.icxorderbook.poolOwner.length).toBeGreaterThan(0)
+  })
+})
+
+describe('ICX', () => {
+  beforeAll(async () => {
+    await testing.container.start()
+    await testing.container.waitForWalletCoinbaseMaturity()
+    await testing.icxorderbook.setup()
+  })
+
+  afterAll(async () => {
+    await testing.container.stop()
+  })
+
+  it('should createDFISellOrder', async () => {
+    const createOrder = {
+      receivePubkey: '037f9563f30c609b19fd435a19b8bde7d6db703012ba1aba72e9f42a87366d1941',
+      amountFrom: new BigNumber(15),
+      orderPrice: new BigNumber(0.01)
+    }
+    const { createOrderTxId } = await testing.icxorderbook.createDFISellOrder(createOrder)
+    const listOrders = await testing.rpc.icxorderbook.listOrders()
+    const order = listOrders[createOrderTxId]
+
+    expect(order).toStrictEqual({
+      status: ICXOrderStatus.OPEN,
+      type: ICXOrderType.INTERNAL,
+      tokenFrom: testing.icxorderbook.symbolDFI,
+      chainTo: 'BTC',
+      receivePubkey: createOrder.receivePubkey,
+      ownerAddress: testing.icxorderbook.accountDFI,
+      amountFrom: createOrder.amountFrom,
+      amountToFill: createOrder.amountFrom,
+      orderPrice: createOrder.orderPrice,
+      amountToFillInToAsset: createOrder.amountFrom.multipliedBy(createOrder.orderPrice),
+      height: expect.any(BigNumber),
+      expireHeight: expect.any(BigNumber)
+    })
+  })
+
+  it('should createDFIBuyOffer', async () => {
+    const createOrder = {
+      receivePubkey: '037f9563f30c609b19fd435a19b8bde7d6db703012ba1aba72e9f42a87366d1941',
+      amountFrom: new BigNumber(15),
+      orderPrice: new BigNumber(0.01)
+    }
+    const { createOrderTxId } = await testing.icxorderbook.createDFISellOrder(createOrder)
+
+    const makeOffer = {
+      orderTx: createOrderTxId,
+      amount: new BigNumber(0.10)
+    }
+    const { makeOfferTxId } = await testing.icxorderbook.createDFIBuyOffer(makeOffer)
+
+    const listOrders = await testing.rpc.icxorderbook.listOrders({ orderTx: createOrderTxId })
+    const offer = listOrders[makeOfferTxId]
+
+    expect(offer).toStrictEqual({
+      orderTx: createOrderTxId,
+      status: ICXOrderStatus.OPEN,
+      amount: makeOffer.amount,
+      amountInFromAsset: makeOffer.amount.dividedBy(createOrder.orderPrice),
+      ownerAddress: testing.icxorderbook.accountBTC,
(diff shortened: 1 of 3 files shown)


packages/jellyfish-testing/src/icxorderbook.ts
```diff
@@ -0,0 +1,314 @@
...
+  public accountBTC: string
+  public poolOwner: string
+  public idDFI: string
+  public idBTC: string
+  public ICX_TAKERFEE_PER_BTC: number
+  public DEX_DFI_PER_BTC_RATE: number
+
+  constructor (
+    private readonly testing: Testing,
+    private readonly container: MasterNodeRegTestContainer,
```
surangap, line 27: remove this. access it via `this.testing.container` throughout the source?

packages/jellyfish-testing/src/icxorderbook.ts
```diff
@@ -0,0 +1,314 @@
...
+    await this.mintBTCtoken(100)
+    await this.fundAccount(this.accountDFI, this.symbolDFI, 500)
+    await this.fundAccount(this.accountBTC, this.symbolDFI, 10) // for fee
+    await this.createBTCDFIPool()
+    await this.addLiquidityToBTCDFIPool(1, 100)
+    await this.setTakerFee(0.001)
+  }
+
+  async createAccounts (): Promise<void> {
+    this.accountDFI = await this.container.call('getnewaddress')
```
surangap, line 54: `this.testing` has a `generateAddress`

packages/jellyfish-testing/src/icxorderbook.ts
```diff
@@ -0,0 +1,314 @@
...
+      receivePubkey,
+      amountFrom,
+      orderPrice
+    }
+
+    const { txid: createOrderTxId } = await this.rpc.icxorderbook.createOrder(order)
+    await this.container.generate(1)
+
+    const ordersAfterCreateOrder = await this.rpc.icxorderbook.listOrders()
+    expect(ordersAfterCreateOrder[createOrderTxId].status).toStrictEqual(ICXOrderStatus.OPEN)
```
surangap, line 162: maybe remove this expect and line above since now we have a separate test file for these functions. and add the test there if not already.

surangap (changes requested):
some code can be moved to tests and refactor.

packages/jellyfish-testing/src/icxorderbook.ts
```diff
@@ -0,0 +1,314 @@
...
+  orderPrice: BigNumber
+}
+
+interface CreateDFIBuyOfferData {
+  orderTx: string
+  amount: BigNumber
+  ownerAddress?: string
+}
+
+interface CreateDFIBuyOfferData {
```
jingyi2811, line 288: Duplicate

packages/jellyfish-testing/src/icxorderbook.ts
```diff
@@ -0,0 +1,314 @@
...
+interface submitExtHTLCForDFIBuyOfferData {
+  offerTx: string
+  amount: BigNumber
+  hash: string
+  htlcScriptAddress: string
+  ownerPubkey: string
+  timeout?: number
+}
+
+export {
```
jingyi2811, line 310: Somehow feel weird export the variables which are imported from other file

packages/jellyfish-testing/src/icxorderbook.ts
```diff
@@ -0,0 +1,314 @@
+import { Testing } from './index'
+import BigNumber from 'bignumber.js'
+import { MasterNodeRegTestContainer } from '@defichain/testcontainers'
+import { JsonRpcClient } from '@defichain/jellyfish-api-jsonrpc'
+import { icxorderbook } from '@defichain/jellyfish-api-core'
+import { accountToAccount } from '@defichain/testing'
```
jingyi2811, line 6: We probably shouldn't use anything from `'@defichain/testing'` anymore

Jouzo: There is no `account` in jellyfish-testing yet.

jingyi2811: The code inside is not much

```
const to = options?.to ?? await getNewAddress(container)

  await container.call('accounttoaccount', [options.from, { [to]: `${amount.toString()}@${symbol}` }])
  await container.generate(1)

  return to
```

surangap: yes, should not use `'@defichain/testing'` going forward as it will add more refactoring to the future.
But at the same time can not add all that are not available in ` jellyfish-testing` in one go. so maybe add easy things on the go, if not use `container.call('accounttoaccount'` for now.

packages/jellyfish-testing/src/icxorderbook.ts
```diff
@@ -0,0 +1,242 @@
...
+      isDAT: true,
+      mintable: true,
+      tradeable: true,
+      collateralAddress: this.accountBTC
+    })
+    await this.testing.container.generate(1)
+  }
+
+  async initializeTokensIds (): Promise<void> {
+    let tokenInfo = await this.testing.container.call('gettoken', [this.symbolBTC])
```
surangap, line 60: can  use `this.testing.rpc.token.getToken()` here?
It's the same thing underneath, but good as a standard. so for RPCs that are not available yet on the jellyfish side, we could use `this.testing.container.call()`, otherwise `this.testing.rpc.`

surangap: https://github.com/DeFiCh/jellyfish/issues/431

packages/jellyfish-testing/__tests__/icxorderbook.test.ts
```diff
@@ -0,0 +1,250 @@
...
+    expect(typeof testing.icxorderbook.poolOwner).toStrictEqual('string')
+    expect(testing.icxorderbook.poolOwner.length).toBeGreaterThan(0)
+  })
+})
+
+describe('ICX', () => {
+  beforeAll(async () => {
+    await testing.container.start()
+    await testing.container.waitForWalletCoinbaseMaturity()
+    await testing.icxorderbook.setup()
```
surangap, line 42: this is okay for this test suite. But when refactoring ICX tests, it is better to list individual setup steps called here. It would give extra details about the setup without going into the icxorderbook.setup(). too much aggregation sometimes reduces context when it comes to testing. easy for the reader/reviewer.

packages/jellyfish-testing/__tests__/icxorderbook.test.ts
```diff
@@ -0,0 +1,250 @@
...
+      amountInEXTAsset: HTLC.amount.multipliedBy(createOrder.orderPrice),
+      hash: HTLC.hash,
+      timeout: new BigNumber(1440),
+      type: ICXHTLCType.DFC,
+      refundHeight: expect.any(BigNumber),
+      height: expect.any(BigNumber)
+    })
+  })
+
+  it('should createDFCHTLCForDFIBuyOffer', async () => {
```
surangap, line 139:
```suggestion
  it('should submitExtHTLCForDFIBuyOffer', async () => {
```

surangap (approved):
seems good now.

packages/jellyfish-testing/src/icxorderbook.ts
```diff
@@ -0,0 +1,243 @@
...
+}
+
+interface CreateDFCHTLCForDFIBuyOfferData {
+  offerTx: string
+  amount: BigNumber
+  hash: string
+  timeout?: number
+}
+
+interface submitExtHTLCForDFIBuyOfferData {
```
fuxingloh, line 236:
```suggestion
interface SubmitExtHTLCForDFIBuyOfferData {
```

packages/jellyfish-testing/src/icxorderbook.ts
```diff
@@ -0,0 +1,243 @@
...
+  }
+
+  async setup (): Promise<void> {
+    await this.createAccounts()
+    await this.createBTCToken()
+    await this.initializeTokensIds()
+    await this.mintBTCtoken(100)
+    await this.fundAccount(this.accountDFI, this.symbolDFI, 500)
+    await this.fundAccount(this.accountBTC, this.symbolDFI, 10) // for fee
+    await this.createBTCDFIPool()
```
fuxingloh, line 37:
```suggestion
```

A large part of the setup can be replaced with:
```ts
const pair = await testing.fixture.createPoolPair({
  a: { amount: '500', symbol: 'DFI' },
  b: { amount: '100', symbol: 'BTC' }
})
await testing.icxorderbook.setTakerFee(0.001)
```

Maybe we could just use this instead, since it's cleaner contextually what you are setting up also.

fuxingloh (changes requested):

packages/jellyfish-testing/src/icxorderbook.ts
fuxingloh: Contextually the only setup you are doing here is setting TakerFee, other than that it's creating the pool pair and we have an existing fixture for that.

Jouzo: I think we still need setup() in order to set 
```
    this.accountDFI = ''
    this.accountBTC = ''
    this.poolOwner = ''
    this.idDFI = ''
    this.idBTC = ''
    this.ICX_TAKERFEE_PER_BTC = new BigNumber(0)
    this.DEX_DFI_PER_BTC_RATE = new BigNumber(0)
```

packages/jellyfish-testing/src/icxorderbook.ts
```diff
@@ -0,0 +1,209 @@
...
+
+    await this.testing.fixture.createPoolPair({
+      a: { amount: '1', symbol: this.symbolBTC },
+      b: { amount: '100', symbol: this.symbolDFI }
+    })
+    this.DEX_DFI_PER_BTC_RATE = new BigNumber(100)
+
+    await this.initializeTokensIds()
+    await this.setTakerFee(new BigNumber(0.001))
+    await this.setBTCDFIPoolOwner()
```
surangap, line 44: can remove this also. the idea is that pool creation is separate from ICX, but the `DEX_DFI_PER_BTC_RATE ` which is set based on the pool is required for ICX inside defid.

surangap (changes requested):
some more refactor.

packages/jellyfish-testing/src/icxorderbook.ts
Jouzo: It seemed to me that the setup point is to set all public field, isn't it ?

surangap: most of them yes. but here if you check, pool owner is actually not used in the tests.

surangap (approved):

jingyi2811 (approved):

fuxingloh (approved):

Issue #533, opened 2021-08-01
`DfTx` - ICX


fuxingloh:
> Similar to https://github.com/DeFiCh/jellyfish/issues/108, this issue tracks the implementation of custom DfTx in `jellyfish-transaction` and `jellyfish-transaction-builder`. Must implement in `jellyfish-transaction` and test in `jellyfish-transaction-builder` for each PR.

### ICX DfTX

- [x] ICXCreateOrder: `'1'` @Jouzo  #608 
- [x] ICXMakeOffer: `'2'` @Jouzo #572
- [x] ICXSubmitDFCHTLC: `'3'` @surangap https://github.com/DeFiCh/jellyfish/pull/561
- [x] ICXSubmitEXTHTLC: `'4'` @surangap https://github.com/DeFiCh/jellyfish/pull/589
- [x] ICXClaimDFCHTLC: `'5'` @surangap  https://github.com/DeFiCh/jellyfish/pull/593
- [x] ICXCloseOrder: `'6'` @Jouzo #578 
- [ ] ICXCloseOffer: `'7'` @Jouzo #579

Issue #580, opened 2021-08-10
consolidation of ocean/whale/playground/salmon into `DeFiCh/jellyfish`


fuxingloh:
#### What would you like to be added:

Currently, we have 3 projects (ocean, whale, and playground) that are extensions of the jellyfish project. We should consolidate them into a singular code base so we don't need to manage dependencies updates across the ecosystem.

#### Why is this needed:

- better synergy of DeFiChain open source development across concerns
- single versioning
- single source of truth
- single documentation for the entirety via `jellyfish.defichain.com`
- early regression detection

/assign @fuxingloh 
/triage accepted

0ptim:
Does this mean that in the future, there will be one large SDK for calling RPC with a local full node as well as calling the ocean API or will these still be separate SDKs/libraries?

fuxingloh:
> Does this mean that in the future, there will be one large SDK for calling RPC with a local full node as well as calling the ocean API or will these still be separate SDKs/libraries?

Yup, one large SDK/Ecosystem that is divided into multiple smaller pieces within itself.

PR #608, opened 2021-08-20
 Added dftx ICXCreateOrder

Jouzo:

/kind feature

Adds support to create ICXCreateOrder custom transactions from jellyfish side.

Fixes part of #533


canonbrother (dismissed):

packages/jellyfish-transaction-builder/src/txn/txn_builder_icxorderbook.ts
```diff
@@ -0,0 +1,18 @@
+import { OP_CODES, Script, TransactionSegWit, ICXCreateOrder } from '@defichain/jellyfish-transaction'
+import { P2WPKHTxnBuilder } from './txn_builder'
+
+export class TxnBuilderICX extends P2WPKHTxnBuilder {
```
surangap, line 4: equivalent names as jellyfish-core
```suggestion
export class TxnBuilderICXOrderBook extends P2WPKHTxnBuilder {
```

packages/jellyfish-transaction-builder/src/index.ts
```diff
@@ -4,6 +4,7 @@ import { TxnBuilderUtxo } from './txn/txn_builder_utxo'
 import { TxnBuilderAccount } from './txn/txn_builder_account'
 import { TxnBuilderOracles } from './txn/txn_builder_oracles'
 import { TxnBuilderLiqPool } from './txn/txn_builder_liq_pool'
+import { TxnBuilderICX } from './txn/txn_builder_icxorderbook'
```
surangap, line 7:
```suggestion
import { TxnBuilderICXOrderBook } from './txn/txn_builder_icxorderbook'
```

packages/jellyfish-transaction-builder/src/index.ts
```diff
@@ -26,5 +28,6 @@ export class P2WPKHTransactionBuilder extends P2WPKHTxnBuilder {
   public readonly account = new TxnBuilderAccount(this.feeProvider, this.prevoutProvider, this.ellipticPairProvider)
   public readonly liqPool = new TxnBuilderLiqPool(this.feeProvider, this.prevoutProvider, this.ellipticPairProvider)
   public readonly oracles = new TxnBuilderOracles(this.feeProvider, this.prevoutProvider, this.ellipticPairProvider)
+  public readonly icxorderbook = new TxnBuilderICX(this.feeProvider, this.prevoutProvider, this.ellipticPairProvider)
```
surangap, line 31:
```suggestion
  public readonly icxorderbook = new TxnBuilderICXOrderBook(this.feeProvider, this.prevoutProvider, this.ellipticPairProvider)
```

surangap (approved):

packages/jellyfish-transaction/src/buffer/buffer_composer.ts
```diff
@@ -209,6 +209,32 @@ export abstract class ComposableBuffer<T> implements BufferComposer {
...
+        }
+      },
+      toBuffer: (buffer: SmartBuffer): void => {
+        const hex = getter() ?? ''
+        const buff: Buffer = Buffer.from(hex, 'hex')
+        writeVarUInt(buff.length, buffer)
+        buffer.writeBuffer(buff)
+      }
+    }
+  }
```
fuxingloh, line 236: The suggested change is more deterministic, best not to have multiple condition writing the same hex output.
Making `undefined` represent `0x00` and nothing else for a more bi-directional and consistent parsing.

```suggestion
  /**
   * HEX String with length specified, encoded into Buffer as the same order of the Hex String.
   * In short this read a hex and push it into the Buffer. It will not re-order the endian.
   *
   * When hex is `undefined` it will write `0x00` signifying zero byte length.
   *
   * @param getter to read HEX String. Writes its length then write the HEX string. Defaults to empty string.
   * @param setter to read ordered Buffer and set as the same ordered HEX String
   */
  static varUIntHex (getter: () => string | undefined, setter: (data: string | undefined) => void): BufferComposer {
    return {
      fromBuffer: (buffer: SmartBuffer): void => {
        const length = readVarUInt(buffer)
        if (length > 0) {
          const buff = Buffer.from(buffer.readBuffer(length))
          setter(buff.toString('hex'))
        }
      },
      toBuffer: (buffer: SmartBuffer): void => {
        const hex = getter()
        if (hex !== undefined ) {
          if (hex === '') { 
            throw new Error('ComposableBuffer.varUIntHex.toBuffer attempting to write empty buffer')
          }
          const buff: Buffer = Buffer.from(hex, 'hex')
          writeVarUInt(buff.length, buffer)
          buffer.writeBuffer(buff)
        } else {
          writeVarUInt(0x00, buffer)
        }
      }
    }
  }
```

packages/jellyfish-transaction/src/script/dftx/dftx_icxorderbook.ts
```diff
@@ -0,0 +1,50 @@
...
+  static OP_NAME = 'OP_DEFI_TX_ICX_CREATE_ORDER'
+
+  composers (cco: ICXCreateOrder): BufferComposer[] {
+    return [
+      ComposableBuffer.uInt8(() => cco.orderType, v => cco.orderType = v),
+      ComposableBuffer.varUInt(() => cco.tokenId, v => cco.tokenId = v),
+      ComposableBuffer.single<Script>(() => cco.ownerAddress, v => cco.ownerAddress = v, v => new CScript(v)),
+      ComposableBuffer.optionalVarUIntHex(() => cco.receivePubkey, v => cco.receivePubkey = v),
+      ComposableBuffer.satoshiAsBigNumber(() => cco.amountFrom, v => cco.amountFrom = v),
+      ComposableBuffer.satoshiAsBigNumber(() => cco.amountFrom, v => cco.amountFrom = v), // Represents amountToFill: how much is left to fill the order. In case of createOrder, amountToFill is always equal to amountFrom. See https://github.com/DeFiCh/ain/blob/ff53dcee23db2ffe0da9b147a0a53956f4e7ee31/src/masternodes/icxorder.h#L32
```
fuxingloh, line 45: Enforce this on transaction builder (by throwing an error when !=) instead of magically handling it on `CICXCreateOrder`. 

I think a serializer and deserializer should compose and decompose a buffer as it is without any logic. The above interface (`export interface ICXCreateOrder {`) should represent the expected buffer outcome without second guessing. This kind of handling creates intractable buffer management from a evolving code stance.

fuxingloh (changes requested):
Overall looks good with the two minor requested changes.

packages/jellyfish-transaction/src/buffer/buffer_composer.ts
Jouzo: Okay let me change it.
What about the naming though ? It does not follow the same logic as `varUIntUtf8BE`, `varUIntUtf8LE` and `varUIntArray`

fuxingloh: Maybe `varUIntOptionalHex` moving the optional back since `varUInt` is always present but if 0 it gives an `OptionalHex`

packages/jellyfish-transaction/src/script/dftx/dftx_icxorderbook.ts
surangap: or else simply we forget about the `icx_createorder ` interface.

Jouzo: Yes I handled it the same way as C++ side but I can add it to the interface and assert that `amountToFill` === `amountFrom`

Jouzo: > or else simply we forget about the `icx_createorder ` interface.

What do you mean by that ?

surangap: @Jouzo check the `icx_createorder ` rpc, it only has `amountFrom` field.

surangap: so when crating a new order, `amountToFill ` is redundant

Jouzo: Yes exactly, I did it this way to match the RPC interface. 
we can ignore it, add amountToFill to composer interface and assert that amountToFill equals amountFrom

fuxingloh:
Give me 1-2 days to think about this, this has some concern towards how we index data in the future.

packages/jellyfish-transaction/src/script/dftx/dftx_icxorderbook.ts
fuxingloh: I spend a few hours thinking about it, here is my train of thought. 

@surangap pointed out to me that on RPC, it's redundant. However, on DfTx, it's magically pushing redundant buffer into the chain. This was done because it was indexed as such on the storage layer. When a partially offer DfTx is created, the same DfTx is then updated with that partial amountToFill. I guess the same data structure is reused for a different context.

Honestly, not a big fan of this design on chain.

Personally, I rather have a separate indexing serialization layer and transaction sending serialization layer. This implementation creates a debate within me on how `jellyfish-indexer` (2022 Q1 goal) should be implemented whether we should share Transaction DfTx structure with data being indexed. I am standing on the side where we should separate.

However, back to the subject of DfTx. I believe we should still serialize and deserialize accordingly to the interface. The interface should present the buffer explicitly. Double writing and double reading, although make sense, creates unnecessary cognitive complexity and side-effect (think if something was reacting to reading or writing of buffers).

Jouzo: Regarding the DfTx, I agree that the serialisation should match the interface.
From the discussion we had, it was not clear for the reviewers where this `amountToFill` was coming from and it is much better to explicitly pass it and check that it is equal to `amountFrom`.
Ideally, the design on chain should be improved as well.

On the indexer matter, I don't grasp the full implication. What this separation would imply ?

fuxingloh: This is mainly towards jellyfish ecosystem play, as this PR was not fundamentally wrong in any way, it just requires some thinking on my end for aligning goals on how we should implement it #580. Merging this has a profound effect on data indexing.

---

> Providing additional context to the discussion.

As part of the transition https://github.com/DeFiCh/jellyfish/issues/580 (2022 Q4-Q1 long-term planning) where we consolidate whale/ocean/playground into a singular codebase. This transition moves the indexing responsibility to jellyfish. DeFiCh/ain indexes for consensus while jellyfish indexes for consumer/product facing use cases. For consensus, think UTXO unspent set for validating unspent. For product, think querying transaction activity tied to an address. 

Naturally for fast lookup, we have an LSM DB for sorted key querying. 

How it will looked like indexed:
- `[[Key: Data],[Key: Data],[Key: Data]]`
- `[[TxId: Masternode],[TxId: Masternode],[TxId: Masternode]]`

Which is essentially a DfTx:
- `[[TxId: CreateMasternodeDfTx],[TxId: CreateMasternodeDfTx],[TxId: CreateMasternodeDfTx]]`
- `[[TxId: CreatePoolPair],[TxId: CreatePoolPair],[TxId: CreatePoolPair]]`

On the `DeFiCh/whale` side, the indexes were designed rather quickly as we wanted to push out light wallet ASAP. (Build me a light wallet NOW!)
- Redundant double indexing on the whale project, bad
- Use of `JSON.stringify()` for data at REST, boo 👎, waste space
- Requires a high level of heuristic and implementation know-how to index and invalidate
- This leads to the creation of these tickets
  - https://github.com/DeFiCh/whale/issues/50
  - https://github.com/DeFiCh/whale/issues/51

The problems surfaced as fast as they could when Webster, Jimmy, and Ahmed contributed to the code base as development was slow and unnecessarily complex. Take a look at these and you could see how badly the code was, https://github.com/DeFiCh/whale/blob/main/src/module.indexer/model/dftx/update.oracle.ts it works but it sucks.

Pivoting to https://github.com/DeFiCh/jellyfish/issues/580, I did some thinking and I think we should fall back to indexing as similar as possible to how DeFiCh/ain does it. Look above, via `[Key: Data]`. **Where we use BufferComposer and store buffer as data in the indexes.**

Jumping to the subject on separation.

`DeFiCh/ain` indexes consensus protocol heavy object. Consensus protocol heavy objects are pretty straightforward, you index something with consistent lookup behavior. Think `txid`, always consistent E.g. `[TxId: DfTx]`: `[TxId: ICXCreateOrder]` 

Given you need to maintain a list of `DfTx` e.g. `ICX Order List`
Having the same DfTx, the structure allows a list of data as such: `[[TxId: ICXOrder], [TxId: ICXOrder]]`

Using the same DfTx abstraction: 
- `ICXCreateOrder extends ICXOrder` creates a new ICXOrder and add into `[[TxId: ICXOrder], [TxId: ICXOrder]]`
- `ICXUpdateOrder extends ICXOrder` with txId updates the same ICXOrder in `[[TxId: ICXOrder], [TxId: ICXOrder]]` via lookup.

`TxId` being as good as a random id, makes querying in a sorted list rather useless from the product API perspective. Meanless order.

This PR creates an inflection towards #580 (playing devil advocate with myself using newfound information from this PR), should we use the same `DfTx` for on-chain broadcasting and indexing or create another buffer interface purely for indexing. Initially, I wanted to just store `[MeaningfulKey: DfTx]`. This PR just signals the need for separation to design better consumer-friendly product APIs.

monstrobishi: @fuxingloh how does this impact indexers that don't rely directly on tx data, e.g. Aggregated Oracle Prices, Masternode stats, etc. Would those still live on jellyfish?

fuxingloh: Yup, leading up to that path, they will be on jellyfish too. The separations are inevitable where product-facing APIs need to index non-Tx-based data.

fuxingloh (approved):

PR #612, opened 2021-08-24
Add `createAnchor` rpc

canonbrother:

/kind feature

add `createAnchor` rpc

Fixes #

### DIFF
--- a/packages/jellyfish-api-core/__tests__/category/spv/createAnchor.test.ts
+++ b/packages/jellyfish-api-core/__tests__/category/spv/createAnchor.test.ts
@@ -1,0 +1,241 @@
+import { BigNumber, RpcApiError } from '@defichain/jellyfish-api-core'
+import { MasterNodeRegTestContainer, ContainerGroup, GenesisKeys } from '@defichain/testcontainers'
+import { ContainerAdapterClient } from '../../container_adapter_client'
+
+describe('Spv', () => {
+  const group = new ContainerGroup([
+    new MasterNodeRegTestContainer(GenesisKeys[0]),
+    new MasterNodeRegTestContainer(GenesisKeys[1]),
+    new MasterNodeRegTestContainer(GenesisKeys[2])
+  ])
+
+  const clients = [
+    new ContainerAdapterClient(group.get(0)),
+    new ContainerAdapterClient(group.get(1)),
+    new ContainerAdapterClient(group.get(2))
+  ]
+
+  beforeAll(async () => {
+    await group.start()
+
+    await setup()
+  })
+
+  afterAll(async () => {
+    await group.stop()
+  })
+
+  async function setMockTime (
+    clients: ContainerAdapterClient[], pastHour: number, futureHour = 0
+  ): Promise<void> {
+    const offset = Date.now() - (pastHour * 60 * 60 * 1000) + (futureHour * 60 * 60 * 1000)
+    for (let i = 0; i < clients.length; i += 1) {
+      await clients[i].misc.setMockTime(offset)
+    }
+  }
+
+  async function setup (): Promise<void> {
+    const anchorAuths = await group.get(0).call('spv_listanchorauths')
+    expect(anchorAuths.length).toStrictEqual(0)
+
+    // time travel back 13 hours ago
+    await setMockTime(clients, 13)
+
+    const anchorFreq = 15
+    for (let i = 0; i < anchorFreq; i += 1) {
+      const container = group.get(i % clients.length)
+      await container.generate(1)
+      await group.waitForSync()
+    }
+
+    {
+      const count = await group.get(0).getBlockCount()
+      expect(count).toStrictEqual(15)
+    }
+
+    // check the auth and confirm anchor mn teams at current height 15
+    await group.get(0).waitForAnchorTeams(clients.length)
+
+    // assertion for team
+    for (let i = 0; i < clients.length; i += 1) {
+      const container = group.get(i % clients.length)
+      const team = await container.call('getanchorteams')
+      expect(team.auth.length).toStrictEqual(clients.length)
+      expect(team.confirm.length).toStrictEqual(clients.length)
+      expect(team.auth.includes(GenesisKeys[0].operator.address))
+      expect(team.auth.includes(GenesisKeys[1].operator.address))
+      expect(team.auth.includes(GenesisKeys[2].operator.address))
+      expect(team.confirm.includes(GenesisKeys[0].operator.address))
+      expect(team.confirm.includes(GenesisKeys[1].operator.address))
+      expect(team.confirm.includes(GenesisKeys[2].operator.address))
+    }
+
+    {
+      // anchor auths length should be still zero
+      const auths = await group.get(0).call('spv_listanchorauths')
+      expect(auths.length).toStrictEqual(0)
+    }
+
+    // forward 3 hours from 12 hours ago and generate 15 blocks per hour
+    // set 3 hours because block height and hash chosen is then 3 hours
+    // then every 15 blocks will be matched again
+    for (let i = 1; i < 3 + 1; i += 1) {
+      await setMockTime(clients, 12, i)
+      await group.get(0).generate(15)
+      await group.waitForSync()
+    }
+
+    {
+      const count = await group.get(0).getBlockCount()
+      expect(count).toStrictEqual(60)
+    }
+
+    await group.get(0).waitForAnchorAuths(3)
+
+    // check each container should be quorum ready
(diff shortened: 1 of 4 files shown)


website/docs/jellyfish/api/spv.md
```diff
@@ -181,3 +181,30 @@ interface ListHtlcsOutputsResult {
   spent: SpentInfo
 }
 ```
+
+## createAnchor
+
+
```
jingyi2811, line 187: Should have `Create, sign and send anchor tx, using only spv api` as comment like what you wrote in the rpc

website/docs/jellyfish/api/spv.md
```diff
@@ -181,3 +181,30 @@ interface ListHtlcsOutputsResult {
...
+```ts title=client.spv.createAnchor()"
+interface spv {
+  createAnchor (
+    createAnchorInputs: CreateAnchorInput[], rewardAddress: string, options: CreateAnchorOptions = { send: true, feerate: 1000 }
+  ): Promise<CreateAnchorResult>
+}
+
+interface CreateAnchorOptions {
```
jingyi2811, line 195: Missing CreateAnchorInput interface

website/docs/jellyfish/api/spv.md
```diff
@@ -181,3 +181,30 @@ interface ListHtlcsOutputsResult {
...
+  send: boolean
+  feerate: BigNumber
```
jingyi2811, line 197: Your rpc side, feerate is a number

packages/jellyfish-api-core/__tests__/category/spv/createAnchor.test.ts
```diff
@@ -0,0 +1,241 @@
+import { BigNumber, RpcApiError } from '@defichain/jellyfish-api-core'
```
jingyi2811, line 1: import BigNumber from 'bignumber.js'

packages/jellyfish-api-core/__tests__/category/spv/createAnchor.test.ts
```diff
@@ -0,0 +1,241 @@
...
+    {
+      // anchor auths length should be still zero
+      const auths = await group.get(0).call('spv_listanchorauths')
+      expect(auths.length).toStrictEqual(0)
+    }
+
+    // forward 3 hours from 12 hours ago and generate 15 blocks per hour
+    // set 3 hours because block height and hash chosen is then 3 hours
+    // then every 15 blocks will be matched again
+    for (let i = 1; i < 3 + 1; i += 1) {
```
jingyi2811, line 82: Maybe can change i < 3 +1 to i < 4

canonbrother: the `3` is intentionally set.. better document

jingyi2811: Ah okay

packages/jellyfish-api-core/src/category/spv.ts
```diff
@@ -211,3 +237,40 @@ export interface ListHtlcsOutputsResult {
...
+  vout: number
+  /** Amount of output in satoshis (base unit) */
+  amount: number
+  /** WIF private key of bitcoin of signing this output */
+  privkey: string
+}
+
+export interface CreateAnchorOptions {
+  /** Send to BTC network */
+  send: boolean
```
Jouzo, line 254:
```suggestion
  send?: boolean
```

packages/jellyfish-api-core/src/category/spv.ts
```diff
@@ -211,3 +237,40 @@ export interface ListHtlcsOutputsResult {
...
+  /** Feerate (satoshis) per 1000 bytes */
+  feerate: number
```
Jouzo, line 256:
```suggestion
  feerate?: BigNumber
```

packages/jellyfish-api-core/src/category/spv.ts
```diff
@@ -129,6 +129,32 @@ export class Spv {
...
+   * @param {string} rewardAddress User's P2PKH address (in DeFi chain) for reward
+   * @param {CreateAnchorOptions} [options]
+   * @param {boolean} [options.send=true] Send it to BTC network
+   * @param {BigNumber} [options.feerate=1000] Feerate (satoshis) per KB, default is 1000.
+   * @return {Promise<CreateAnchorResult>}
+   */
+  async createAnchor (
+    createAnchorInputs: CreateAnchorInput[],
+    rewardAddress: string,
+    options: CreateAnchorOptions = { send: true, feerate: 1000 }
```
Jouzo, line 150: In that case, if you set only one key, the other will not default

packages/jellyfish-api-core/src/category/spv.ts
canonbrother: don't need bignumber on feerate as it is per satoshi.. its base unit

packages/jellyfish-api-core/src/category/spv.ts
canonbrother: @Jouzo  hmm... i don't quite get this... :thinking:

Jouzo: As both keys are optional and default to a value, one should be able to call it with `{ send: false }` and have it default to `{ send: false, feerate: 100 }`. In that case, if you call it with one of the two keys, the other one won't default.

canonbrother: yes.. understand.. i guess that `ain` has some reasons to default the keys
https://github.com/DeFiCh/ain/blob/ff53dcee23db2ffe0da9b147a0a53956f4e7ee31/src/spv/spv_rpc.cpp#L88-L89

Jouzo: Yes it defaults on `ain` side as well. 
I still think it is better to set it on jellyfish side not to have unexpected behaviour later on

packages/jellyfish-api-core/__tests__/category/spv/createAnchor.test.ts
```diff
@@ -0,0 +1,242 @@
...
+    }
+
+    {
+      // auths back to zero after the anchor active
+      const auths = await group.get(0).call('spv_listanchorauths')
+      expect(auths.length).toStrictEqual(0)
+    }
+  })
+
+  it('should be failed as min anchor quorum was not reached', async () => {
```
surangap, line 226: This test depends on `should createAnchor`. tests should be independent. maybe update the test so that first create an anchor. then within the same test try to create another anchor by the same node.

packages/jellyfish-api-core/__tests__/category/spv/createAnchor.test.ts
```diff
@@ -0,0 +1,242 @@
...
+    expect(anchor.defiHeight).toStrictEqual(15)
+    expect(anchor.estimatedReward).toStrictEqual(new BigNumber(0))
+    expect(anchor.cost).toStrictEqual(new BigNumber(3556))
+    expect(anchor.sendResult).toStrictEqual(0)
+    expect(anchor.sendMessage).toStrictEqual('Success')
+
+    {
+      const pending = await group.get(0).call('spv_listanchorspending')
+      // pending does not contain active property
+      expect(pending.active).toBeUndefined()
```
surangap, line 182: pending does not have `active` property anyways. so maybe instead check the length of the `pending` object?

packages/jellyfish-api-core/__tests__/category/spv/createAnchor.test.ts
```diff
@@ -0,0 +1,242 @@
...
+
+    {
+      // anchor auths length should be still zero
+      const auths = await group.get(0).call('spv_listanchorauths')
+      expect(auths.length).toStrictEqual(0)
+    }
+
+    // forward 3 hours from 12 hours ago and generate 15 blocks per hour
+    // set 3 hours because block height and hash chosen is then 3 hours
+    // then every 15 blocks will be matched again
```
surangap, line 82: sorry did not understand this properly. could you explain it a bit more or point me to any reference?

canonbrother: https://github.com/DeFiCh/ain/wiki/What-is-an-anchor%3F#anchor-data-creation

packages/jellyfish-api-core/src/category/spv.ts
canonbrother: i think i finally got what you mean
```
createSomething (details) {
  return { ...default, ...details }
}
```

PR #621, opened 2021-08-30
jellyfish-testing supports multinodes

canonbrother:

/kind feature

To support multinodes on jellyfish-testing

Fixes https://github.com/DeFiCh/jellyfish/issues/637


jingyi2811 (approved):

packages/jellyfish-testing/src/misc.ts
```diff
@@ -0,0 +1,15 @@
...
+import { MasterNodeRegTestContainer } from '@defichain/testcontainers'
+
+export class TestingMisc {
+  constructor (
+    private readonly container: MasterNodeRegTestContainer,
+    private readonly rpc: JsonRpcClient
+  ) {
+  }
+
+  async offsetTimeHourly (pastHour: number, futureHour = 0): Promise<void> {
```
fuxingloh, line 11:
```suggestion
  /**
   * @param ... offset can be a positive/negative number to offset into the future or past...
   */
  async offsetTimeHour (offsetBy: number): Promise<void> {
```

packages/jellyfish-testing/src/index.ts
```diff
@@ -81,3 +84,50 @@ class TestingJsonRpcClient extends JsonRpcClient {
...
 }
+
+export class TestingGroup {
+  private constructor (
+    public readonly group: ContainerGroup,
+    public readonly testings: Testing[]
+  ) {
+  }
+
+  static create (n = 1): TestingGroup {
```
fuxingloh, line 95:
```suggestion
  static create (
    n = 1, 
    init: (index: number) => MasterNodeRegTestContainer = new MasterNodeRegTestContainer(GenesisKeys[index])
  ): TestingGroup {
```

packages/jellyfish-testing/src/index.ts
```diff
@@ -81,3 +84,50 @@ class TestingJsonRpcClient extends JsonRpcClient {
...
+    const containers: MasterNodeRegTestContainer[] = []
+    const testings: Testing[] = []
+    for (let i = 0; i < n; i += 1) {
+      const container = new MasterNodeRegTestContainer(GenesisKeys[i])
```
fuxingloh, line 99:
```suggestion
      const container = init(i)
```

packages/jellyfish-testing/src/index.ts
```diff
@@ -81,3 +84,50 @@ class TestingJsonRpcClient extends JsonRpcClient {
...
+  }
+
+  async start (): Promise<void> {
+    return await this.group.start()
+  }
+
+  async stop (): Promise<void> {
+    return await this.group.stop()
+  }
+
```
fuxingloh, line 125:
```suggestion

async exec(runner: (testing: Testing) => Promise<void>): Promise<void> {
  for each testing in this.testings {
      runner(testing)
  }
}

```

allowing you to do this

```ts
testingGroup.exec((testing: Testing) => {
  testing.rpc.misc....
})
```

and it will automatically run on all container

fuxingloh (changes requested):

packages/jellyfish-testing/src/index.ts
surangap, line 95: @canonbrother How about if we address https://github.com/DeFiCh/jellyfish/issues/637 here?
we should be able to pass the containers from outside also as some tests may require it.
maybe we can pass an optional array of containers with `TestingGroup::create(n)`

surangap: leave the `Testing` class unchanged, as it would require existing test refactoring.

surangap: https://github.com/DeFiCh/jellyfish/pull/621/files#r703286221

canonbrother: yea updated the changes

packages/jellyfish-testing/src/index.ts
```diff
@@ -81,3 +84,56 @@ class TestingJsonRpcClient extends JsonRpcClient {
...
 }
+
+export class TestingGroup {
+  private constructor (
+    public readonly group: ContainerGroup,
+    public readonly testings: Testing[]
+  ) {
+  }
+
+  static create (n: number, init: (index: number) => MasterNodeRegTestContainer): TestingGroup {
```
surangap, line 95: We still need the ability to call `TestingGroup.create(3)` for default container types. i.e. `init()` should be optional. please check if there's a better syntax to write this. but this one works!
```suggestion
  static create (n: number, init: (index: number) => MasterNodeRegTestContainer = (index) => new MasterNodeRegTestContainer(GenesisKeys[index])): TestingGroup {
```

packages/jellyfish-api-core/__tests__/category/spv/listAnchors.test.ts
```diff
@@ -1,64 +1,47 @@
...
-    new MasterNodeRegTestContainer(GenesisKeys[1]),
-    new MasterNodeRegTestContainer(GenesisKeys[2])
-  ])
-
-  const clients = [
-    new ContainerAdapterClient(group.get(0)),
-    new ContainerAdapterClient(group.get(1)),
-    new ContainerAdapterClient(group.get(2))
-  ]
+  const tGroup = TestingGroup.create(3, i => new MasterNodeRegTestContainer(GenesisKeys[i]))
```
surangap, line 5: once https://github.com/DeFiCh/jellyfish/pull/621/files#r704077129 is done, you can call `TestingGroup.create(3)` here.

packages/jellyfish-testing/src/index.ts
canonbrother: yes will change to optional
but this one is the most simplified ady 
```init: (index: number) => MasterNodeRegTestContainer)```

Comment (surangap) [reply packages/jellyfish-testing/src/index.ts:95]: clarification needed.
then where is the default function body? what you are giving as `init: (index: number) => MasterNodeRegTestContainer)` is the type of the `init` and not the default function. am I missing something, thinking from c++ perspective 😅

Comment (canonbrother) [reply packages/jellyfish-testing/src/index.ts:95]: init: (index: number) => MasterNodeRegTestContainer = (index) => new MasterNodeRegTestContainer(GenesisKeys[index])
is same as 
init: (index: number) => MasterNodeRegTestContainer (a shortcut to write)

Comment (fuxingloh) [reply packages/jellyfish-testing/src/index.ts:95]: We still need the default functions, let me create a PR into this branch

Comment (canonbrother) [reply packages/jellyfish-testing/src/index.ts:95]: @fuxingloh i can update the changes here.. i haven't got idea of setting the default func there yet

Comment (canonbrother) [reply packages/jellyfish-testing/src/index.ts]: FYI,  setting up default func body is not allowed to set optional param
image

I refined the optional mnkeys logic for easier

File: packages/jellyfish-testing/src/index.ts [hunk 7]
```diff
@@ -81,3 +84,63 @@ class TestingJsonRpcClient extends JsonRpcClient {
...
+
+        const testing = Testing.create(container)
+        testings.push(testing)
+      }
+    }
+
+    const group = new ContainerGroup(containers)
+
+    return new TestingGroup(group, testings)
+  }
```
Jouzo, line 115: What about : 

```suggestion
  static create (n: number, Container: typeof MasterNodeRegTestContainer, mnKeys?: MasterNodeKey[]): TestingGroup {
    let containers: MasterNodeRegTestContainer[] = []
    let testings: Testing[] = []

    if (mnKeys !== undefined && mnKeys.length > 0) {
      containers = mnKeys.map(k => new Container(k))
      testings = containers.map(c => Testing.create(c))
    } else {
      for (let i = 0; i < n; i += 1) {
        const container = new MasterNodeRegTestContainer(GenesisKeys[i])
        containers.push(container)

        const testing = Testing.create(container)
        testings.push(testing)
      }
    }

    const group = new ContainerGroup(containers)

    return new TestingGroup(group, testings)
  }
```
to be able to call it as 
```
const group = TestingGroup.create(3, MultiOperatorGovernanceMasterNodeRegTestContainer, GenesisKeys)
```

canonbrother: great.. its more configurable

fuxingloh: Better to use function, as TypeScript might rename the symbols

packages/jellyfish-testing/src/index.ts
surangap: for this one do not need to put `?`, having the initializer means if you don't provide it outside, it will use the default.

packages/jellyfish-testing/src/index.ts
surangap: the final output, `TestingGroup.create(3)` should be available for a test that does not care about container type and the keys.

fuxingloh (approved):

packages/jellyfish-testing/src/index.ts
canonbrother: @Jouzo  i think i'm going backward :sweat_smile: .. btw @fuxingloh  helps solve the code ady.. thanks :pray:

Jouzo (approved):

PR #624, opened 2021-08-30
jellyfish-wallet to use `coinType/purpose/0/account` pathing

fuxingloh:
/kind feature

Updated JellyfishWallet to use path "coinType/purpose/0/account" non-hardened derivation with:
- `coinType` support  for BTC/DFI for ICX
- `purpose` support for LIGHT_WALLET and MASTERNODE locked UTXO.
- `account` for addresses within the purpose use case. E.g. light wallet addresses, masternodes.

This won't break compatibility with DeFiCh/wallet because currently we only use one default address.

Where account = `0`.

Before:
- `1129/account/0/0 -> 1129/0/0/0`

After:
- COIN_TYPE = 1129, default
- PURPOSE = 0, default
- `coinType/purpose/0/account -> 1129/0/0/0`

### DIFF
--- a/packages/jellyfish-wallet/__tests__/wallet.test.ts
+++ b/packages/jellyfish-wallet/__tests__/wallet.test.ts
@@ -25,7 +25,7 @@ describe('discover accounts', () => {
 
   it('should discover [] when [1] has activity', async () => {
     const accountProvider = new TestAccountProvider([
-      'bcrt1qn823x6sqpc66fuy2g0ylufe4sfaculfs87sulp'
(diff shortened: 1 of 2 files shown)


canonbrother (approved):

monstrobishi:
Account and purpose are swapped so the addresses coming out are different. This won't break existing users wallet?

monstrobishi:
Nevermind read the bottom of the description

monstrobishi (approved):

ivan-zynesis (approved):

PR #663, opened 2021-09-15
added 2 sample project in `/examples`

fuxingloh:

/kind chore

Added 2 sample projects in `/examples`:

1. `/ocean-dex-api` - Ocean DEX API - This is a very simple example to fetch live DEX Pool Pair information from ocean.defichain.com.
1. `/ocean-masternodes-api` - DeFiChain Masternode - A very simple example to fetch Masternode information from ocean.defichain.com with pagination.

These 2 are the first of many example projects that will be added to `DeFiCh/jellyfish`. As part of the consolidation effort https://github.com/DeFiCh/jellyfish/issues/580, this is a great place to put all and everything related to the jellyfish ecosystem. Be it an example project or general API/SDK documentation. 

We will be adding more projects of different use cases and calibrate over the coming weeks.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,6 +23,9 @@ parcel-bundle-reports
 
 # node_modules
 node_modules
+package-lock.json
+!/package-lock.json
+yarn.lock
 
 # dist & pack
 dist
--- a/examples/ocean-masternodes-api/pages/index.jsx
+++ b/examples/ocean-masternodes-api/pages/index.jsx
@@ -15,7 +15,7 @@ export default function Index ({
   const [token, setToken] = useState(nextToken)
 
   async function loadMore () {
-    if (next === undefined) {
+    if (token === undefined) {
       return
     }
     const response = await api.masternodes.list(60, token)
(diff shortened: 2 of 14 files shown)


.gitignore
```diff
@@ -23,6 +23,9 @@ parcel-bundle-reports
 
 # node_modules
 node_modules
+package-lock.json
```
jingyi2811, line 26: I feel like there is something wrong here

jingyi2811: I am unsure why you wrote `package-lock.json `and` !/package-lock.json`

examples/ocean-masternodes-api/pages/index.jsx
```diff
@@ -0,0 +1,75 @@
...
+
+export default function Index ({
+  masternodes,
+  nextToken
+}) {
+  const [list, setList] = useState(masternodes)
+  const [token, setToken] = useState(nextToken)
+
+  async function loadMore () {
+    if (next === undefined) {
```
jingyi2811, line 18: Is `next` a keyword in Next js?

fuxingloh: thanks! it's a mistake

.gitignore
fuxingloh: to prevent other `package-lock.json` from being committed as there should only be one `package-lock.json` at the root.

jingyi2811 (approved):

PR #695, opened 2021-09-29
Add listAuctions rpc

jingyi2811:

/kind feature

Add listAuctions rpc

Fixes part of #488


packages/testcontainers/src/containers/RegTestContainer/LoanContainer.ts
```diff
@@ -4,7 +4,7 @@ import { MasterNodeRegTestContainer } from './Masternode'
 
 export class LoanMasterNodeRegTestContainer extends MasterNodeRegTestContainer {
   constructor (masternodeKey: MasterNodeKey = RegTestFoundationKeys[0]) {
-    super(masternodeKey, 'defi/defichain:master-da0bbfa')
+    super(masternodeKey, 'defi/defichain:HEAD-c4fc84b')
```
canonbrother, line 7: huh?? why not refer to the latest??

packages/jellyfish-api-core/src/category/loan.ts
```diff
@@ -313,6 +313,23 @@ export class Loan {
...
+    return await this.client.call('listauctions', [], 'bignumber')
+  }
+}
+
+export interface AuctionDetail {
+  vaultId: string
+  batchCount: BigNumber
+  liquidationPenalty: BigNumber
+  liquidationHeight: BigNumber
+  batches: AuctionBatchDetails[]
```
canonbrother, line 332: 3 more fields
loanSchemeId
ownerAddress
state

packages/jellyfish-api-core/src/category/loan.ts
```diff
@@ -313,6 +313,23 @@ export class Loan {
...
     return await this.client.call('paybackloan', [metadata, utxos], 'number')
   }
+
+  /**
+   * List all available auctions.
+   *
+   * @return {Promise<AuctionDetail[]>}
+   */
+  async listAuctions (): Promise<AuctionDetail[]> {
+    return await this.client.call('listauctions', [], 'bignumber')
```
canonbrother, line 323:
```suggestion
    return await this.client.call('listauctions', [], { batches: 'bignumber' })
```

packages/jellyfish-api-core/__tests__/category/loan/listAuctons.test.ts
```diff
@@ -0,0 +1,182 @@
...
+      fixedIntervalPriceId: 'TSLA/USD'
+    })
+    await testing.generate(1)
+  })
+
+  afterAll(async () => {
+    await testing.container.stop()
+  })
+
+  it('should listAuctions', async () => {
```
canonbrother, line 68: one more pagination test

packages/testcontainers/src/containers/RegTestContainer/LoanContainer.ts
jingyi2811: Mine is the latest in `v2.0.x`. As it was merged to master, I shall change it to `master-da0bbfa`

packages/jellyfish-api-core/src/category/loan.ts
jingyi2811: Good suggestion.

jingyi2811:
Please don't review this PR anymore.
Like `listauctionhistory`, the pagination feature does not work as expected.

jingyi2811:
Pagination does not work as expected. List is able to be filtered by including_start, limit and height but unable to be filtered by vaultId

jingyi2811:
In general, the pagination feature works fine except there is a weird issue needs to be rectified later.

packages/jellyfish-api-core/__tests__/category/loan/listAuctons.test.ts
```diff
@@ -0,0 +1,521 @@
...
+      // including_start = false
+      {
+        const page = await testing.rpc.loan.listAuctions(
+          { start: { height: auction4.liquidationHeight } }
+        )
+        expect(page.length).toStrictEqual(0)
+      }
+    })
+
+    it('should listAuctions with vaultId', async () => {
```
jingyi2811, line 387: ListAuction with vaultId only won't work and this is not recommended. 
Added test unit here and show that the result is unpredictable.

packages/jellyfish-api-core/__tests__/category/loan/listAuctons.test.ts
```diff
@@ -0,0 +1,521 @@
...
+      {
+        const page = await testing.rpc.loan.listAuctions(
+          { start: { vaultId: auction1.vaultId, height: auction2.liquidationHeight } }
+        )
+        expect(page.length).toStrictEqual(2)
+        expect(page[0].vaultId).toStrictEqual(auction3.vaultId)
+        expect(page[1].vaultId).toStrictEqual(auction4.vaultId)
+      }
+
+      // List for liquidation height of third vault and vaultId of second vault
```
jingyi2811, line 447: This is not recommended too.
Added test unit here and show that the result is unpredictable.

packages/jellyfish-api-core/__tests__/category/loan/listAuctons.test.ts
```diff
@@ -0,0 +1,481 @@
...
+      {
+        const page = await testing.rpc.loan.listAuctions(
+          { start: { vaultId: auction1.vaultId } }
+        )
+        expect(page.length).toStrictEqual(3) // Unable to filter by vaultId only. Need to filter both vaultId and height together.
+      }
+    })
+
+    it('should listAuctions with vaultId and height', async () => {
+      // List for liquidation height of second vault and vaultId of second vault
```
jingyi2811, line 405: I do not add the test for liquidation height and vaultId that belongs to different vault (For eg: liquidation height of third vault and vaultId of second vault).  The reason is the result is unpredictable where different result is returned if you run the test for many times.

packages/jellyfish-api-core/__tests__/category/loan/listAuctons.test.ts
jingyi2811: Added the test unit here because the return result is predictable. The result is filtered by including_start only.

ivan-zynesis: Looks like some misunderstanding here, don't see anything wrong in the RPC behavior.

All 4 listed data are having the correct vaultId right?
The including_start mean using the offset itself like (greater than vs greater than and equal to)

docs/node/RPC Category/16-loan.md
```diff
@@ -549,7 +549,7 @@ interface loan {
 
 interface TakeLoanMetadata {
   vaultId: string
-  amounts: string | string[] // amount@symbol
+  amounts: string // amount@symbol
```
surangap, line 552:
```suggestion
  amounts: string | string[] // amount@symbol
```
takeloan support multiple tokens at once

packages/jellyfish-api-core/__tests__/category/loan/listAuctons.test.ts
surangap: pagination and filtering are two different things. Here can not filter by vault id.
<heigh, vault id> is only for pagination to give the start point in the DB. Hence should be considered together. db has the combined index <heigh, vault id>  for this.

surangap: we can remove this test I think😛

packages/jellyfish-api-core/__tests__/category/loan/listAuctons.test.ts
surangap: not sure the reason for this. But when using height and vault id from different auctions, basically the entry point that we are giving is non-existent. Not sure how it behaves with a combined index.

packages/jellyfish-api-core/__tests__/category/loan/listAuctons.test.ts
ivan-zynesis: I think the test is correct though, it is testing the `vaultId` as offset not filter. But the response simply doesn't seems to respect the offset correctly.

full: [a, b, c, d]
start from "b" still include [a, b, c, d]

seems like a RPC bug to me

docs/node/RPC Category/16-loan.md
jingyi2811: HAHAHA TakeLoanMetadata not related this this PR right? Will fix anyway.

packages/jellyfish-api-core/__tests__/category/loan/listAuctons.test.ts
surangap: can not use `vault id` only.
when you use that the height has the default value of 0; The index is  <height, vault ID>, not other way around as per current implementation.
```
{
        READWRITE(WrapBigEndian(height));
        READWRITE(vaultId);
    }
```
so in this case it will include all the auctions in the db

jingyi2811: Changed to `auction2.vaultId`
In my head, I have been questioning myself whether should I add this test or not for 10 times.

surangap: not sure why they changed the order though😛
@Jouzo

Jouzo: The logic is that you can have multiple vaults per height but not multiple heights per vault

Jouzo: It was <height, vaultId> from the start. It was just accidently inverted during the refactor form std::pair to struct

docs/node/RPC Category/16-loan.md
surangap: yea, but this PR updates that structure. should be a mistake.

packages/jellyfish-api-core/__tests__/category/loan/listAuctons.test.ts
surangap: We can keep the test since it expresses the actual behaviour of the RPC. But let's change the name
```suggestion
    it('should not filter listAuctions with vaultId only', async () => {
```

jingyi2811: Hesitate to change, but I am going to change anyway. For the past test units, we use the words `Should not` ONLY when the exception thrown.

jingyi2811: Fixed

surangap: okay, but the current name `should listAuctions with vaultId` does not suit the test content coz it implies it's working. can change to another more appropriate.

surangap (approved):

packages/jellyfish-api-core/src/category/loan.ts
```diff
@@ -378,6 +378,24 @@ export class Loan {
...
   }
+
+  /**
+   * List all available auctions.
+   *
+   * @param {AuctionPagination} pagination
+   * @param {AuctionPaginationStart} [pagination.start]
+   * @param {string} [pagination.start.vaultId]
+   * @param {number} [pagination.start.height]
+   * @param {boolean} [pagination.including_start]
```
canonbrother, line 389: thinking will remove `including_start` and do optional on height in future :thinking:

jingyi2811: You mean in the C++ side, they should not have including_start right?

canonbrother (approved):

packages/jellyfish-api-core/src/category/loan.ts
canonbrother: we have `including_start` very early.. so just leave it for now

jingyi2811: I won't add this to whale anyway.

jingyi2811: But, the funny thing is listAuctionHistory RPC does not have including_start. LOL

fuxingloh (approved):

PR #661, opened 2021-09-14
Add `depositToVault`, `takeLoan` rpc

canonbrother:
#### What kind of PR is this?:

/kind feature

#### What this PR does / why we need it:
Add `depositToVault`, `takeLoan` rpc

#### Which issue(s) does this PR fixes?:

Fixes part of #488

PR #699, opened 2021-09-30
Add `loanPayback` rpc

canonbrother:

/kind feature

Add `loanPayback` rpc

Fixes part of #488

### DIFF
--- a/packages/jellyfish-api-core/__tests__/category/loan/loanPayback.test.ts
+++ b/packages/jellyfish-api-core/__tests__/category/loan/loanPayback.test.ts
@@ -1,0 +1,433 @@
+import { LoanMasterNodeRegTestContainer } from './loan_container'
+import { GenesisKeys } from '@defichain/testcontainers'
+import BigNumber from 'bignumber.js'
+import { TestingGroup } from '@defichain/jellyfish-testing'
+import { RpcApiError } from '@defichain/jellyfish-api-core'
+
+const tGroup = TestingGroup.create(3, i => new LoanMasterNodeRegTestContainer(GenesisKeys[i]))
+let vaultId: string
+let vaultAddress: string
+let vaultId1: string
+let vaultAddress1: string
+let liqVaultId: string
+let loanTokenAddress: string
+
+async function setup (): Promise<void> {
+  // token setup
+  const collateralAddress = await tGroup.get(0).container.getNewAddress()
+  await tGroup.get(0).token.dfi({ address: collateralAddress, amount: 100000 })
+  await tGroup.get(0).generate(1)
+  await tGroup.get(0).token.create({ symbol: 'BTC', collateralAddress })
+  await tGroup.get(0).generate(1)
+  await tGroup.get(0).token.mint({ symbol: 'BTC', amount: 30000 })
+  await tGroup.get(0).generate(1)
+
+  // oracle setup
+  const addr = await tGroup.get(0).generateAddress()
+  const priceFeeds = [
+    { token: 'DFI', currency: 'USD' },
+    { token: 'BTC', currency: 'USD' },
+    { token: 'TSLA', currency: 'USD' },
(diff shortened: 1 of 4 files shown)


packages/jellyfish-api-core/__tests__/category/loan/takeLoan.test.ts
```diff
@@ -114,69 +162,61 @@ describe('Loan', () => {
...
 
-      await tGroup.get(0).container.call('updateloantoken', ['TSLA', { mintable: true }])
-      await tGroup.get(0).generate(1)
-      await tGroup.waitForSync()
+    afterAll(async () => {
+      await tGroup.stop()
     })
 
-    it('should takeLoan', async () => {
+    it('should takeLoan with multiple amounts', async () => {
```
canonbrother, line 219: cover amounts
https://github.com/DeFiCh/jellyfish/pull/661#discussion_r716370270

packages/jellyfish-api-core/__tests__/category/loan/loanPayback.test.ts
```diff
@@ -0,0 +1,459 @@
...
+let bobVaultAddr: string
+let bobVaultId1: string
+let bobVaultAddr1: string
+let bobLiqVaultId: string
+let bobloanAddr: string
+let bobColAddr: string
+
+async function setup (): Promise<void> {
+  // token setup
+  const aliceColAddr = await alice.container.getNewAddress()
```
jingyi2811, line 20: You can keep   

let bobColAddr: string
let bobVaultAddr1: string

inside the setup function

packages/jellyfish-api-core/__tests__/category/loan/loanPayback.test.ts
```diff
@@ -0,0 +1,459 @@
...
+    expect(vaultAfter.loanValue).toStrictEqual(54.00006042) // 27.00003021 * 2 (::1 TSLA = 2 DFI)
+    expect(vaultAfter.currentRatio).toStrictEqual(27778) // 15000 / 54.00006042 * 100
+
+    const burnInfoAfter = await bob.container.call('getburninfo')
+    expect(burnInfoAfter.paybackburn).toStrictEqual(0.00000297)
+  })
+})
+
+describe('loanPayback multiple amounts', () => {
+  it('should loanPayback with multiple amounts', async () => {
```
jingyi2811, line 295:
```suggestion
  it('should loanPayback if takeLoan more than 1 times ', async () => {
```

jingyi2811: Or other text that you think is better

jingyi2811:
Add 2 more tests

- If calling loanPayBack without deposit any collateral tokens, you should get `Vault with id %s has no collaterals`
- Try calling loanPayBack two times, and for every time, pay back only half on the loan amount

canonbrother:
>     * Try calling loanPayBack two times, and for every time, pay back only half of the loan amount (Will the interest rate need to be paid reduced after calling loanPayBack for the first time?)

the interest has not included in loan yet (ain will fix it)

canonbrother:
>     * Try calling loanPayBack two times, and for every time, pay back only half of the loan amount (Will the interest rate need to be paid reduced after calling loanPayBack for the first time?)

1. interest won't be accrued within an auction progress
2. interest charged is not included yet after liquidated <-- issue, ain will fix

canonbrother:
> Also can add a test like this to `takeloan` RPC tests for completion.
> 
> https://github.com/DeFiCh/jellyfish/blob/6e4ece7bc73572bd83e6d81bf78d709e3fbf6013/packages/jellyfish-transaction-builder/__tests__/txn/txn_builder_loan_take_loan.test.ts#L262

hmm.. i don't quite understand this request...

PR #713, opened 2021-10-06
move GenesisKey in `testcontainers` to `@defichain/jellyfish-network`

fuxingloh:

/kind refactor

Move RegTestGenesisKey into `@defichain/jellyfish-network` so that it can be used without importing the entire testcontainers tooling.


fuxingloh:
> I felt it is actually more appropriate to reside in test/jellyfish-test 😅

I thought about but, jellyfish-testing imports testcontainers eventually 😅. In env such as the playground module where we are not using testcontainers those deps are unnecessary.

Issue #662, opened 2021-09-14
flaky anchor tests


fuxingloh:
#### What happened:

A few tests in anchor seem to be flaky, a showstopper for CI automation. We need to fix them.

/area jellyfish-api-core
/triage accepted
/priority urgent-now

PR #740, opened 2021-10-15
disable 3 flaky anchor test

fuxingloh:

/kind chore

Related to #662 and lowering the priority to focus on other work.

Due to the flaky nature of the anchor test, we have disabled it now so that it does not impact our CI workflow.
These tests are not deemed a high priority, they are random and sporadic by design, enough to allow failure of tests.


canonbrother:
> Yup, sounds good, we can fix them up and add those in at a later date. Just temporarily muting them for now, due to the re-run fatigue

yea understand
lemme push another changes to skip those using `waitForAnchoAuths`

PR #787, opened 2021-10-30
Add listAuctionHistory rpc

jingyi2811:

/kind feature

Add `listauctionhistory` rpc

Fixes part of #488


canonbrother (dismissed):

packages/jellyfish-api-core/__tests__/category/loan/listAuctionHistory.test.ts
```diff
@@ -0,0 +1,592 @@
...
+        const data3 = data.find(d => d.vaultId === vaultId3)
+        expect(data3).toBeDefined()
+
+        const data4 = data.find(d => d.vaultId === vaultId4)
+        expect(data4).toBeDefined()
+      }
+    })
+  })
+
+  describe('listAuctionHistory with pagination', () => { // For pagination, always filter by 'all'.
```
surangap, line 484: `For pagination, always filter by 'all'.` that is actually not true. you can paginate the results which are filtered by `mine ` or `owner`. and it's real world use case that we filter by `mine ` or `owner` and get the results paginated.
Can add a test for that too.
note that `owner` is filtration and all fields in `ListAuctionHistoryPagination` are for pagination.
Here also database index has `height` as first. So it's crucial that we pass the intended `height` that the pagination should start.

from cpp side
```
inline void SerializationOp(Stream& s, Operation ser_action) {

        if (ser_action.ForRead()) {
            READWRITE(WrapBigEndian(blockHeight));
            blockHeight = ~blockHeight;
        } else {
            uint32_t blockHeight_ = ~blockHeight;
            READWRITE(WrapBigEndian(blockHeight_));
        }

        READWRITE(owner);
        READWRITE(vaultId);
        READWRITE(index);
    }
```

jingyi2811: Added the test units for 
- 'mine' and address.
- height + vaultId
- height + index

surangap: Added some minor refactor to pagination tests.

surangap (approved):

canonbrother (approved):

fuxingloh (approved):

packages/jellyfish-api-core/src/category/loan.ts
```diff
@@ -396,6 +396,24 @@ export class Loan {
...
+   *
+   * @param {string} [owner] address or reserved word : mine / all (Default to mine)
+   * @param {ListAuctionHistoryPagination} pagination
+   * @param {number} [pagination.maxBlockHeight] Maximum block height
+   * @param {string} [pagination.vaultId] Vault Id
+   * @param {number} [pagination.index] Auction index
+   * @param {number} [pagination.limit = 100]
+   * @return {Promise<ListAuctionHistoryDetail>}
+   */
+  async listAuctionHistory (owner: string = 'mine', pagination?: ListAuctionHistoryPagination): Promise<ListAuctionHistoryDetail[]> {
```
ivan-zynesis, line 411: hmm, I know the vault id here is solely sorting index. But feel like list auction history as per vault is more useful for UI like `GET /auctions/:vaultId`

Issue #532, opened 2021-08-01
`DeFiCh/ain:2.x.x` Loan DfTx


fuxingloh:
> Similar to https://github.com/DeFiCh/jellyfish/issues/108, this issue tracks the implementation of custom DfTx in `jellyfish-transaction` and `jellyfish-transaction-builder`. Must implement in `jellyfish-transaction` and test in `jellyfish-transaction-builder` for each PR.

### Loan DfTX

- [x] createLoanScheme: `'L'` @jingyi2811 #560
- [x] updateLoanScheme: `'L'` @jingyi2811 #591
- [x] setDefaultLoanScheme: `'d'` @jingyi2811 #592
- [x] destroyLoanScheme: `'D'` @jingyi2811 #594
- [x] setCollateralToken: `'c'` @jingyi2811 #619
- [x] setLoanToken: `'g'` @jingyi2811 #587 
- [x] updateLoanToken: `'f'` @jingyi2811 #588
- [x] createVault: `'V'` @surangap  #582
- [x] updateVault: `'v'` @jingyi2811 #701 
- [x] depositToVault `'S'` @surangap https://github.com/DeFiCh/jellyfish/pull/705
- [x] withdrawfromvault `'J'` @ivan-zynesis #750  
- [x] auctionBid `'I'` @canonbrother #722 
- [x] takeLoan `'X'` @surangap https://github.com/DeFiCh/jellyfish/pull/719
- [x] loanPayBack `'H'` @canonbrother #782 
- [x] closeVault `'e'` @jingyi2811 #731
- [x] setGovVariableHeight @ivan-zynesis #790
- [x] compositeswap @ivan-zynesis #779 
---
We have composite swap and perhaps other smaller DfTx in FC, we need to check them.

PR #790, opened 2021-11-01
DfTx: `setGovHeight`

ivan-zynesis:

/kind feature

New DfTx type in FC

Fixes part of #532 

#### Additional comments?:
Not txn-builder is added as it is only for foundation (same for `setGov`), solely for decoding purpose.

### DIFF
--- a/packages/jellyfish-transaction-builder/src/txn/txn_builder_governance.ts
+++ b/packages/jellyfish-transaction-builder/src/txn/txn_builder_governance.ts
@@ -1,4 +1,4 @@
-import { CreateCfp, CreateVoc, OP_CODES, Script, TransactionSegWit, Vote } from '@defichain/jellyfish-transaction'
+import { CreateCfp, CreateVoc, OP_CODES, Script, TransactionSegWit, Vote, SetGovernance, SetGovernanceHeight } from '@defichain/jellyfish-transaction'
 import { P2WPKHTxnBuilder } from './txn_builder'
 import { TxnBuilderError, TxnBuilderErrorType } from './txn_builder_error'
 import BigNumber from 'bignumber.js'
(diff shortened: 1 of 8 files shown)


surangap (commented):
1. need to add transaction builder tests?
2. Code for `CGovernanceVar` composers is not that elegant🤔. But I guess we can not do much about it without changing the way how cpp side handles gov vars.

ivan-zynesis:
> 1. need to add transaction builder tests?
> 2. Code for `CGovernanceVar` composers is not that elegant🤔. But I guess we can not do much about it without changing the way how cpp side handles gov vars.

1. No transaction-builder is added at all thus no test. Consider the `CSetGovernanceVar` wasn't there either. The jellyfish-txn is mainly for scan or other client to decode/read existing dftx history.
2. Yes, cannot really make changes, there is no GovVar length. Also quite impossible to change for sake of backward compatibility.

canonbrother (dismissed):

surangap:
> > 1. need to add transaction builder tests?
> > 2. Code for `CGovernanceVar` composers is not that elegant🤔. But I guess we can not do much about it without changing the way how cpp side handles gov vars.
> 
> 1. No transaction-builder is added at all thus no test. Consider the `CSetGovernanceVar` wasn't there either. The jellyfish-txn is mainly for scan or other client to decode/read existing dftx history.
> 2. Yes, cannot really make changes, there is no GovVar length. Also quite impossible to change for sake of backward compatibility.

so for number 1, your point is that since only the foundation sets gov vars, jellyfish does not need it?

ivan-zynesis:
> > > 1. need to add transaction builder tests?
> > > 2. Code for `CGovernanceVar` composers is not that elegant🤔. But I guess we can not do much about it without changing the way how cpp side handles gov vars.
> > 
> > 
> > 
> > 1. No transaction-builder is added at all thus no test. Consider the `CSetGovernanceVar` wasn't there either. The jellyfish-txn is mainly for scan or other client to decode/read existing dftx history.
> > 2. Yes, cannot really make changes, there is no GovVar length. Also quite impossible to change for sake of backward compatibility.
> 
> so for number 1, your point is that since only the foundation sets gov vars, jellyfish does not need it for setting?

The implementation in this PR at the moment cover only `jellyfish-transaction` but not `jellyfish-transaction-builder` thus no test for txn-builder

We can/should add the builder if there is need to perform `setGov` using jellyfish library together with foundation auth. If that's the case, we will have to add `setGov` dftx builder as well. Is the need foreseen? What do you think? @surangap @fuxingloh

fuxingloh:
I am in the middle, with builder test at least any basic test, we can validate the authenticity and safeguard ourself from regression. It's also the only way to know it's correct and can be submitted to chain now.

Many time it was useful in migration where upstream altered the structure without informing.

I guess it's also forward looking as we want to eventually achieve full node compatibility using N-API?

I guess we can meet in the middle now with extremely basic positive test to submit?

ivan-zynesis:
Okay will add the txn builder tomorrow (should very quick one, literally a function calling super.createDfTx) and minimal test.

packages/jellyfish-transaction-builder/src/txn/txn_builder_governance.ts
```diff
@@ -59,4 +59,32 @@ export class TxnBuilderGovernance extends P2WPKHTxnBuilder {
...
+   */
+  async setGoverance (setGov: SetGovernance, changeScript: Script): Promise<TransactionSegWit> {
+    return await this.createDeFiTx(
+      OP_CODES.OP_DEFI_TX_SET_GOVERNANCE(setGov),
+      changeScript
+    )
+  }
+
+  /**
+   * Set governance variable.
```
surangap, line 78:
```suggestion
   * Set governance variable at the given height
```

packages/jellyfish-transaction-builder/src/txn/txn_builder_governance.ts
```diff
@@ -59,4 +59,32 @@ export class TxnBuilderGovernance extends P2WPKHTxnBuilder {
...
+   *
+   * @param {SetGovernance} setGov txn to create
```
surangap, line 80:
```suggestion
      * @param {SetGovernanceHeights} setGovHeight txn to create
```

surangap (commented):
looks good. minor comments. Also please add the non foundation failure test.

surangap (approved):

jingyi2811 (approved):

fuxingloh (approved):

PR #791, opened 2021-11-01
Disable docker pruning

ivan-zynesis:

/kind fix

multiple test cases prune container clashing each other
Base on multiple occurrences in our test case, hypothesis made: Dockerode prune API is singleton, one execution at one time regardless the option is specifying particular container.

Fixes #

### DIFF
--- a/packages/testcontainers/src/containers/RegTestContainer/ContainerGroup.ts
+++ b/packages/testcontainers/src/containers/RegTestContainer/ContainerGroup.ts
@@ -104,14 +104,14 @@ export class ContainerGroup {
    * Stop container group and all containers associated with it
    */
   async stop (): Promise<void> {
-    const containerIds = this.containers.map(c => c.id)
     for (const container of this.containers) {
       await this.requireNetwork().disconnect({ Container: container.id })
       await container.stop()
     }
     await this.requireNetwork().remove()
-    for (const id of containerIds) {
-      await this.docker.pruneContainers({ Container: id })
-    }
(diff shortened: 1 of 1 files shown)


packages/testcontainers/src/containers/RegTestContainer/ContainerGroup.ts
```diff
@@ -104,11 +104,14 @@ export class ContainerGroup {
...
   async stop (): Promise<void> {
+    const containerIds = this.containers.map(c => c.id)
     for (const container of this.containers) {
       await this.requireNetwork().disconnect({ Container: container.id })
       await container.stop()
     }
     await this.requireNetwork().remove()
-    await this.docker.pruneContainers()
+    for (const id of containerIds) {
+      await this.docker.pruneContainers({ Container: id })
```
surangap, line 114: Do you think passing `{ Container: id }` to pruneContainers is effective? 
https://docs.docker.com/engine/reference/commandline/container_prune/

ivan-zynesis: Arghh, this is a serious problem now, our tests always run 30+ minutes and fail. Let's just drop the `prune`, we just put it locally when having docker issue.

ivan-zynesis: But that's really weird, it could be the `prune` API of dockerode (not docker) itself has to be singleton instead of by ID. I don't think the random 1 of 10 millions random name can clash.

fuxingloh (approved):

jingyi2811 (approved):

PR #789, opened 2021-11-01
RPC: `setGovHeight` and `listGovs`

ivan-zynesis:
#### What kind of PR is this?:

/kind feature

#### What this PR does / why we need it:
New RPC in FC

#### Which issue(s) does this PR fixes?:

Fixes #

#### Additional comments?:
Doing both RPC together, only `setGovHeight` will cause significant change in `listGovs` result

PR #753, opened 2021-10-19
`2.0.0` - FortCanning Upgrade

fuxingloh:
#### What kind of PR is this?:

/kind feature

#### Which issue(s) does this PR fixes?:

Fixes #761

PR #794, opened 2021-11-02
RPC update: `getGov`

ivan-zynesis:

/kind feature

Data structure change in RPC response

Fixes #

#### Additional comments?:
Part of #532 
Changes include #789 (require new setGov rpc)

### DIFF
--- a/packages/jellyfish-api-core/src/category/masternode.ts
+++ b/packages/jellyfish-api-core/src/category/masternode.ts
@@ -141,6 +141,22 @@ export class Masternode {
   }
 
   /**
+   * Set special governance variables with activation height specified
+   *
+   * @param {Record<string, any>} input json object
+   * @param {number} activationHeight
+   * @return {Promise<string>} hash
+   *
+   */
+  async setGovWithHeight (input: Record<string, any>, activationHeight: number): Promise<string> {
+    const currentHeight: number = await this.client.call('getblockcount', [], 'number')
+    if (activationHeight < currentHeight) {
+      throw new Error('GovVar activation height must be higher than current block height')
+    }
+    return await this.client.call('setgovheight', [input, activationHeight], 'number')
+  }
+
+  /**
    * Get information about governance variable
    *
    * @param {string} name governance name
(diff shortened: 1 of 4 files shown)


canonbrother (approved):

packages/jellyfish-api-core/src/category/masternode.ts
```diff
@@ -141,15 +141,49 @@ export class Masternode {
...
+   * @param {string} name governance name
+   * @return {Promise<Array<Record<string, any>>} governance information as json object
+   * @example
+   * {
+   *   "LP_SPLITS": { "3": BigNumber, "4": BigNumber },
+   *   "5000": { "3": BigNumber, "4": BigNumber }, // value after block height 5000
+   *   "8000": { "3": BigNumber, "4": BigNumber } // value after block height 8000
+   * }
+   */
+  async getGovWithHeight (name: string): Promise<Array<Record<string, any>>> {
```
fuxingloh, line 183: Due to regression in https://github.com/DeFiCh/jellyfish/pull/753 we can't deprecate getGov, we have to replace it and fix all the tests that broke due to it. This change has to be moved to #753 due to various regression

ivan-zynesis: Porting changes by hand, no more multiple versions.

surangap: was checking why we need `Array<Record<string, any>>` in the return object. This is the structure of the return data
```
 [
      { LP_SPLITS: { '3': [BigNumber], '4': [BigNumber] } },
      { '109': { '3': [BigNumber], '4': [BigNumber] } },
      { '114': { '3': [BigNumber], '4': [BigNumber] } }
    ]
```

seems bit odd to me. Why can't  it return `Record<string, any>` type. i.e
```
{
      LP_SPLITS: { '3': [BigNumber], '4': [BigNumber] },
      '109': { '3': [BigNumber], '4': [BigNumber] } ,
      '114': { '3': [BigNumber], '4': [BigNumber] }
}
```
does it need the extra Array?

surangap: btw we also have a `listgovs` RPC

ivan-zynesis:
Defid reverted

PR #731, opened 2021-10-12
Add closeVault dftx

jingyi2811:
#### What kind of PR is this?:

/kind feature

#### What this PR does / why we need it:

Add closeVault dftx

#### Which issue(s) does this PR fixes?:

Fixes part of #532

PR #816, opened 2021-11-09
Fix flaky test and update timeout / type for waitfor functions

jingyi2811:

/kind fix

1. Fix flanky test. (See https://github.com/DeFiCh/jellyfish/runs/4142447619?check_suite_focus=true#step:5:3782)

2. Change timeout from 100,000 to 30,000 ms for WaitFor functions.

3. activePrice / nextPrice should be parsed as string. (See https://github.com/DeFiCh/jellyfish/pull/731#discussion_r744466882)

Fixes #

### DIFF
--- a/packages/testcontainers/src/containers/RegTestContainer/Masternode.ts
+++ b/packages/testcontainers/src/containers/RegTestContainer/Masternode.ts
@@ -199,9 +199,8 @@ export class MasterNodeRegTestContainer extends RegTestContainer {
       if (!data.isLive) {
         await this.generate(1)
         return false
-      } else {
-        return true
       }
+      return true
     }, timeout, 100, 'waitForPriceValid')
   }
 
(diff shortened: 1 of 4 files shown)


packages/testcontainers/src/containers/RegTestContainer/Masternode.ts
```diff
@@ -229,15 +229,15 @@ export class MasterNodeRegTestContainer extends RegTestContainer {
...
-  async waitForActivePrice (fixedIntervalPriceId: string, activePrice: number, timeout = 100000): Promise<void> {
+  async waitForActivePrice (fixedIntervalPriceId: string, activePrice: string, timeout = 30000): Promise<void> {
     return await waitForCondition(async () => {
       const data: any = await this.call('getfixedintervalprice', [fixedIntervalPriceId])
       // eslint-disable-next-line
-      if (data.activePrice !== activePrice) {
+      if (data.activePrice.toString() !== activePrice) {
         await this.generate(1)
         return false
       } else {
```
canonbrother, line 243: style this way
```
if (....) {
  ....
  return false
}
return true

jingyi2811: Ya your way is recommended by one of the linter. I forget the name

surangap (approved):

fuxingloh (approved):

PR #819, opened 2021-11-09
Fix DUSD related fixture

canonbrother:

/kind fix

1. DUSD fixture correction
old: set DUSD token as normal token and minted
new: set DUSD token as loan token and should own DUSD from `takeLoan`

2. test paybackLoan with DUSD
3. test zero interest

### DIFF
--- a/packages/jellyfish-api-core/__tests__/category/loan/closeVault.test.ts
+++ b/packages/jellyfish-api-core/__tests__/category/loan/closeVault.test.ts
@@ -32,29 +32,20 @@ describe('Loan', () => {
   async function setup (): Promise<void> {
     // token setup
     const collateralAddress = await tGroup.get(0).container.getNewAddress()
-    await tGroup.get(0).token.dfi({ address: collateralAddress, amount: 40000 })
-    await tGroup.get(0).token.create({ symbol: 'DUSD', collateralAddress })
-    await tGroup.get(0).generate(1)
-    await tGroup.get(0).token.mint({ symbol: 'DUSD', amount: 600000 })
-    await tGroup.get(0).generate(1)
-    await tGroup.get(0).poolpair.create({ tokenA: 'DUSD', tokenB: 'DFI' })
-    await tGroup.get(0).generate(1)
-    await tGroup.get(0).poolpair.add({
-      a: { symbol: 'DUSD', amount: 25000 },
-      b: { symbol: 'DFI', amount: 10000 }
-    })
-    await tGroup.get(0).generate(1)
+    await tGroup.get(0).token.dfi({ address: collateralAddress, amount: 200000 })
 
     // oracle setup
     const addr = await tGroup.get(0).generateAddress()
     const priceFeeds = [
       { token: 'DFI', currency: 'USD' },
-      { token: 'TSLA', currency: 'USD' }
+      { token: 'TSLA', currency: 'USD' },
+      { token: 'DUSD', currency: 'USD' }
     ]
     oracleId = await tGroup.get(0).rpc.oracle.appointOracle(addr, priceFeeds, { weightage: 1 })
     await tGroup.get(0).generate(1)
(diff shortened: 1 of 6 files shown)


surangap (dismissed):

jingyi2811 (dismissed):

packages/jellyfish-api-core/__tests__/category/loan/closeVault.test.ts
```diff
@@ -32,29 +32,20 @@ describe('Loan', () => {
   async function setup (): Promise<void> {
     // token setup
     const collateralAddress = await tGroup.get(0).container.getNewAddress()
-    await tGroup.get(0).token.dfi({ address: collateralAddress, amount: 40000 })
```
jingyi2811, line 35: A question here: Why do you change this file? Because the test failed right with the latest ain code ?

canonbrother: nope.. i want make use of `collateralAddress` for something else

packages/jellyfish-api-core/__tests__/category/loan/paybackLoan.test.ts
```diff
@@ -548,6 +562,56 @@ describe('paybackLoan success', () => {
...
+
+    const vaultAfter = await bob.container.call('getvault', [bobVaultId])
+    expect(vaultAfter).toStrictEqual({
+      vaultId: bobVaultId,
+      loanSchemeId: 'scheme',
+      ownerAddress: bobVaultAddr,
+      state: 'active',
+      collateralAmounts: ['10000.00000000@DFI', '1.00000000@BTC'],
+      loanAmounts: ['0.00001084@DUSD', '40.00006849@TSLA'],
+      interestAmounts: ['0.00000000@DUSD', '0.00006849@TSLA'],
```
ivan-zynesis, line 607: Just curious, 0 DUSD interest is because the interest of 0.00001084 is too small? Shouldn't here will continue accrue dust among our interest?

ivan-zynesis (dismissed):

packages/jellyfish-api-core/__tests__/category/loan/paybackLoan.test.ts
canonbrother: nope.. the amount paid the interest first

ivan-zynesis: I remember previously the behavior of `interestAmount` is a projected amount if no more payment in current block? Is that fixed? I understand the previous interest are paid, but current remaining 0.00001084 will continue incur interest right

canonbrother: > I remember previously the behavior of interestAmount is a projected amount if no more payment in current block? Is that fixed?
yep.. the interest block height re-allocated calculation is fixed.. no projected amount.

> I understand the previous interest are paid, but current remaining 0.00001084 will continue incur interest right
yup the remaining `0.00001084` should incur the interest.. its worth to test on this dust amount

canonbrother: > yup the remaining 0.00001084 should incur the interest.. its worth to test on this dust amount
@ivan-zynesis 
bug :lady_beetle:  :sweat:

jingyi2811 (approved):

surangap (approved):
What about the other loan tokens that we mint in regtest🤔. IMO don't see a problem as long as if it does not interfere with the test scope.

fuxingloh (approved):

fuxingloh:
> What about the other loan tokens that we mint in regtest🤔. IMO don't see a problem as long as if it does not interfere with the test scope.

Yeah definitely something we need to refactor.

PR #820, opened 2021-11-10
This is a loan playground PR to test auctionbid

jingyi2811:

/kind

Fixes #


packages/jellyfish-api-core/__tests__/category/loan/placeAuctionBid.test.ts
```diff
@@ -0,0 +1,643 @@
...
+    await alice.generate(10)
+    const auctionsAfter1 = await bob.container.call('listauctions')
+    expect(auctionsAfter).toStrictEqual(auctionsAfter1)
+
+    await bob.generate(40)
+
+    // should not allow place bid
+    const txid1 = await alice.rpc.loan.placeAuctionBid({
+      vaultId: bobVaultId,
+      index: 0,
```
jingyi2811, line 316: Okay, you are right. It can be bid again because the vault has two indexes
Once index 0 is won by someone, index = 1 will become index = 0. And this is why, the second bidding on index 0 is still allowed here.

jingyi2811: I think we probably want to inform the C++ side that we should not allow element INDEX to be reorganised. Imagine someone is bidding for index 1 and suddenly he is required to bid for 0. A bit confusing here.

jingyi2811:
For your information, the reason why `listhistoryauction` returns empty object is because it needs to wait until ALL INDEXED successfully bid, then only the records will appear there, which I think also not very good.

packages/jellyfish-api-core/__tests__/category/loan/placeAuctionBid.test.ts
canonbrother: that's the design... i think better to hold this discussion internally

jingyi2811: You see, assume that wallet is pointing to index 3 and let the user bid. 
Suddenly 2 indexes won and index 3 become 1. Wallet does not know what is the new number for 3

PR #912, opened 2021-12-23
feat(packages/playground): playground testing blockchain for simulated env

fuxingloh:

> As part of #580 consolidation efforts. We had multiple projects that were extensions of the jellyfish project. The separated projects allowed us to move quickly initially but proves to be a bottleneck when it comes to development.

Consolidation and redesign of https://github.com/DeFiCh/playground for more seamless integration with `apps/ocean-api`.

## `@defichain/playground`

> This package is not published, for internal use within `@defichain-apps/ocean-api` only.

`@defichain/playground` is a specialized testing blockchain isolated from MainNet for testing DeFi applications. Assets are not real, they can be minted by anyone. Blocks are configured to generate every 3 seconds, the chain can reset anytime.

A bot-like design centers the playground as a mechanism that allows bootstrapping with an interval cycle. This allows the developer to mock any behaviors they want with a simulated testing blockchain.

### Playground Design

Playground follows the chain of responsibility pattern (think Filter in ExpressJS). Each bot has its own concern that it manages; it can be generating block, publishing oracle data, or setting up foundation auth. Each bot enforced with `AbstractBot` abstraction comes with `bot.bootstrap()` and `bot.cycle()` method.

```typescript
class AbstractBot {
  constructor (
    protected readonly apiClient: ApiClient,
    protected readonly logger: BotLogger
  ) {
  }

  bootstrap (): Promise<void> {
  }

  cycle (nextBlockCount: number): Promise<void> {
  }
}
```

#### Initialize

```typescript
const logger = {
  info: (action: string, message: string) => console.log(`...`)
}
const rpc = new JsonRpcClient()
const playground = new Playground(rpc, logger)
```

#### Boostrap

```typescript
await playground.bootstrap()
```

#### Run each cycle and your defined interval

```typescript
await playground.cycle()
```


packages/playground/src/bots/BlockGenerateBot.ts
```diff
@@ -0,0 +1,20 @@
+import { AbstractBot } from '../AbstractBot'
+import { FoundationBot } from './FoundationBot'
+
+/**
+ * Generate a block every cycle
+ */
+export class BlockGenerateBot extends AbstractBot {
+  private static randomNodeAddress (): string {
```
ivan-zynesis, line 8: Initially thinking if move this static method to FoundationBot.

Changed my mind, I think the purpose here is to allow the random address pool to include extra injected non foundation masternodes?

fuxingloh: The main purpose is to allow a pool of addresses so we can split up the UTXO. Ideally when we have a full node jellyfish wallet, I would like to move the responsibility over to that instead of defid.

chee-chyuan (approved):
still pretty much clueless on the general direction. i guess im not seeing the big picture yet. but otherwise lgtm

canonbrother (approved):

PR #913, opened 2021-12-23
feat(testcontainers): add mockTime for waitForWalletCoinbaseMaturity to generate block faster

fuxingloh:

`@defichain/testcontainers` `waitForWalletCoinbaseMaturity` is kinda slow and it makes us wait 30-40 seconds between each test cycle. Pretty bad developer experience.

This PR introduces `mockTime=true` to simulate post genesis time at 5x speed to speed up the initial block minting process to get past coinbase limit quickly. It will restore to default system time after coinbase.

---

Alternatively, we can submit 100 pre-defined deterministic blocks. However, this design is less static and more dynamic allowing for greater testing evolution.

### DIFF
--- a/packages/testcontainers/src/containers/RegTestContainer/Masternode.ts
+++ b/packages/testcontainers/src/containers/RegTestContainer/Masternode.ts
@@ -80,7 +80,7 @@ export class MasterNodeRegTestContainer extends RegTestContainer {
    * Wait for block height by minting towards the target
    *
    * @param {number} height to wait for
-   * @param {number} [timeout=90000] in ms
+   * @param {number} [timeout=590000] in ms
    */
   async waitForBlockHeight (height: number, timeout = 590000): Promise<void> {
     return await waitForCondition(async () => {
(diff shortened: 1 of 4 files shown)


packages/testcontainers/src/containers/RegTestContainer/Masternode.ts
```diff
@@ -101,9 +101,24 @@ export class MasterNodeRegTestContainer extends RegTestContainer {
...
+  async waitForWalletCoinbaseMaturity (timeout: number = 180000, mockTime: boolean = true): Promise<void> {
+    if (!mockTime) {
+      return await this.waitForBlockHeight(100, timeout)
+    }
+
+    let fakeTime: number = 1579045065
+    await this.call('setmocktime', [fakeTime])
+
+    const intervalId = setInterval(() => {
+      void this.call('setmocktime', [fakeTime++])
```
ivan-zynesis, line 115: regtest is 3 secs per block? Maybe we can forward 3 sec straight (per 200ms)?

ivan-zynesis (approved):

canonbrother (approved):

PR #928, opened 2021-12-31
feat(ocean-api): integrate playground with ocean-api

fuxingloh:

Integrated playground to automatically run when `PLAYGROUND_ENABLE` is provided on `regtest`.

### DIFF
--- a/apps/ocean-api/__tests__/modules/PlaygroundModule.test.ts
+++ b/apps/ocean-api/__tests__/modules/PlaygroundModule.test.ts
@@ -13,7 +13,7 @@ describe('regtest', () => {
     await apiTesting.stop()
   })
 
-  it('block count should not increment', async () => {
+  it('should not increment block count', async () => {
     const initial = await apiTesting.rpc.blockchain.getBlockCount()
 
     await new Promise((resolve) => {
@@ -37,7 +37,7 @@ describe('playground', () => {
     await apiTesting.stop()
   })
 
-  it('block count should increment', async () => {
+  it('should increment block count', async () => {
     const initial = await apiTesting.rpc.blockchain.getBlockCount()
 
     await waitForExpect(async () => {
--- a/apps/ocean-api/__tests__/modules/PlaygroundModule.test.ts
+++ b/apps/ocean-api/__tests__/modules/PlaygroundModule.test.ts
@@ -1,0 +1,48 @@
+import { OceanApiTesting } from '../../testing/OceanApiTesting'
+import waitForExpect from 'wait-for-expect'
+
+describe('regtest', () => {
+  const apiTesting = OceanApiTesting.create()
+
+  beforeEach(async () => {
+    apiTesting.playgroundEnable(false)
+    await apiTesting.start()
+  })
+
+  afterEach(async () => {
+    await apiTesting.stop()
+  })
+
+  it('block count should not increment', async () => {
+    const initial = await apiTesting.rpc.blockchain.getBlockCount()
+
+    await new Promise((resolve) => {
+      setTimeout(_ => resolve(0), 5000)
+    })
+
+    const next = await apiTesting.rpc.blockchain.getBlockCount()
+    expect(next).toStrictEqual(initial)
+  })
+})
+
+describe('playground', () => {
+  const apiTesting = OceanApiTesting.create()
+
+  beforeEach(async () => {
+    apiTesting.playgroundEnable(true)
+    await apiTesting.start()
+  })
+
+  afterEach(async () => {
+    await apiTesting.stop()
+  })
+
+  it('block count should increment', async () => {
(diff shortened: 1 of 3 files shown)


canonbrother (approved):

monstrobishi (dismissed):

apps/ocean-api/__tests__/modules/PlaygroundModule.test.ts
```diff
@@ -0,0 +1,48 @@
...
+  beforeEach(async () => {
+    apiTesting.playgroundEnable(false)
+    await apiTesting.start()
+  })
+
+  afterEach(async () => {
+    await apiTesting.stop()
+  })
+
+  it('block count should not increment', async () => {
```
kodemon, line 16: `nit:` do we have any conventions for `it` method name argument in our tests? In this case we could consider
```
it('should not increment block count')
```
being potentially more correct sentence structure here.

surangap (approved):

apps/ocean-api/__tests__/modules/PlaygroundModule.test.ts
```diff
@@ -0,0 +1,48 @@
...
+  beforeEach(async () => {
+    apiTesting.playgroundEnable(true)
+    await apiTesting.start()
+  })
+
+  afterEach(async () => {
+    await apiTesting.stop()
+  })
+
+  it('block count should increment', async () => {
```
kodemon, line 40: `nit:` Following up on the previous statement
```
it('should increment block count')
```

apps/ocean-api/__tests__/modules/PlaygroundModule.test.ts
```diff
@@ -0,0 +1,48 @@
...
+  afterEach(async () => {
+    await apiTesting.stop()
+  })
+
+  it('block count should not increment', async () => {
+    const initial = await apiTesting.rpc.blockchain.getBlockCount()
+
+    await new Promise((resolve) => {
+      setTimeout(_ => resolve(0), 5000)
+    })
```
kodemon, line 21: `question:` Could a test utility box with a method be valid here for similar use cases in the future?
```ts
async function waitForTimeout(ms: number) {
  await new Promise((resolve) => {
    setTimeout(_ => resolve(0), ms)
  })
}
```
I am still working on catching up on project conventions so posing this as a question for further elaboration and knowledge gains.

chee-chyuan (approved):

kodemon (approved):
Just had a few questions to gain some more insights into project code conventions.

apps/ocean-api/__tests__/modules/PlaygroundModule.test.ts
fuxingloh:
```suggestion
  it('should not increment block count', async () => {
```

Thanks! You can also add suggestions for me to batch commit in

apps/ocean-api/__tests__/modules/PlaygroundModule.test.ts
fuxingloh:
```suggestion
  it('should increment block count', async () => {
```

apps/ocean-api/__tests__/modules/PlaygroundModule.test.ts
fuxingloh: Hmm, I am a little skeptical doing this now since it's one such pattern. But definetly something to revisit later.

ivan-zynesis: Preferably (my pref) to stick with more native syntax in the test itself. Especially for simple stuff, to limit the "black box" from growing.

kodemon: Thanks for the feedback, will mentally log it and move on! 😄

kodemon (approved):

apps/ocean-api/src/modules/PlaygroundModule.ts
```diff
@@ -0,0 +1,65 @@
...
+  ],
+  providers: [
+    {
+      provide: 'PLAYGROUND_RUNNER',
+      useFactory: (configService: ConfigService, apiClient: ApiClient): PlaygroundRunner | undefined => {
+        if (configService.get<boolean>('PLAYGROUND_ENABLE') === false) {
+          return undefined
+        }
+
+        if (configService.get<string>('API_NETWORK') !== 'regtest') {
```
ivan-zynesis, line 22: Not sure if any chance we can use this for testnet, like pre deployment test.

ivan-zynesis (approved):

apps/ocean-api/src/modules/PlaygroundModule.ts
fuxingloh: Probably not since we need the privateKeys

PR #938, opened 2022-01-04
feat(jellyfish-api-core): add network getPeerInfo RPC

kodemon:

/kind feature

Added getPeerInfo rpc as part of ongoing #48

### DIFF
--- a/packages/jellyfish-api-core/__tests__/category/net/getPeerInfo.test.ts
+++ b/packages/jellyfish-api-core/__tests__/category/net/getPeerInfo.test.ts
@@ -23,7 +23,7 @@ function createNetworkTests (name: string, container: RegTestContainer | MasterN
 
       expect(peers.length).toBeGreaterThan(0)
 
-      for (const peer of await client.net.getPeerInfo()) {
+      for (const peer of peers) {
         expect(typeof peer.id).toStrictEqual('number')
 
         expectValidAddress(peer.addr)
--- a/packages/jellyfish-api-core/__tests__/category/net/getPeerInfo.test.ts
+++ b/packages/jellyfish-api-core/__tests__/category/net/getPeerInfo.test.ts
@@ -1,21 +1,35 @@
 import { MasterNodeRegTestContainer, RegTestContainer } from '@defichain/testcontainers'
 import { PeerInfo } from 'packages/jellyfish-api-core/src/category/net'
 import { ContainerAdapterClient } from '../../container_adapter_client'
+import Dockerode from 'dockerode'
 
-createNetworkTests('Network without masternode', new RegTestContainer())
-createNetworkTests('Network on masternode', new MasterNodeRegTestContainer())
+createNetworkTests('Network without masternode', [new RegTestContainer(), new RegTestContainer()])
+createNetworkTests('Network on masternode', [new MasterNodeRegTestContainer(), new MasterNodeRegTestContainer()])
 
-function createNetworkTests (name: string, container: RegTestContainer | MasterNodeRegTestContainer): void {
+function createNetworkTests (name: string, container: RegTestContainer[] | MasterNodeRegTestContainer[]): void {
   describe(name, () => {
-    const client = new ContainerAdapterClient(container)
+    const client = new ContainerAdapterClient(container[0])
+    let createdNetwork: Dockerode.NetworkInspectInfo
 
     beforeAll(async () => {
-      await container.start()
-      await container.addNode(await container.getIp('bridge'))
+      await container[0].start()
+      await container[0].createNetwork('test')
+      const networks = await container[0].listNetworks()
+      createdNetwork = networks.find(n => n.Name === 'test') as Dockerode.NetworkInspectInfo
+      await container[0].connectNetwork(createdNetwork.Id)
+
+      await container[1].start()
+      await container[1].connectNetwork(createdNetwork.Id)
+
+      // addnode
+      await container[0].addNode(await container[1].getIp('test'))
     })
 
     afterAll(async () => {
-      await container.stop()
+      await container[0].stop()
+      await container[1].stop()
+
+      await container[0].removeNetwork(createdNetwork.Id)
     })
 
     it('should getPeerInfo', async () => {
--- a/packages/jellyfish-api-core/__tests__/category/net/getPeerInfo.test.ts
+++ b/packages/jellyfish-api-core/__tests__/category/net/getPeerInfo.test.ts
@@ -21,40 +21,44 @@ function createNetworkTests (name: string, container: RegTestContainer | MasterN
       await expect(client.net.getPeerInfo()).resolves.toHaveLength(2)
     })
 
-    it('should return expected schema', async () => {
+    it('should return valid schema', async () => {
       for (const peer of await client.net.getPeerInfo()) {
         expect(typeof peer.id).toStrictEqual('number')
-        expect(typeof peer.addr).toStrictEqual('string')
-        expect(typeof peer.addrbind).toStrictEqual('string')
+
+        expectValidAddress(peer.addr)
+
+        expectUndefinedOrValidAddress(peer.addrbind)
+        expectUndefinedOrValidAddress(peer.addrlocal)
+
         expect(typeof peer.services).toStrictEqual('string')
+        expect(peer.services).toHaveLength(16)
+
         expect(typeof peer.relaytxes).toStrictEqual('boolean')
         expect(typeof peer.lastsend).toStrictEqual('number')
         expect(typeof peer.lastrecv).toStrictEqual('number')
+        expect(typeof peer.bytessent).toStrictEqual('number')
+        expect(typeof peer.bytesrecv).toStrictEqual('number')
         expect(typeof peer.conntime).toStrictEqual('number')
         expect(typeof peer.timeoffset).toStrictEqual('number')
+
+        expectUndefinedOrEquals('number', peer.pingtime)
+        expectUndefinedOrEquals('number', peer.minping)
+        expectUndefinedOrEquals('number', peer.pingwait)
+
         expect(typeof peer.version).toStrictEqual('number')
         expect(typeof peer.subver).toStrictEqual('string')
         expect(typeof peer.inbound).toStrictEqual('boolean')
         expect(typeof peer.addnode).toStrictEqual('boolean')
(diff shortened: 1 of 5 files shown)


packages/jellyfish-api-core/src/category/net.ts
```diff
@@ -29,6 +38,38 @@ export class Net {
   }
 }
 
+export interface PeerInfo {
+  id: number
+  addr: string
+  addrbind: string
+  addrlocal: string
```
kodemon, line 45: I am wondering if these should be set as potentially undefined when looking at the [net.cpp](https://github.com/DeFiCh/ain/blob/master/src/rpc/net.cpp#L145-L148) implementation?

canonbrother, line 45:
```suggestion
  addrbind?: string
  addrlocal?: string
```
yea.. those contains condition.. just make it optional

canonbrother: below has few condition properties which are required to set optional as well :)

kodemon: @canonbrother Would this also be the case for the following value assignments?

[1](https://github.com/DeFiCh/ain/blob/master/src/rpc/net.cpp#L157-L162)
[2](https://github.com/DeFiCh/ain/blob/master/src/rpc/net.cpp#L171-L180)

canonbrother: yep

kodemon: Ok great, will update interface and tests accordingly

fuxingloh:
We also moved to "Semantic" style PR title instead of Chat Ops via `/kind feature`.

This means your PR title should be `feat(jellyfish-api-core): add network getPeerInfo RPC `

kodemon:
Currently experiencing some difficulty properly getting stable unit tests passing.

During testing I am failing to consistently being able to retrieve a list of peers from the test container.

While digging through the various test container I found that an approach @chee-chyuan used performs an `.addNode` surrounded by some network requirements. Which differs from the approach I am taking in this PR. I am not sure if I should be trying to use the group container although I do not need the group functionality, or if there is another way to ensure a consistent peerInfo list result.

Here is some links to logic on group container:

https://github.com/DeFiCh/jellyfish/blob/main/packages/testcontainers/src/containers/RegTestContainer/ContainerGroup.ts#L43-L46

Here its starting the containers and running the `.add` method which triggers:

https://github.com/DeFiCh/jellyfish/blob/main/packages/testcontainers/src/containers/RegTestContainer/ContainerGroup.ts#L71-L79

Here I am noting line 72 which performs a require network function and then performing an `.addNode`.

In this PR I am using the `.addNode` as noted here: https://github.com/DeFiCh/jellyfish/pull/938/files#diff-9a976903bc2708f896fec50b1072d524efe1f494c867727b367a61e029a5b051R14 but this does not come with the require network part. I am not sure if this is needed or not.

After digging through the various test containers I didn't see a clean way or reason for why this inconsistency occurs so will need an assist at troubleshooting this specific case.

packages/jellyfish-api-core/__tests__/category/net/getPeerInfo.test.ts
```diff
@@ -0,0 +1,92 @@
...
+        expectValidBytesPerMessage(peer.bytesrecv_per_msg)
+      }
+    })
+  })
+}
+
+function expectUndefinedOrEquals (expected: any, actual?: any): void {
+  if (actual !== undefined) {
+    expect(typeof actual).toStrictEqual(expected)
+  }
```
canonbrother, line 74: the assertion might not be triggered due to the "if"
so if writing this way, we need to use `expect.assections(num)` as well to ensure the "if" is triggered

canonbrother (approved):
general wise.. good for me

packages/jellyfish-api-core/__tests__/category/net/getPeerInfo.test.ts
```diff
@@ -0,0 +1,92 @@
...
+createNetworkTests('Network without masternode', new RegTestContainer())
+createNetworkTests('Network on masternode', new MasterNodeRegTestContainer())
+
+function createNetworkTests (name: string, container: RegTestContainer | MasterNodeRegTestContainer): void {
+  describe(name, () => {
+    const client = new ContainerAdapterClient(container)
+
+    beforeAll(async () => {
+      await container.start()
+      await container.addNode(await container.getIp('bridge'))
```
canonbrother, line 14: hmm.. here should be like this
await container.addNode(await container.getIp('testcontainer-{num}')

canonbrother (changes requested):

packages/jellyfish-api-core/__tests__/category/net/getPeerInfo.test.ts
kodemon: @canonbrother thanks for the feedback, I have some questions on this

```ts
await container.addNode(await container.getIp('testcontainer-{num}')
```

When I looked at [DockerContainer](https://github.com/DeFiCh/jellyfish/blob/main/packages/testcontainers/src/containers/DockerContainer.ts#L18-L22) it seems to require a name that exists in the networks list. When I printed out the network lists value in this test file the only name that exists is `bridge`.

I did see that in [ContainerGroup.add](https://github.com/DeFiCh/jellyfish/blob/main/packages/testcontainers/src/containers/RegTestContainer/ContainerGroup.ts#L75) its using name assigned when creating the docker network at [ContainerGroup.start](https://github.com/DeFiCh/jellyfish/blob/main/packages/testcontainers/src/containers/RegTestContainer/ContainerGroup.ts#L29)

So when executing `start` directly on [MasterNode](https://github.com/DeFiCh/jellyfish/blob/main/packages/testcontainers/src/containers/RegTestContainer/Masternode.ts) or [RegTestContainer](https://github.com/DeFiCh/jellyfish/blob/main/packages/testcontainers/src/containers/RegTestContainer/index.ts) instances I do not see any network assignment options as part of it.

So I am wondering if perhaps I should be using the ContainerGroup in this test perhaps?

packages/jellyfish-api-core/__tests__/category/net/getPeerInfo.test.ts
```diff
@@ -0,0 +1,106 @@
...
+      await container[0].addNode(await container[1].getIp('test'))
+    })
+
+    afterAll(async () => {
+      await container[0].stop()
+      await container[1].stop()
+
+      await container[0].removeNetwork(createdNetwork.Id)
+    })
+
```
canonbrother, line 34: added the docker network operation
here's the workflow for reference

packages/jellyfish-api-core/__tests__/category/net/getPeerInfo.test.ts
canonbrother: we don't ContainerGroup as the containers are synced once they're started
we need containers which are not synced to test `addnode`

maybe out of the box.. i added the docker network op
refer to https://github.com/DeFiCh/jellyfish/pull/938/files#r779342382

kodemon: Markings as resolved following updates from https://github.com/DeFiCh/jellyfish/pull/938#discussion_r779342382

chee-chyuan (approved):

surangap (approved):

fuxingloh (approved):

packages/jellyfish-api-core/__tests__/category/net/getPeerInfo.test.ts
```diff
@@ -0,0 +1,106 @@
+import { MasterNodeRegTestContainer, RegTestContainer } from '@defichain/testcontainers'
+import { PeerInfo } from 'packages/jellyfish-api-core/src/category/net'
+import { ContainerAdapterClient } from '../../container_adapter_client'
+import Dockerode from 'dockerode'
+
+createNetworkTests('Network without masternode', [new RegTestContainer(), new RegTestContainer()])
+createNetworkTests('Network on masternode', [new MasterNodeRegTestContainer(), new MasterNodeRegTestContainer()])
+
+function createNetworkTests (name: string, container: RegTestContainer[] | MasterNodeRegTestContainer[]): void {
+  describe(name, () => {
```
fuxingloh, line 10: Should not do this. 

Ideally, we don't want to introduce any abstraction at all to the test structure. Should take a different stance from writing code vs writing test. The test should be as straightforward, unoptimized, and as explicit as possible. 

You are trying to optimize the reading/reviewing experience of each test for the reviewer rather than the code itself. While introducing abstraction to optimize the code makes sense for non-test code, it creates a mental overhead for the reviewer.

Think of it as writing a story or narrative for your test, each scenario should be clear and obvious. 

Avoid abstraction, avoid introducing functions, avoid doing anything that is not part of the jest conventions. By doing so you are forcing the test conformity within the function. And you might lose IDE support, eslint support along the way with this pattern.

fuxingloh (changes requested):
You should optimize your test for ease of review rather than ease of implementation.

packages/jellyfish-api-core/__tests__/category/net/getPeerInfo.test.ts
kodemon: Got it, will refactor to use proper conventions.

canonbrother:
@kodemon 
my bad as missing the `waitForSync`
can you help add
```ts
await tGroup.waitForSync()
```
at `packages/jellyfish-api-core/__tests__/category/loan/takeLoan.test.ts:347`
?? :sweat_smile:

kodemon:
@canonbrother am I correct in assuming we want to perform a wait for sync operation after each generate call?

canonbrother:
> @canonbrother am I correct in assuming we want to perform a wait for sync operation after each generate call?

like this
eg:
```
alice.rpc.call...
waitForSync
// do waitForSync getting another node call
bob.rpc.call
```
after sync bob will have same data with alice

kodemon:
> after sync bob will have same data with alice

Thanks, makes sense, in this case `bob` creates a vault, and we need to wait for all containers to sync so that `alice` can perform a deposit to the vault that `bob` created.

canonbrother (approved):

fuxingloh (approved):

Issue #930, opened 2021-12-31
`ocean-api`: RawTxController `send` and `test` endpoint


fuxingloh:
> As part of #580 consolidation efforts. 

#### What would you like to be added:

After https://github.com/DeFiCh/jellyfish/pull/929 is merged. We need to add the 2 remaining endpoints to the `RawTxControlller`. https://github.com/DeFiCh/whale/blob/main/src/module.api/rawtx.controller.ts

- `rawtx/send`
- `rawtx/test`

Different from the whale development testing style, everything in `apps/ocean-api` will be e2e tests. Any spec if required should be modularly separated into their individual packages and tested in isolation.

/sig ecosystem
/area packages apps
/triage accepted

#### Why is this needed:

Providing the simplified `rawtx` endpoint for light application to interface with the blockchain. As part of the next-generation OceanApi infrastructure transition.

PR #948, opened 2022-01-05
chore(ocean-api): rawtx/test API

canonbrother:
Move `rawtx/test` from whale to jellyfish

Fixes part of #930


fuxingloh:
For the `ValidationPipe` I think we should also add `forbidUnknownValues: true`, I remember reading somewhere it's best if we enable it. But can't seem to find the issue talking about it now.


> https://docs.nestjs.com/techniques/validation

apps/ocean-api/__tests__/controllers/RawTxController.test.ts
```diff
@@ -1,40 +1,118 @@
...
+      }
+      await apiTesting.container.generate(1)
     }
-    await apiTesting.container.generate(1)
-  }
 
-  const feeRate = await apiTesting.client.rawtx.feeEstimate()
-  expect(feeRate).not.toStrictEqual(0.00005000)
+    const feeRate = await apiTesting.client.rawtx.feeEstimate()
+    expect(feeRate).not.toStrictEqual(0.00005000)
```
chee-chyuan, line 38: newbie here. im wondering if it would be better to expect the fee to be greater than 0.00005000 when there are transactions?

apps/ocean-api/__tests__/controllers/RawTxController.test.ts
```diff
@@ -1,40 +1,118 @@
...
+        code: 400,
+        type: 'BadRequest',
+        message: '400 - BadRequest  : Transaction decode failed',
+        at: expect.any(Number),
+        url: '/v1/regtest/rawtx/test'
+      })
+    }
+  })
+
+  it('should reject due to high fees', async () => {
```
chee-chyuan, line 96: question: is this test added as ain client no longer supports a boolean?

apps/ocean-api/__tests__/controllers/RawTxController.test.ts
canonbrother: min fee 0.00005 is a fixed rate
need to tight strictly using == instead of >, >=, <=, <
if ain updates/or mistakenly update the min fee rate.. then we know

apps/ocean-api/__tests__/controllers/RawTxController.test.ts
canonbrother: hmm... i don't quite get it.. support boolean???

apps/ocean-api/__tests__/controllers/RawTxController.test.ts
chee-chyuan: ar in this case cos adding a tx will increase the fee from the base tx. so im wondering if > would be better?

canonbrother: the test is testing the min fee should be 0.00005 whilst no txns

canonbrother: gosh.. i was referring wrong statement..  ignore my comment above :sweat_smile: 
in this case.. using `>=` is more accurate

apps/ocean-api/__tests__/controllers/RawTxController.test.ts
canonbrother: nope.. diff.. the test resulted the fee was greater the `maxFeeRate: 1` set, hence error
if assign bool on 2nd params should throw the error you stated above

apps/ocean-api/__tests__/controllers/RawTxController.test.ts
canonbrother: the min fee rate is `0.00001` set on ain
BTW, the `0.00005` is hardcoded when no txs, its originally `{ errors: [ 'Insufficient data or no feerate found' ], blocks: 0 }`
the fee could be lesser than `0.00005` but greater than `0.00001`
the test applied `!= 0.00005` as its not estimated

@fuxingloh kindly correct me if that's something not right on my checking

chee-chyuan: thanks for clearing the air

chee-chyuan (approved):
some doubts of mine have been answered. lgtm

kodemon (dismissed):

packages/ocean-api-client/src/errors/BadRequestException.ts
```diff
@@ -0,0 +1,13 @@
+import { ApiException, ApiErrorType } from './ApiException'
+
+export class BadRequestApiException extends ApiException {
+  constructor (message?: string) {
+    super({
+      code: 400,
+      type: ApiErrorType.BadRequest,
+      at: Date.now(),
+      message: message,
+      payload: undefined
```
canonbrother, line 10: follow the map
https://github.com/DeFiCh/jellyfish/blob/a59989a5e687b8cfefca83a83efcbfb73e7c8535/apps/libs/filters/src/ErrorFilter.ts#L51

apps/libs/filters/src/GlobalValidationPipe.ts
```diff
@@ -5,6 +5,7 @@ export class GlobalValidationPipe extends ValidationPipe {
   constructor () {
     super({
       transform: true,
+      // forbidUnknownValues: true,
```
canonbrother, line 8: reply for https://github.com/DeFiCh/jellyfish/pull/948#issuecomment-1005532103

just disabled as it's causing pg test failed
https://github.com/DeFiCh/jellyfish/blob/c271cc56402e37611e51586a254576ac216f72e8/apps/playground-api/__tests__/controllers/RpcController.test.ts#L14

kodemon (approved):

eli-lim (approved):

apps/ocean-api/__tests__/controllers/rawtx/test.test.ts
```diff
@@ -0,0 +1,77 @@
...
+beforeAll(async () => {
+  await apiTesting.start()
+  await apiTesting.testing.container.waitForWalletCoinbaseMaturity()
+})
+
+afterAll(async () => {
+  await apiTesting.stop()
+})
+
+it('should accept valid txn', async () => {
```
ivan-zynesis, line 18:
```suggestion
it('fund() - should accept valid txn', async () => {
```

canonbrother: huh??
the test is for `rawtx.test()`

cwkang1998 (approved):

apps/ocean-api/__tests__/controllers/rawtx/test.test.ts
ivan-zynesis: My bad, updated the `suggestion`. I mean the test is not enclosed with a `describe()`, but nothing major.

ivan-zynesis (approved):

PR #952, opened 2022-01-06
feat(jellyfish-api-core): add blockchain verifyChain RPC

kodemon:

/kind feature

Add verifyChain rpc as part of ongoing #48


kodemon:
@fuxingloh I was wondering if you could give me some insights into producing a invalid chain on the testing side so I can create proper failure cases for this PR.

PR #953, opened 2022-01-06
feat(jellyfish-api-core): add wallet listWallets RPC

kodemon:

/kind feature

Add listWallets rpc as part of ongoing #48

### DIFF
--- a/packages/jellyfish-api-core/__tests__/category/wallet/listWallets.test.ts
+++ b/packages/jellyfish-api-core/__tests__/category/wallet/listWallets.test.ts
@@ -22,19 +22,20 @@ describe('Wallet without masternode', () => {
 })
 
 describe('Wallet with masternode', () => {
-  const { container, rpc: { wallet } } = Testing.create(new MasterNodeRegTestContainer())
+  const master = new MasterNodeRegTestContainer()
+  const testing = Testing.create(master)
 
   beforeAll(async () => {
-    await container.start()
-    await wallet.createWallet('test')
+    await testing.container.start()
+    await testing.rpc.wallet.createWallet('test')
   })
 
   afterAll(async () => {
-    await container.stop()
+    await testing.container.stop()
   })
 
   it('should listWallets', async () => {
-    const wallets = await wallet.listWallets()
+    const wallets = await testing.rpc.wallet.listWallets()
     expect(wallets).toEqual(['', 'test'])
   })
 })
--- a/packages/jellyfish-api-core/__tests__/category/wallet/listWallets.test.ts
+++ b/packages/jellyfish-api-core/__tests__/category/wallet/listWallets.test.ts
@@ -16,7 +16,8 @@ describe('Wallet without masternode', () => {
   })
 
   it('should listWallets', async () => {
-    await expect(client.wallet.listWallets()).resolves.toEqual(['', 'test'])
+    const wallets = await client.wallet.listWallets()
+    expect(wallets).toStrictEqual(['', 'test'])
   })
 })
 
--- a/packages/jellyfish-api-core/__tests__/category/wallet/listWallets.test.ts
+++ b/packages/jellyfish-api-core/__tests__/category/wallet/listWallets.test.ts
@@ -1,0 +1,38 @@
+import { MasterNodeRegTestContainer, RegTestContainer } from '@defichain/testcontainers'
+import { ContainerAdapterClient } from '../../container_adapter_client'
+
+describe('Wallet without masternode', () => {
+  const container = new RegTestContainer()
+  const client = new ContainerAdapterClient(container)
+
+  beforeAll(async () => {
+    await container.start()
+    await client.wallet.createWallet('test')
+  })
+
+  afterAll(async () => {
+    await container.stop()
+  })
+
+  it('should listWallets', async () => {
+    await expect(client.wallet.listWallets()).resolves.toEqual(['', 'test'])
+  })
+})
+
+describe('Wallet with masternode', () => {
+  const container = new MasterNodeRegTestContainer()
+  const client = new ContainerAdapterClient(container)
+
+  beforeAll(async () => {
+    await container.start()
+    await client.wallet.createWallet('test')
+  })
+
+  afterAll(async () => {
+    await container.stop()
+  })
+
+  it('should listWallets', async () => {
+    await expect(client.wallet.listWallets()).resolves.toEqual(['', 'test'])
+  })
+})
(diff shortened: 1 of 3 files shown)


canonbrother (dismissed):

chee-chyuan (dismissed):

packages/jellyfish-api-core/__tests__/category/wallet/listWallets.test.ts
```diff
@@ -0,0 +1,38 @@
+import { MasterNodeRegTestContainer, RegTestContainer } from '@defichain/testcontainers'
+import { ContainerAdapterClient } from '../../container_adapter_client'
+
+describe('Wallet without masternode', () => {
+  const container = new RegTestContainer()
+  const client = new ContainerAdapterClient(container)
```
surangap, line 6: use `testing` instead of this.
https://github.com/DeFiCh/jellyfish/pull/957#discussion_r780976313

kodemon: Currently testing does not support `RegTestContainer`, only `MasterNodeRegTestContainer`. Should we be testing with both Master and non Master nodes or just Master node available to `@jellyfish/testing` ?

surangap: hmm, that's true. you can keep the non masternode test as it is. trying to minimize the changeset when we move the container creation inside `Testing`
https://github.com/DeFiCh/jellyfish/issues/965 will track this.

kodemon: I have a local partial version of RegTestContainer addition to testing, will put up a draft later this week.

chee-chyuan (approved):

packages/jellyfish-api-core/__tests__/category/wallet/listWallets.test.ts
```diff
@@ -0,0 +1,38 @@
+import { Testing } from '@defichain/jellyfish-testing'
+import { MasterNodeRegTestContainer, RegTestContainer } from '@defichain/testcontainers'
+import { ContainerAdapterClient } from '../../container_adapter_client'
+
+describe('Wallet without masternode', () => {
+  const container = new RegTestContainer()
+  const client = new ContainerAdapterClient(container)
```
surangap, line 7: hold till https://github.com/DeFiCh/jellyfish/pull/1035 is merged.

surangap: https://github.com/DeFiCh/jellyfish/pull/1035#discussion_r808761851

surangap (approved):

canonbrother (approved):

packages/jellyfish-api-core/__tests__/category/wallet/listWallets.test.ts
```diff
@@ -0,0 +1,38 @@
...
+    await container.start()
+    await client.wallet.createWallet('test')
+  })
+
+  afterAll(async () => {
+    await container.stop()
+  })
+
+  it('should listWallets', async () => {
+    await expect(client.wallet.listWallets()).resolves.toEqual(['', 'test'])
```
fuxingloh, line 19:
```suggestion
    const wallets = await client.wallet.listWallets()
    expect(wallets).toStrictEqual(['', 'test'])
```

For test need to optimize for better readability, the above has 3 levels of complexity in one line, you should separate into 2 lines: first line to get the data, send to test the data.

kodemon: Noted, will generate more separated assignments from now on.

packages/jellyfish-api-core/__tests__/category/wallet/listWallets.test.ts
```diff
@@ -0,0 +1,39 @@
...
+    await container.start()
+    await wallet.createWallet('test')
+  })
+
+  afterAll(async () => {
+    await container.stop()
+  })
+
+  it('should listWallets', async () => {
+    await expect(wallet.listWallets()).resolves.toEqual(['', 'test'])
```
fuxingloh, line 37: Same for this

kodemon: done, what is your thoughts on https://github.com/DeFiCh/jellyfish/pull/953/files#diff-1045b44753a823e6c6779b841a5e8d54fb6286c2961837d95fb2232b711b7218R25 ?

Would you expect something more like

```ts
const master = new MasterNodeRegTestContainer()
const { container, rpc: { wallet } } = Testing.create(master)
```

or

```ts
const master = new MasterNodeRegTestContainer()
const container = Testing.create(master)
```

Personally I am more leaning towards the way its currently written, but would be good to align with expectations in Jellyfish for these sort of approaches.

Pulling out variables of objects like this was a pretty neat ES update and I find myself using it quite often in other projects for method arguments and variable assignment.

[Destructuring Assignment](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment)

fuxingloh: IMO, destructuring is great in functional method parameters, but **in tests**, it's pretty counterintuitive as it introduces too much symbol memorization when one is trying to understand what is going on.

I am also towards reducing unnecessary symbols that are not part of the testing context. By only introducing symbols that are part of the testing context. Optimizing for test readability over anything else, as for writing test and writing component we should take a different approach.

E.g.
```ts
const master = new MasterNodeRegTestContainer()
const { container, rpc: { wallet } } = Testing.create(master)
```

For the above, you expose 2 new symbols for the reader to track and understand as they are reading the test.

Vs.
```ts
const testing = Testing.create(new MasterNodeRegTestContainer())
```

For the above, you only have 1 symbol which is the "`testing`" framework object for the reader to follow along.

Subsequently:

```ts
testing.container.generate(1)
testing.rpc.wallet.doSomething(1)
```

It's contextually more understandable for the user to know that you are accessing the testing - container - to generate 1 block.

That's for tests. And it only makes sense **for tests**, as you are trying to make it as effortless as possible for the reviewer to understand what's going on. **However for non-tests**, I am in favor of destructuring as it reduces non-relevant symbols by focusing on what is required for each method/function.

kodemon: Great, thanks for the feedback and I agree with your views on this. Will update the tests to reflect this pattern.

kodemon: Updated tests, thanks again for the input on this 👍🏻

fuxingloh (approved):

PR #954, opened 2022-01-06
docs(whale-api-client): Token Balances & Vault for `/address/` endpoint

matteshe:

The Ocean REST API is not fully documented. I added some more endpoints which can be used to get information on the defichain

Fixes #


fuxingloh (approved):
I cleaned up the docs for publishing

PR #964, opened 2022-01-12
chore(lint): add no only on precommit

canonbrother:
happy dev

annoying while forgot to remove then push
time wasted on CI
image

Fixes #


fuxingloh:
I think we might be able to accomplish this on eslint?

fuxingloh:
Probably also a good idea to update the enforcement of testing with `toStrictEqual` 

> https://www.npmjs.com/package/eslint-plugin-jest might be useful

.eslintrc
```diff
@@ -34,7 +35,9 @@
       }
     ],
     "import/no-default-export": "error",
-    "no-else-return": "error"
+    "no-else-return": "error",
+    "jest/no-disabled-tests": "error",
+    "jest/prefer-strict-equal": "error"
```
canonbrother, line 40: FYI, it only restricts `toEqual` but `toBe`

fuxingloh (approved):

fuxingloh:
These eslint changes highlight a few linting issues with our existing test, need to decide whether we should silent them or follow convention.

PR #966, opened 2022-01-12
feat(jellyfish-testing): add regtestcontainer support

### LINKED ISSUE #965: `packages/jellyfish-testing`: Add support to accept `RegTestContainer` in `Testing`
check the possibility of accepting a `RegTestContainer` instead of `MasterNodeRegTestContainer` in `Testing` class.

#### Why is this needed:
there can be tests that specifically require non-masternodes.

kodemon:

Since we are moving our container testing to exclusively use jellyfish-testing package we need to enable support for RegTestContainer scenarios.

Fixes #965 

#### Additional comments?:

This PR adds two additional testing classes

**TestingLight**

Works the same way as `Testing` but facilitates containers using `RegTestContainer` instead of `MasterNodeRegTestContainer`. The secondary change is that `TestingLight` has an additional `master` property attached to it.

Sample

```ts
const testing = TestingLight.create(new RegTestContainer())

beforeAll(async () => {
  await testing.start() // starts both master and regtest container
  await testing.master.generate(1) // wait for master node to generate a block
})

afterAll(async () => {
  await testing.stop() // stops both master and regtest container
})

await testing.rpc.net.getConnectionCount()
```

**TestingLightGroup**

Same as `TestingGroup` but facilitates `LightTesting` containers.

### Note

After some inquiry into the behavior and functionality in our test cases, without and with master node the ticket and PR aiming to introduce this testing layer may be moot.

Since the test cases `without master node` is essentially performing tests against containers that doesn't have master nodes, it may be moot to add a master node instance to a new `Light` testing container.


packages/testcontainers/src/containers/RegTestContainer/index.ts
```diff
@@ -37,11 +37,27 @@ export class RegTestContainer extends DeFiDContainer {
...
     return await this.getPort('19554/tcp')
   }
+
+  async generate (nblocks: number, address?: string, maxTries: number = 1000000): Promise<void> {
+    throw new Error('RegTestContainer Violation: generate is not implemented for non master node containers')
+  }
+
+  async fundAddress (address: string, amount: number): Promise<{ txid: string, vout: number }> {
+    throw new Error('RegTestContainer Violation: fundAddress is not implemented for non master node containers')
+  }
```
kodemon, line 62: Commenting on this file as a whole, I don't think we should be needed to introduce this sort of methods. And we don't strictly need these for testing in our packages outside of jellyfish-testing since the returned type of Container is correctly type on instantiation.

```
const testing = Testing.create(new RegTestContainer())
```

Will correctly only display functionality available to it.

That however doesn't work internally in jellyfish-testing if we look at the https://github.com/DeFiCh/jellyfish/blob/kodemon/testing-regtest-support/packages/jellyfish-testing/src/testing.ts#L17-L22 part of testing instance since these sub dependencies assumes the container is `MasterNodeRegTestContainer`.

So we may need to adjust the sub dependencies to be flexible and not make container type assumptions.

kodemon:
This is resolved by #1035

PR #990, opened 2022-01-19
chore(playground): 100 deterministic blocks

### LINKED ISSUE #914: `@defichain/playground`: GenesisCoinbaseBot to submit 100 deterministic blocks

A bot that submits 100 deterministic blocks so we can set up consistent playground behavior for greater "light" implementation and testing. Also not relying on `defid` wallet for better jellyfish development agility.

What should the 100 blocks be is not yet decided but it should be basis like setting up DEX, Oracle, Loan and etc.

/area packages
/sig ecosystem
/triage accepted

canonbrother:

kind/feature

Fixes #914

### DIFF
--- a/packages/playground/src/bots/GenesisCoinbaseBot.ts
+++ b/packages/playground/src/bots/GenesisCoinbaseBot.ts
@@ -7,7 +7,6 @@ import { AppointOracle, OP_CODES, Script, CTransactionSegWit, TransactionSegWit,
 import { ApiClient } from '@defichain/jellyfish-api-core'
 import { OraclePriceFeed } from '@defichain/jellyfish-api-core/src/category/oracle'
 import { SmartBuffer } from 'smart-buffer'
-import { P2WPKH } from '@defichain/jellyfish-address'
 
 enum PriceDirection {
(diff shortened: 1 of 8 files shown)


packages/playground/src/bots/GenesisCoinbaseBot.ts
```diff
@@ -0,0 +1,719 @@
...
+  }
+
+  async cycle (n: number): Promise<void> {
+  }
+
+  async setup (): Promise<void> {
+    await this.setupTokens()
+    await this.setupLoanSchemes()
+    await this.setupOracles()
+    // await new Promise(resolve => setInterval(() => resolve(this.setupOracleData()), 6000))
```
canonbrother, line 185: no idea how to work like `@interval` in nestjs yet
https://github.com/DeFiCh/playground/blob/d35822dd50955adbffe23f9b66e804a3fa9fb6c5/src/module.playground/bot/oracle.bot.ts#L138

canonbrother: move to `cycle`

PR #1003, opened 2022-01-25
feat(jellyfish-network): isSkippedTxId for MainNet & TestNet

fuxingloh:

https://github.com/DeFiCh/ain/pull/476

Introduce `isSkippedTxId` for `@defichain/jellyfish-network`. Post Dakota hard fork we do not allow tx to be skipped, so a tx that's a part of the block at the main chain is applied otherwise we search in a static array of knowing tx that are marked as skipped.


surangap (commented):
looks like we don't have an RPC  exactly for this need. 
This way might have to maintain the list on both sides.
might be worth adding it to cpp side in the future?

packages/jellyfish-network/src/TransactionSkipped.ts
```diff
@@ -0,0 +1,373 @@
+import { Network } from './Network'
+
+const MainNetSkippedTxId = new Set([
```
kodemon, line 3:
```suggestion
const MainNetSkippedTxIds = new Set([
```
Wonder if this should assigned in plural form as its a Set of ids

packages/jellyfish-network/src/TransactionSkipped.ts
```diff
@@ -0,0 +1,373 @@
...
+  '938628f96d7e1bf5fada66b1681c8cf3731463451a66a9a792e099ca09f59910',
+  '58484171dbe293667f81b76c6f0d493928a92543bf46076cc62de3d0d1ce02cf',
+  '0863180da779d530ff5184908bc1ddd85910a5c643d1317fe0f3796f47eba414',
+  '27273069347a23a4e5fee10346ab67813f99b8e09c3d31d4b1c4384fa6bed588',
+  'f872f706976de4ebf97429946b83d7c1e0480496f26442d4253440a0f462ba82',
+  '949c7243483d52b85e6ef058fe8814b5fd6b307a529fd34c07daa8eae5759770',
+  '7a1001461506e3b5a4de3c3de74a1a838ab6b6d1530042ff9bd802e6bda90e91'
+])
+
+const TestNetSkippedTxId = new Set([
```
kodemon, line 349:
```suggestion
const TestNetSkippedTxIds = new Set([
```

packages/jellyfish-network/src/TransactionSkipped.ts
```diff
@@ -0,0 +1,373 @@
...
+  'af264140932d5ec7049639ae1d8d3453a307d6ed3272a7d12ead4d09370cca9c',
+  'a65a1c2e37461c01e19f99675e448eb23b64d0dd1234547a30b45ebf5e9e7ba2',
+  'dfd54905c6a323d92fd923a2fcc592e005fa8488e2f3774cb585f7fd33068433',
+  'f5166734c9fe55d46325c2406454daddc698ff9d76491dff8e6553027d53878f',
+  'e30ad732c4ec80946a9aef649e3eef521bf408f29539c0ff254b67272cc908ea'
+])
+
+export function isSkippedTxId (txId: string, network: Network['name'] = 'mainnet'): boolean {
+  if (network === 'mainnet') {
+    return MainNetSkippedTxId.has(txId)
```
kodemon, line 367:
```suggestion
    return MainNetSkippedTxIds.has(txId)
```

packages/jellyfish-network/src/TransactionSkipped.ts
```diff
@@ -0,0 +1,373 @@
...
+  }
+  if (network === 'testnet') {
+    return TestNetSkippedTxId.has(txId)
```
kodemon, line 370:
```suggestion
    return TestNetSkippedTxIds.has(txId)
```

kodemon (commented):
`nit:` `opinion:` I am thinking `Set` is a plural assignment in terms of naming. It reads more naturally when doing a `txIds.has(id)` vs `txId.has(id)`. If I were to import this into another file I would potentially think that `MainNetSkippedTxId` is a single value reference.

fuxingloh:
> looks like we don't have an RPC exactly for this need. This way might have to maintain the list on both sides. might be worth adding it to cpp side in the future?

Nope, I would say it's an one off implementation. On cpp side, currently included in `decodecustomtx` rpc as "valid".

packages/jellyfish-network/__tests__/TransactionSkipper.test.ts
```diff
@@ -0,0 +1,23 @@
...
+it('should not be skipped on MainNet as txId is on TestNet', () => {
+  expect(isSkippedTxId('4f5f620484e5359eafb9fe799b568a59ca75202bfe6aa1546c54820c99889437')).toStrictEqual(false)
+})
+
+it('should be skipped on TestNet', async () => {
+  expect(isSkippedTxId('4f5f620484e5359eafb9fe799b568a59ca75202bfe6aa1546c54820c99889437', 'testnet')).toStrictEqual(true)
+  expect(isSkippedTxId('b22fca714f70f68e45bef2babc5d7f0a1c81fd892ddfd347edb43cc80fc31db2', 'testnet')).toStrictEqual(true)
+})
+
+it('should not be be skipped on TestNet as txID is on MainNet', async () => {
```
fuxingloh, line 21:
```suggestion
it('should not be skipped on TestNet as txID is on MainNet', async () => {
```

surangap (approved):

kodemon (approved):

Issue #943, opened 2022-01-04
`/apps/ocean-api`: CD Publish Docker Image


fuxingloh:
> As part of #580 consolidation efforts. 
> https://github.com/DeFiCh/whale/tree/main/.github/workflows

#### What would you like to be added:

On "Release", "PR into main" and "merge into main"; build and publish a docker image of `apps/ocean-api`.

- Release: Tagged with `release version` and `latest`
- PR into main: Tagged with `commit hash` and `main`
- PR for merge into main: Tagged with `commit hash` and `pr-{number}`

#### Why is this needed:

Shift left.

/sig ecosystem
/area apps
/triage accepted

fuxingloh:
To setup build cache also, 40 minutes on ARM is too long.

fuxingloh:
/priority important-soon

Btw we need this prioritized over #942

kodemon:
Understood, will focus on this one.

Issue #942, opened 2022-01-04
`/apps/ocean-api`: CI Build Docker Image with Sanity Testing


fuxingloh:
> As part of #580 consolidation efforts. 
> https://github.com/DeFiCh/whale/tree/main/.github/workflows

#### What would you like to be added:

On PR into main, build a docker image of `apps/ocean-api` to check the docker artifact can be created without errors. Followed by bootstrapping the docker container with sanity testing to make sure it works end to end before publishing. 

The short and simple test should be proactively added, it should also be committed into `apps/ocean-api/__sanity__/` to keep the concerns tight.

On the old implementation at DeFiCh/whale, the testing was conducted with [Postman - Newman Action](https://github.com/marketplace/actions/newman-action). This is however redundant as we can just continue using jest setup to connect directly to the container.

#### Why is this needed:

Shift left.

/sig ecosystem
/area apps
/triage accepted

fuxingloh:
/priority important-soon as more and more apps are integrated in

fuxingloh:
Closed with https://github.com/JellyfishSDK/jellyfish/pull/1314

PR #1005, opened 2022-01-26
feat(whale): migrate `apps/whale-api`

### LINKED ISSUE #977: `/apps/whale-api` - Consolidation & Migration

/sig ecosystem
/area apps
/triage accepted

kodemon:

Consolidation of our public apis into the jellyfish ecosystem.

Fixes #977

#### Additional comments?:

Depends on following PRs:

 - https://github.com/DeFiCh/jellyfish/pull/1167

Moves `whale/src` to `apps/whale-api` and fixing imports that are no longer valid in the new ecosystem. Replaces `@src` targets to relative `../src` imports. After this went file by file to make sure imports are not repeated and not targeting `dist`. Because of this a few jellyfish packages has updated indexes to export variables needed by whale. `export { x }` > `export * as x`.


kodemon:
Quick update here to spark some discussion on how to approach this migration.

Now I am imagine the end result here is that we want to consolidate all our development efforts into jellyfish. I am not quite sure how we see ourselves optimally make this transition. Mainly on the back thought that whale still has open PRs.

I think currently my first priority is to get whale working in jellyfish, once that is done I think we can formulate a final strategy for how we want to transfer current work in Whale over into Jellyfish. I do see some older PRs that we can just have each author port over manually.

I have at least set it up to easily move updates to Whale over into this PR as we go so that is not a huge concern atm. The biggest point to be made is that once we start working in jellyfish rather than in whale we want to lock down the src at that time and not do anymore work in whale.

I am still working out some kinks to get tests to work, and I also need to get it running in jellyfish. Once that is done I think we can pull the trigger on transferring work from whale to jellyfish.

Please let me know your thoughts and correct some of my misconceptions where applicable 🙏🏻

kodemon:
For some reason in `whale` if you delete the `package-lock.json` file and `node_modules` folder and do a `npm i` the following [line of code](https://github.com/DeFiCh/whale/blob/main/src/module.database/provider.memory/memory.database.ts#L11) fails.

After poking @canonbrother about this issue we are both rather baffled by why this would be the case since the lock files doesn't seem to change the dependencies in any discernible way.

This is currently presenting a small blocker for progressing on this PR.

fuxingloh:
> I think currently my first priority is to get whale working in jellyfish, once that is done I think we can formulate a final strategy for how we want to transfer current work in Whale over into Jellyfish. I do see some older PRs that we can just have each author port over manually.

Yup, that's sound first priority towards getting `DeFiCh/whale` project to fit into `DeFiCh/jellyfish` structure.

tsconfig.base.json
```diff
@@ -3,7 +3,7 @@
     "lib": [
       "es2020"
     ],
-    "target": "es6",
+    "target": "es2020",
```
kodemon: Changed target to `es2020` to properly support certain functionality in TypeScript.

The [testing](https://github.com/DeFiCh/jellyfish/blob/main/packages/jellyfish-testing/src/testing.ts#L14-L19) is using some functionality that errors using `es6`. Error is `Property '$1' is used before its initialization. ts(2729)`

Following the trail of the following typescript [ticket 1](https://github.com/microsoft/TypeScript/issues/37573) [ticket 2](https://github.com/microsoft/TypeScript/issues/44298) and documentation [useDefineForClassFields](https://www.typescriptlang.org/tsconfig#useDefineForClassFields).

When setting the base target to `es2020` the error goes away as expected.

tsconfig.json
```diff
@@ -7,7 +7,6 @@
         "packages/*/src"
       ]
     },
-    "target": "esnext",
```
kodemon: Now using base target across jellyfish, I suggest we override this in individual build configs if we need to.

apps/whale-api/migrate.js
```diff
@@ -0,0 +1,61 @@
...
+    return './'
+  }
+  return new Array(depth).fill('../').reduce((path, val) => {
+    return path + val
+  }, '')
+}
+
+migrate(path.resolve(__dirname, 'src')).then(() => {
+  process.exit(0)
+})
```
kodemon, line 61: This script updates all the import paths to no longer use alias paths defined in [whale](https://github.com/DeFiCh/whale/blob/main/tsconfig.json#L18-L21) as they are not supported in jellyfish structure. Alias paths are defined in the jellyfish root and jest does not properly support sub aliases.

My solution to this is to just make all the `@src` imports relative and remove any `/dist` import targets to be pure package imports. Eg. `jellyfish-package/dist/...` => `jellyfish-package`. Because of this I have updated a couple of jellyfish package [exports](https://github.com/DeFiCh/jellyfish/blob/kodemon/whale-api-client-migration/packages/jellyfish-api-core/src/index.ts#L38-L55) to support this.

Once we run migration there is only a handful of imports that needs updating. This approach allows us to simply just copy the entire `src` folder from whale to jellyfish. Run this script, update a couple of imports and we are good to go. Once we are ready to switch our development efforts of whale in jellyfish ecosystem we can run a final migration pass and delete this file.

apps/whale-api/custom.d.ts
```diff
@@ -0,0 +1,2 @@
+declare module 'browserify-aes'
+declare module 'bip66'
```
kodemon, line 2: These came up complaining and has no typing available so added module declares here.

kodemon (commented):
I got the initial things working here, tests and builds should be passing. I added some comments to highlight some of the more notable changes.

The primary idea here is to be able to easily move over any upcoming changes from whale to jellyfish until we are ready to completely switch over.

How we optimally do this is still up in the air, but once this PR is fully ready we can discuss how we want to accomplish the move in more detail.

tsconfig.base.json
fuxingloh: https://github.com/DeFiCh/jellyfish/pull/902

We had to set back to "es6" because downstream (React Native + Expo) couldn't compile. Would it be possible to set this for just this project?

fuxingloh: I think that's also the reason why we set target to "esnext" for `tsconfig.json` when developing.

kodemon: Noted, thanks for the reference. Let me do some more testing on this to confirm.

kodemon: @fuxingloh that worked fine, reverted tsconfig.base back to `es6` and set `es2020` for tsconfig which removes any development side complaints for testing. `esnext` kicks of with errors on that functionality.

kodemon:
Quick update on progress here.

I have finished the initial structured migration of whale into jellyfish. Tests should be passing and builds are executing successfully.

I think the next step here is to finish up whale priority PRs, and once they are merged I do a final port of whale src code and finalize the PR ready for review _(meaning we no longer expect anymore merges in whale)_.

Once that is done I will need an assist in ensuring all the deployment stuff is good to go. Once done this should be good to merge and we can continue any performance tweaking in jellyfish going forwards.

apps/whale-api/custom.d.ts
fuxingloh: https://github.com/DeFiCh/jellyfish/pull/1027

kodemon:
@fuxingloh need your assist/clarification on two subjects.

State of whale in relation to this migration. As it stands whale still has active PRs, so I think that this PR can stay frozen until there are no priority PRs in whale or we decide not to merge anymore PRs and rather focus on moving work over to jellyfish somehow. Please advice what steps you think is the best here in terms of what approach to take as I think its now technically ready to go _(just needs a final version of the whale code before we can merge)_

Second query is about deployments, I am assuming whale api gets deployed somewhere. I am not seeing an obvious path of deployment, could you point me in the right direction on how we can confirm the jellyfish setup works as needed for deployment purposes now that it will be deploying out of jellyfish > whale.

Will focus on other tasks until we are in a more defined space for finalization on this PR.

fuxingloh:
>State of whale in relation to this migration. As it stands whale still has active PRs, so I think that this PR can stay frozen until there are no priority PRs in whale or we decide not to merge anymore PRs and rather focus on moving work over to jellyfish somehow. Please advice what steps you think is the best here in terms of what approach to take as I think its now technically ready to go _(just needs a final version of the whale code before we can merge)_

Yup I agree, I think it's 1-2 weeks away and we can freeze DeFiCh/whale development for migration.


> Second query is about deployments, I am assuming whale api gets deployed somewhere. I am not seeing an obvious path of deployment, could you point me in the right direction on how we can confirm the jellyfish setup works as needed for deployment purposes now that it will be deploying out of jellyfish > whale.

It's related to this 2 tasks:
- https://github.com/DeFiCh/jellyfish/issues/942
- https://github.com/DeFiCh/jellyfish/issues/943

> Will focus on other tasks until we are in a more defined space for finalization on this PR.

Thanks!

kodemon:
@fuxingloh quick ping for time-frame update

PR #1029, opened 2022-02-04
feat(jellyfish-testing): created testingwrapper as an abstraction layer

### LINKED ISSUE #637: Move test container creation inside the `jellyfish-testing` and refactor tests.
at the moment, to create a `Testing` instance, the container needs to be given as an input parameter.
need to change this so that by default, the container(`MasterNodeRegTestContainer`) is created within the `Testing` class itself. If different type of container is needed, it should be able to pass from outside.

finally in a test, it should be able to call as follows
for single container, `Testing.create(1)`
for multinode setup with 3 nodes,  `Testing.create(3)`

#### Why is this needed:
remove a lot of duplicate code.
much cleaner, simpler

chee-chyuan:
create a new layer of abstraction with the goal of being able to create a test /testGroup instance in a standardised manner.

Fixes #637

#### Additional comments?:
Using the wrapper class
- should be able to create a single testing instance regardless of container
- should be able to create a new group and add new testing instance

### DIFF
--- a/packages/jellyfish-testing/src/testing.ts
+++ b/packages/jellyfish-testing/src/testing.ts
@@ -6,24 +6,31 @@ import { TestingRawTx } from './rawtx'
 import { TestingICX } from './icxorderbook'
 import { TestingMisc } from './misc'
 import { TestingGroupAnchor } from './anchor'
-import { ContainerGroup, MasterNodeRegTestContainer, StartOptions } from '@defichain/testcontainers'
+import { ContainerGroup, DeFiDContainer, MasterNodeRegTestContainer, StartOptions } from '@defichain/testcontainers'
 import { RegTestFoundationKeys } from '@defichain/jellyfish-network'
 import { JsonRpcClient } from '@defichain/jellyfish-api-jsonrpc'
 
 export class Testing {
+  public readonly container: MasterNodeRegTestContainer
   public readonly fixture = new TestingFixture(this)
-  public readonly token = new TestingToken(this.container, this.rpc)
-  public readonly poolpair = new TestingPoolPair(this.container, this.rpc)
-  public readonly rawtx = new TestingRawTx(this.container, this.rpc)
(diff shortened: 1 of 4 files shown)


packages/jellyfish-testing/src/testing.ts
```diff
@@ -6,24 +6,31 @@ import { TestingRawTx } from './rawtx'
...
-  public readonly token = new TestingToken(this.container, this.rpc)
-  public readonly poolpair = new TestingPoolPair(this.container, this.rpc)
-  public readonly rawtx = new TestingRawTx(this.container, this.rpc)
-  public readonly icxorderbook = new TestingICX(this)
-  public readonly misc = new TestingMisc(this.container, this.rpc)
+  public readonly token: TestingToken
+  public readonly poolpair: TestingPoolPair
+  public readonly rawtx: TestingRawTx
+  public readonly icxorderbook: TestingICX
+  public readonly misc: TestingMisc
```
canonbrother, line 20: that's the issue and solution found out by @kodemon 
`Property xXx is used before its initialization`
preferably fix it via tsconfig, use es2020

canonbrother, line 20: that's the issue and solution found out by @kodemon 
`Property xXx is used before its initialization`
preferably fix it via tsconfig, use es2020

chee-chyuan: yea noticed it too before but it wasnt bugging me. now it is as im changing the constructor param abit. im now passing in a more generic DefiDContainer instead of `MasterNodeRegTestContainer` and forcibly cast it as `MasterNodeRegTestContainer` hence the change in the constructor

kodemon: Quick [reference](https://github.com/DeFiCh/jellyfish/blob/kodemon/whale-api-client-migration/tsconfig.json#L10) to change in whale migration used to solve this issue.

Issue #100, opened 2021-04-07
@defichain/testcontainers should automatically update itself when there is an release


fuxingloh:
#### What would you like to be added:

Automatically create a pull request to jellyfish/packages/testcontainers with the update docker version.

#### Why is this needed:

1. Manually creating PR is tiresome, this pipeline should be automated.
2. PR will be automatically tested, reviewer can just approve and merge them in and get immediate CI checks.

fuxingloh:
This capability is now managed by Blockchain team to test jellyfish compatibility before release, we don't need this anymore as each major release PR will be tested on jellyfish before release/merged.

https://github.com/DeFiCh/jellyfish/pull/184

PR #1035, opened 2022-02-07
feat(jellyfish-testing): created testingwrapper as an abstraction layer 

### LINKED ISSUE #637: Move test container creation inside the `jellyfish-testing` and refactor tests.

### LINKED ISSUE #965: `packages/jellyfish-testing`: Add support to accept `RegTestContainer` in `Testing`

chee-chyuan:
create a new layer of abstraction with the goal of being able to create a test /testGroup instance in a standardised manner.

Fixes #637
Fixes #965 

#### Additional comments?:
Using the wrapper class
- should be able to create a single testing instance regardless of container
- should be able to create a new group and add new testing instance

### DIFF
--- a/packages/jellyfish-testing/__tests__/testingcontainer.test.ts
+++ b/packages/jellyfish-testing/__tests__/testingcontainer.test.ts
@@ -23,7 +23,7 @@ describe('create a single test container using testwrapper', () => {
   })
 
   it('should be able to create regtest non masternode container', async () => {
-    testing = testingWrapper.create(1, index => new RegTestContainer()) as Testing
+    testing = testingWrapper.create(1, () => new RegTestContainer())
     await testing.container.start()
 
     // call rpc
@@ -37,7 +37,7 @@ describe('create a single test container using testwrapper', () => {
   })
 
   it('should be able to create mainnet container', async () => {
-    testing = testingWrapper.create(1, index => new MainNetContainer()) as Testing
+    testing = testingWrapper.create(1, () => new MainNetContainer())
     await testing.container.start()
     // call rpc
     const { chain } = await testing.container.getMiningInfo()
@@ -45,7 +45,7 @@ describe('create a single test container using testwrapper', () => {
   })
 
   it('should be able to create testnet container', async () => {
-    testing = testingWrapper.create(1, index => new TestNetContainer()) as Testing
+    testing = testingWrapper.create(1, () => new TestNetContainer())
     await testing.container.start()
     // call rpc
     const { chain } = await testing.container.getMiningInfo()
@@ -101,7 +101,7 @@ describe('create multiple test container using testwrapper', () => {
     expect(blockHashAlice).toStrictEqual(blockHashBob)
 
     // create a non masternode RegTestTesting
-    const nonMasternodeTesting = testingWrapper.create(1, index => new RegTestContainer()) as Testing
+    const nonMasternodeTesting = testingWrapper.create(1, () => new RegTestContainer())
     await nonMasternodeTesting.container.start()
 
     // add to group
@@ -118,7 +118,7 @@ describe('create multiple test container using testwrapper', () => {
   it('should be able to create individual container of masternode and regtest and create a group from it', async () => {
     const masternodeTesting = testingWrapper.create()
     await masternodeTesting.container.start()
-    const nonmasternodeTesting = testingWrapper.create(1, (index) => new RegTestContainer()) as Testing
+    const nonmasternodeTesting = testingWrapper.create(1, () => new RegTestContainer())
     await nonmasternodeTesting.container.start()
 
     const tGroup = testingWrapper.group([masternodeTesting, nonmasternodeTesting])
--- a/packages/jellyfish-testing/__tests__/testingcontainer.test.ts
+++ b/packages/jellyfish-testing/__tests__/testingcontainer.test.ts
@@ -61,6 +61,15 @@ describe('create a single test container using testwrapper', () => {
     // eslint-disable-next-line @typescript-eslint/dot-notation
     expect(softforks['fortcanninghill'].height).toStrictEqual(11)
   })
+
+  it('should create 1 testing instance even if 0 is passed in as a param', async () => {
+    testing = testingWrapper.create(0, () => new RegTestContainer())
+    await testing.container.start()
+
+    // call rpc
+    const blockHeight = await testing.rpc.blockchain.getBlockCount()
+    expect(blockHeight).toStrictEqual(0)
+  })
 })
 
 describe('create multiple test container using testwrapper', () => {
--- a/packages/jellyfish-testing/__tests__/testingcontainer.test.ts
+++ b/packages/jellyfish-testing/__tests__/testingcontainer.test.ts
@@ -121,11 +121,8 @@ describe('create multiple test container using testwrapper', () => {
     const nonmasternodeTesting = testingWrapper.create(1, (index) => new RegTestContainer()) as Testing
     await nonmasternodeTesting.container.start()
 
-    const tGroup = testingWrapper.group()
+    const tGroup = testingWrapper.group([masternodeTesting, nonmasternodeTesting])
     await tGroup.start()
-
-    await tGroup.addTesting(masternodeTesting)
-    await tGroup.addTesting(nonmasternodeTesting)
 
     await masternodeTesting.container.generate(1)
     await tGroup.waitForSync()
--- a/packages/jellyfish-testing/__tests__/testingcontainer.test.ts
+++ b/packages/jellyfish-testing/__tests__/testingcontainer.test.ts
@@ -2,7 +2,6 @@ import { Testing } from '@defichain/jellyfish-testing'
 import { MainNetContainer, RegTestContainer, StartFlags, TestNetContainer } from '@defichain/testcontainers'
 import { TestingWrapper } from '../src/testingwrapper'
 
-const testingWrapper = new TestingWrapper()
 describe('create a single test container using testwrapper', () => {
   let testing: Testing
 
@@ -11,7 +10,7 @@ describe('create a single test container using testwrapper', () => {
   })
 
   it('should be able to create and call single regtest masternoderegtest container', async () => {
-    testing = testingWrapper.create()
+    testing = TestingWrapper.create()
     await testing.container.start()
 
     // call rpc
@@ -23,7 +22,7 @@ describe('create a single test container using testwrapper', () => {
   })
 
   it('should be able to create regtest non masternode container', async () => {
-    testing = testingWrapper.create(1, () => new RegTestContainer())
+    testing = TestingWrapper.create(1, () => new RegTestContainer())
     await testing.container.start()
 
     // call rpc
@@ -37,7 +36,7 @@ describe('create a single test container using testwrapper', () => {
   })
 
   it('should be able to create mainnet container', async () => {
-    testing = testingWrapper.create(1, () => new MainNetContainer())
+    testing = TestingWrapper.create(1, () => new MainNetContainer())
     await testing.container.start()
     // call rpc
     const { chain } = await testing.container.getMiningInfo()
@@ -45,7 +44,7 @@ describe('create a single test container using testwrapper', () => {
   })
 
   it('should be able to create testnet container', async () => {
-    testing = testingWrapper.create(1, () => new TestNetContainer())
+    testing = TestingWrapper.create(1, () => new TestNetContainer())
     await testing.container.start()
     // call rpc
     const { chain } = await testing.container.getMiningInfo()
@@ -53,7 +52,7 @@ describe('create a single test container using testwrapper', () => {
   })
 
   it('should be able to override start option', async () => {
-    testing = testingWrapper.create()
+    testing = TestingWrapper.create()
     const startFlags: StartFlags[] = [{ name: 'fortcanninghillheight', value: 11 }]
     await testing.container.start({ startFlags })
 
@@ -63,7 +62,7 @@ describe('create a single test container using testwrapper', () => {
   })
 
   it('should create 1 testing instance even if 0 is passed in as a param', async () => {
-    testing = testingWrapper.create(0, () => new RegTestContainer())
+    testing = TestingWrapper.create(0, () => new RegTestContainer())
     await testing.container.start()
 
     // call rpc
@@ -74,7 +73,7 @@ describe('create a single test container using testwrapper', () => {
 
 describe('create multiple test container using testwrapper', () => {
   it('should create a masternode group and test sync block and able to add and sync non masternode container', async () => {
-    const tGroup = testingWrapper.create(2)
+    const tGroup = TestingWrapper.create(2)
     await tGroup.start()
 
     const alice = tGroup.get(0)
@@ -110,7 +109,7 @@ describe('create multiple test container using testwrapper', () => {
     expect(blockHashAlice).toStrictEqual(blockHashBob)
 
     // create a non masternode RegTestTesting
-    const nonMasternodeTesting = testingWrapper.create(1, () => new RegTestContainer())
+    const nonMasternodeTesting = TestingWrapper.create(1, () => new RegTestContainer())
     await nonMasternodeTesting.container.start()
 
     // add to group
@@ -125,12 +124,12 @@ describe('create multiple test container using testwrapper', () => {
   })
 
   it('should be able to create individual container of masternode and regtest and create a group from it', async () => {
-    const masternodeTesting = testingWrapper.create()
+    const masternodeTesting = TestingWrapper.create()
     await masternodeTesting.container.start()
-    const nonmasternodeTesting = testingWrapper.create(1, () => new RegTestContainer())
+    const nonmasternodeTesting = TestingWrapper.create(1, () => new RegTestContainer())
     await nonmasternodeTesting.container.start()
 
-    const tGroup = testingWrapper.group([masternodeTesting, nonmasternodeTesting])
+    const tGroup = TestingWrapper.group([masternodeTesting, nonmasternodeTesting])
     await tGroup.start()
 
     await masternodeTesting.container.generate(1)
--- a/packages/jellyfish-testing/src/testing.ts
+++ b/packages/jellyfish-testing/src/testing.ts
@@ -89,7 +89,7 @@ export class TestingGroup {
     n: number,
     init?: (index: number) => DeFiDContainer
   ): TestingGroup {
-    const containers: MasterNodeRegTestContainer[] = []
+    const containers: DeFiDContainer[] = []
     const testings: Testing[] = []
 
     if (init === undefined) {
@@ -97,10 +97,10 @@ export class TestingGroup {
     }
 
     for (let i = 0; i < n; i += 1) {
-      const container = init(i) as MasterNodeRegTestContainer
+      const container = init(i)
       containers.push(container)
 
-      const testing = Testing.create(container)
+      const testing = Testing.createBase(container)
       testings.push(testing)
     }
 
--- a/packages/jellyfish-testing/src/testing.ts
+++ b/packages/jellyfish-testing/src/testing.ts
@@ -10,28 +10,27 @@ import { ContainerGroup, DeFiDContainer, MasterNodeRegTestContainer, StartOption
 import { RegTestFoundationKeys } from '@defichain/jellyfish-network'
 import { JsonRpcClient } from '@defichain/jellyfish-api-jsonrpc'
 
-export class Testing<Container extends DeFiDContainer = DeFiDContainer> {
-  public readonly fixture: TestingFixture
-  public readonly icxorderbook: TestingICX
-
-  public readonly token!: Container extends MasterNodeRegTestContainer ? TestingToken : undefined
-  public readonly poolpair!: Container extends MasterNodeRegTestContainer ? TestingPoolPair : undefined
-  public readonly rawtx!: Container extends MasterNodeRegTestContainer ? TestingRawTx : undefined
-  public readonly misc!: Container extends MasterNodeRegTestContainer ? TestingMisc : undefined
+export class Testing {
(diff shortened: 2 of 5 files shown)


packages/jellyfish-testing/__tests__/testingcontainer.test.ts
```diff
@@ -0,0 +1,139 @@
...
+    await tGroup.stop()
+  })
+
+  it('should be able to create individual container of masternode and regtest and create a group from it', async () => {
+    const masternodeTesting = testingWrapper.create()
+    await masternodeTesting.container.start()
+    const nonmasternodeTesting = testingWrapper.create(1, (index) => new RegTestContainer()) as Testing
+    await nonmasternodeTesting.container.start()
+
+    const tGroup = testingWrapper.group()
```
surangap, line 124: how about we add `testingWrapper.group(testings: Testing[])`
and then do `await tGroup.start()`

surangap, line 124: how about we add `testingWrapper.group(testings: Testing[])`
and then do `await tGroup.start()`

surangap, line 124: how about we add `testingWrapper.group(testings: Testing[])`
and then do `await tGroup.start()`

chee-chyuan: hey that seems like a good idea!!!!

packages/jellyfish-testing/__tests__/testingcontainer.test.ts
```diff
@@ -0,0 +1,136 @@
...
+    await tGroup.stop()
+  })
+
+  it('should be able to create individual container of masternode and regtest and create a group from it', async () => {
+    const masternodeTesting = testingWrapper.create()
+    await masternodeTesting.container.start()
+    const nonmasternodeTesting = testingWrapper.create(1, (index) => new RegTestContainer()) as Testing
+    await nonmasternodeTesting.container.start()
+
+    const tGroup = testingWrapper.group([masternodeTesting, nonmasternodeTesting])
```
surangap, line 124: test comment again

surangap: problem resolved.

packages/jellyfish-testing/__tests__/testingcontainer.test.ts
```diff
@@ -0,0 +1,136 @@
...
+
+    // call rpc
+    let blockHeight = await testing.rpc.blockchain.getBlockCount()
+    expect(blockHeight).toStrictEqual(0)
+    await testing.generate(1)
+    blockHeight = await testing.rpc.blockchain.getBlockCount()
+    expect(blockHeight).toStrictEqual(1)
+  })
+
+  it('should be able to create regtest non masternode container', async () => {
```
kodemon, line 25: Seeing this functionality, does this PR now also address #965 ?

packages/jellyfish-testing/src/testingwrapper.ts
```diff
@@ -0,0 +1,32 @@
+import { RegTestFoundationKeys } from '@defichain/jellyfish-network'
+import { ContainerGroup, DeFiDContainer, MasterNodeRegTestContainer } from '@defichain/testcontainers'
+import { Testing, TestingGroup } from '.'
+
+export class TestingWrapper {
+  create (): Testing
+  create (n: number): TestingGroup
+  create (n: number, init: (index: number) => DeFiDContainer): Testing | TestingGroup
```
kodemon, line 8:
```suggestion
  create (): Testing
  create (n: 1): Testing
  create (n: number): TestingGroup
  create (n: 1, init: (index: number) => DeFiDContainer): Testing
  create (n: number, init: (index: number) => DeFiDContainer): Testing | TestingGroup
```
If we add case for `1` input we are addressing the case on line 15 allowing us to avoid having to cast to `Testing` in our unit tests.

kodemon (changes requested):

packages/jellyfish-testing/__tests__/testingcontainer.test.ts
```diff
@@ -0,0 +1,136 @@
...
+    testing = testingWrapper.create(1, index => new RegTestContainer()) as Testing
```
kodemon, line 26: I think we can potentially remove `index` in the methods here as they are unused?

packages/jellyfish-testing/src/testing.ts
```diff
@@ -6,24 +6,31 @@ import { TestingRawTx } from './rawtx'
...
+  public readonly misc: TestingMisc
 
   private readonly addresses: Record<string, string> = {}
 
   private constructor (
-    public readonly container: MasterNodeRegTestContainer,
+    public readonly con: DeFiDContainer,
     public readonly rpc: TestingJsonRpcClient
   ) {
+    this.container = con as MasterNodeRegTestContainer
```
surangap, line 28: coming from a cpp background don't like this downcasting. especially given that we might pass instances that are children of `DeFiDContainer`, but not `MasterNodeRegTestContainer`. i.e `RegTestContainer`
In such a case extra properties of the ``MasterNodeRegTestContainer`` is not available for the casted object.
 Maybe we can check the instance type and not define `token, poolpair, ...` that depends on `MasterNodeRegTestContainer` extra properties.

packages/jellyfish-testing/src/testingwrapper.ts
```diff
@@ -0,0 +1,32 @@
...
+
+  create (n?: number, init?: (index: number) => DeFiDContainer): Testing | TestingGroup {
+    if (init === undefined) {
+      init = (index: number) => new MasterNodeRegTestContainer(RegTestFoundationKeys[index])
+    }
```
kodemon, line 13: I think we might be able to hoist this out of the function?

```ts
type InitDeFiContainerFn = (index: number) => DeFiDContainer

function defaultInitDeFiContainer (index: number) {
  return new MasterNodeRegTestContainer(RegTestFoundationKeys[index])
}
```

```ts
create (n?: number, init: InitDeFiContainerFn = defaultInitDeFiContainer): Testing | TestingGroup {
```

packages/jellyfish-testing/src/testingwrapper.ts
chee-chyuan: lovely! thanks. i like this one

packages/jellyfish-testing/__tests__/testingcontainer.test.ts
chee-chyuan: yup you are right

packages/jellyfish-testing/__tests__/testingcontainer.test.ts
chee-chyuan: hmmm. it does feel like it is the same. @surangap any ideas?

packages/jellyfish-testing/src/testingwrapper.ts
chee-chyuan: thanks for this one too

packages/jellyfish-testing/__tests__/testingcontainer.test.ts
surangap: yes, this covers #965 as well.

packages/jellyfish-testing/src/testing.ts
chee-chyuan: this is indeed a 'dirty' method. plus we have the `generate()` which will only work for `MasterNodeRegTestContainer`

packages/jellyfish-testing/src/testingwrapper.ts
```diff
@@ -0,0 +1,34 @@
...
+type InitDeFiContainerFn = (index: number) => DeFiDContainer
+function defaultInitDeFiContainer (index: number): DeFiDContainer {
+  return new MasterNodeRegTestContainer(RegTestFoundationKeys[index])
+}
+export class TestingWrapper {
+  create (): Testing
+  create (n: 1): Testing
+  create (n: number): TestingGroup
+  create (n: 1, init: InitDeFiContainerFn): Testing
+  create (n: number, init: InitDeFiContainerFn): Testing | TestingGroup
```
kodemon, line 14:
```suggestion
  create (n: number, init: InitDeFiContainerFn): TestingGroup
```
Apologies I missed this case, when a number other than `1` _(hopefully no one puts a 0)_ is provided return type would be `TestingGroup` only.

chee-chyuan: thanks you are right. zero is tricky perhaps `0 | 1 `shall do the trick? it shall create at least 1 instance then

packages/jellyfish-testing/src/testing.ts
chee-chyuan: i have resulted in creating `generate` and `generatenewaddress` method in DefiDcontainer. also checking for the instance of `this.container` before initiating` TestToken,...` and the like. it requires me to forcibly assert the variables to be non null which is something that i dont quite like however

packages/jellyfish-testing/src/testingwrapper.ts
kodemon: that makes sense and allows for correct output for both values if someone were to add a `0`

packages/jellyfish-testing/src/testing.ts
kodemon: Here is one way we can handle this, its dirty but it gives a decent amount of intellisense/inference of the typing.

```ts
export class Testing<Container extends DeFiDContainer = DeFiDContainer> {
  public readonly fixture: TestingFixture
  public readonly icxorderbook: TestingICX

  public readonly token!: Container extends MasterNodeRegTestContainer ? TestingToken : undefined
  public readonly poolpair!: Container extends MasterNodeRegTestContainer ? TestingPoolPair : undefined
  public readonly rawtx!: Container extends MasterNodeRegTestContainer ? TestingRawTx : undefined
  public readonly misc!: Container extends MasterNodeRegTestContainer ? TestingMisc : undefined

  private readonly addresses: Record<string, string> = {}

  private constructor (
    public readonly container: Container,
    public readonly rpc: TestingJsonRpcClient
  ) {
    this.fixture = new TestingFixture(this)
    this.icxorderbook = new TestingICX(this)
    if (container instanceof MasterNodeRegTestContainer) {
      this.token = new TestingToken(container, this.rpc) as any
      this.poolpair = new TestingPoolPair(container, this.rpc) as any
      this.rawtx = new TestingRawTx(container, this.rpc) as any
      this.misc = new TestingMisc(container, this.rpc) as any
    }
  }
  
  ...
  
  static create<Container extends DeFiDContainer> (container: Container): Testing<Container> {
    return this.createBase(container)
  }

  static createBase<Container extends DeFiDContainer> (container: Container = new DeFiDContainer('regtest') as Container): Testing<Container> {
    const rpc = new TestingJsonRpcClient(container)
    return new Testing(container, rpc)
  }
```

With this the Testing context is based on the container provided during creation.

inference-2

When passing a non master node container we get typescript feedback that certain functionality is not available.

inference-1

When passing master node all is well.

This should help inform development efforts when we are writing tests. There is a couple of dirty work arounds required, like casting to `any` on the instanceof. And still needing to assert but now we can at least say they assert to undefined if not a MasterNode.

What do you think?

packages/jellyfish-testing/src/testing.ts
```diff
@@ -75,12 +87,17 @@ export class TestingGroup {
...
     const containers: MasterNodeRegTestContainer[] = []
     const testings: Testing[] = []
+
+    if (init === undefined) {
+      init = (index: number) => new MasterNodeRegTestContainer(RegTestFoundationKeys[index])
+    }
+
     for (let i = 0; i < n; i += 1) {
-      const container = init(i)
+      const container = init(i) as MasterNodeRegTestContainer
```
kodemon, line 100: This can get problematic as the `init` method now returns `DeFiDContainer` which may or may not be a `MasterNodeRegTestContainer` which we are casting to at line `#100`. This can lead to incorrect inference when someone is creating a group that does not support the same functionality as a `MasterNode`.

Here is a [reference](https://github.com/DeFiCh/jellyfish/blob/b459b6615b6f040581902253c9055dda9bd9f2b6/packages/jellyfish-testing/src/testing.ts) for when I was running down this rabbit hole.

packages/jellyfish-testing/src/testing.ts
chee-chyuan: i think it is an improvement since it allow intellisense to help

packages/jellyfish-testing/src/testing.ts
chee-chyuan: indeed tricky as i was also balancing between not affecting too much exisitng stuff vs some major refactor. but! let me read the reference that you shared

packages/jellyfish-testing/src/testing.ts
chee-chyuan: `public readonly container: Container` results in some complains that `Argument of type 'DeFiDContainer' is not assignable to parameter of type 'MasterNodeRegTestContainer'.` unless all instances are changed, it seems that this method may not work? or am i missing something?

kodemon: Could you point me to where this complaint is occurring as I do not see it on my end at the moment?

chee-chyuan: my bad i probably did a mistake as im no longer seeing that specific complain. however there is another one that is occuring in the CI now

chee-chyuan: reverting to my orignal solution for now as after tinkering with this method. it seems that there are a couple of things that may have conflicts. 
1. the fixture file is complaining that there is a potential undefined object and trying to workaround it may not be elegant
2. seems like we are not able to rely on intellisense when using the testing wrapper to create the Testing instance

packages/jellyfish-testing/src/testing.ts
chee-chyuan: just pushed a slight edit to remove `as MasterNodeRegTestContainer`

packages/jellyfish-testing/src/testing.ts
kodemon: I think for now getting the wrapper in takes priority and we can refactor the typing in a second round of refactoring perhaps?

chee-chyuan: yea that seems like the low hanging fruit for now without affecting too many files

kodemon (approved):

packages/jellyfish-testing/src/testing.ts
```diff
@@ -6,24 +6,32 @@ import { TestingRawTx } from './rawtx'
...
 export class Testing {
+  public readonly container: MasterNodeRegTestContainer
   public readonly fixture = new TestingFixture(this)
-  public readonly token = new TestingToken(this.container, this.rpc)
-  public readonly poolpair = new TestingPoolPair(this.container, this.rpc)
-  public readonly rawtx = new TestingRawTx(this.container, this.rpc)
+  public readonly token!: TestingToken
+  public readonly poolpair!: TestingPoolPair
+  public readonly rawtx!: TestingRawTx
   public readonly icxorderbook = new TestingICX(this)
```
surangap, line 19: let's move `icxorderbook` and `fixture` also inside the `MasterNodeRegTestContainer` check.
even though it does not take the container as input param, it's already there in `Testing`.
so for non children of `MasterNodeRegTestContainer` will not have these defined. 
But if it wants to call any of the functionality, it has to call it via `testing.rpc.*` or `testing.container.call()`. responsibility of getting successful or not is up to the caller.

surangap (commented):
looks good to me. comment -> https://github.com/DeFiCh/jellyfish/pull/1035/files#r802407705

kodemon (dismissed):

surangap (approved):

packages/jellyfish-testing/src/testingwrapper.ts
```diff
@@ -0,0 +1,34 @@
+import { RegTestFoundationKeys } from '@defichain/jellyfish-network'
+import { ContainerGroup, DeFiDContainer, MasterNodeRegTestContainer } from '@defichain/testcontainers'
+import { Testing, TestingGroup } from '.'
+
+type InitDeFiContainerFn = (index: number) => DeFiDContainer
+function defaultInitDeFiContainer (index: number): DeFiDContainer {
+  return new MasterNodeRegTestContainer(RegTestFoundationKeys[index])
+}
+export class TestingWrapper {
```
fuxingloh, line 9: Perhaps it would make sense to make all the methods static? Since you are not holding on to any state for TestingWrapper? No point init TestingWrapper once?

Either way?
1. `export const TestingWrapper = {}`
2. `static create ()...`

chee-chyuan: > 

yea agreed. but i tried the last time to overload a static function but it doesnt seem to be possible. hence no choice, im forced to use non static

kodemon: @chee-chyuan is this referring to?
issue

chee-chyuan: @kodemon gave me a workaround. will apply that one and push

fuxingloh (approved):

kodemon (approved):

surangap (approved):

packages/jellyfish-testing/src/testing.ts
```diff
@@ -6,24 +6,34 @@ import { TestingRawTx } from './rawtx'
...
 import { JsonRpcClient } from '@defichain/jellyfish-api-jsonrpc'
 
 export class Testing {
-  public readonly fixture = new TestingFixture(this)
-  public readonly token = new TestingToken(this.container, this.rpc)
-  public readonly poolpair = new TestingPoolPair(this.container, this.rpc)
-  public readonly rawtx = new TestingRawTx(this.container, this.rpc)
-  public readonly icxorderbook = new TestingICX(this)
-  public readonly misc = new TestingMisc(this.container, this.rpc)
+  public readonly container: MasterNodeRegTestContainer
```
ivan-zynesis, line 14: Suggestion

Add a parent class for Testing, which allow the `container` to be `RegTestContainer`
Thus when `Testing` standalone must use MNRegTest, `TestingGroup` can have `testings: RegTestTesting[]`

packages/jellyfish-testing/src/testing.ts
```diff
@@ -6,24 +6,34 @@ import { TestingRawTx } from './rawtx'
...
+  public readonly misc!: TestingMisc
 
   private readonly addresses: Record<string, string> = {}
 
   private constructor (
-    public readonly container: MasterNodeRegTestContainer,
+    public readonly con: DeFiDContainer,
     public readonly rpc: TestingJsonRpcClient
   ) {
+    this.container = this.con as MasterNodeRegTestContainer
```
ivan-zynesis, line 28: Suggestion above should avoid this casting, the arg must be masternode

packages/testcontainers/src/containers/DeFiDContainer.ts
```diff
@@ -27,7 +27,7 @@ export interface StartFlags {
 /**
  * DeFiChain defid node managed in docker
  */
-export abstract class DeFiDContainer extends DockerContainer {
+export class DeFiDContainer extends DockerContainer {
```
ivan-zynesis, line 30: My preference will be keep this abstract, and it seems feasible to me, by change all the direct `DeFiDContainer` usage to `RegTestContainer`. Generally speaking everything under testing package should use RegTestContainer instead of DeFiDContainer

ivan-zynesis (changes requested):

packages/testcontainers/src/containers/DeFiDContainer.ts
ivan-zynesis: The default 8554 port made it more error prone.

surangap: change to `RegTestContainer` will bypass `testnet`, `mainnet` containers. But not sure if we are going to write tests for these networks anytime soon or never. have to make a decision.

ivan-zynesis: Yup, IMO not trying to cater some future possibility, so vote for not bothering other network for now. The logic written is loading the hardcoded regtest elliptic pair anyway, and using "jellyfish-testing" for production doesn't sounds right.

packages/jellyfish-testing/src/testing.ts
chee-chyuan: initial attempts on this will result in needing a way to distinguish the returning of `Testing` and `NonMNTesting` it could end by needing to use `as` again. not good in typescript to think of any tricks here

packages/testcontainers/src/containers/DeFiDContainer.ts
chee-chyuan: im on the fence on this one. no idea how the future will look

packages/jellyfish-testing/src/testing.ts
surangap: Been thinking about this. The main problem is that there are a lot of tests(old) that depend on `testing.container`  to be a MNContainer. rather than depending on the `testing` .
But we need to add the support for non MN containers one way or another.
Looks like there isn't an elegant way to handle this without a large refactor.
Keeping this on the dock until we do a proper design. and moving other important PRs forward.

PR #1037, opened 2022-02-07
feat(jellyfish-network): BlockSubsidy calculations

fuxingloh:

> Related to https://github.com/DeFiCh/whale/issues/310 and https://github.com/DeFiCh/whale/pull/661. This PR simplifies the development with a no-index, zero-lookup, instantaneous computation of supply, and block subsidy at a given height.

Inspired by https://github.com/DeFiCh/whale/pull/661, however, this PR introduces a deterministic DFI block subsidy and DFI supply calculation. All calculations within this utility class are in satoshi, test vectors are taken directly from c++ tests.

This class cache all 1252 reductions block subsidies and milestones allowing instantaneous computation. With very little memory footprint, 1252 x 2 BigNumber classes for both `getSupply` and `getBlockSubsidy`.

It introduces 2 public methods:

```ts
/**
 * DFI supply calculation up to a given height are done with the best effort.
 * This does not take DFI burning into account.
 * Fee burning, burn address, round-down burning, loan burning, etc. are all excluded.
 *
 * @param {number} height
 * @return BigNumber supply in satoshi up to given height
 */
getSupply (height: number): BigNumber
```

```ts
/**
 * @param {number} height
 * @return BigNumber total block subsidy in satoshi at the given height
 */
getBlockSubsidy (height: number): BigNumber
```

### DIFF
--- a/packages/jellyfish-network/src/BlockRewardDistribution.ts
+++ b/packages/jellyfish-network/src/BlockRewardDistribution.ts
@@ -1,0 +1,50 @@
+import BigNumber from 'bignumber.js'
+
+export interface BlockRewardDistribution {
+  masternode: number
+  community: number
+  anchor: number
+  liquidity: number
+  loan: number
+  options: number
+  unallocated: number
+}
+
+export const BlockRewardDistributionPercentage: BlockRewardDistribution = {
+  masternode: 3333,
+  community: 491,
+  anchor: 2,
+  liquidity: 2545,
+  loan: 2468,
+  options: 988,
+  unallocated: 173
+}
+
+/**
+ * Get block reward distribution with block base subsidy
+ *
+ * @param {BigNumber} subsidy
+ * @return BlockRewardDistribution
+ */
+export function getBlockRewardDistribution (subsidy: BigNumber): BlockRewardDistribution {
+  return {
+    masternode: calculateReward(subsidy, BlockRewardDistributionPercentage.masternode),
+    community: calculateReward(subsidy, BlockRewardDistributionPercentage.community),
+    anchor: calculateReward(subsidy, BlockRewardDistributionPercentage.anchor),
+    liquidity: calculateReward(subsidy, BlockRewardDistributionPercentage.liquidity),
+    loan: calculateReward(subsidy, BlockRewardDistributionPercentage.loan),
+    options: calculateReward(subsidy, BlockRewardDistributionPercentage.options),
+    unallocated: calculateReward(subsidy, BlockRewardDistributionPercentage.unallocated)
+  }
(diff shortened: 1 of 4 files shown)


packages/jellyfish-network/src/BlockSubsidy.ts
```diff
@@ -0,0 +1,174 @@
...
+}
+
+/**
+ * All calculation are in satoshi.
+ *
+ * This class cache all 1252 reductions block subsidies and milestones allowing instantaneous computation.
+ * With very little memory footprint, 1252 x 2 BigNumber classes for both getSupply and getBlockSubsidy.
+ */
+export class BlockSubsidy {
+  private readonly reductionBlockSubsidies: BigNumber[] = this.computeBlockReductionSubsidies()
```
ivan-zynesis, line 53: Agree with the comment above, not sure how the `this` will work here, and ensure it is not compute on each instantiation.

fuxingloh: Compute once at init, then we can keep that object reference for future computation.

packages/jellyfish-network/src/BlockRewardDistribution.ts
```diff
@@ -0,0 +1,50 @@
...
+  masternode: number
+  community: number
+  anchor: number
+  liquidity: number
+  loan: number
+  options: number
+  unallocated: number
+}
+
+export const BlockRewardDistributionPercentage: BlockRewardDistribution = {
```
ivan-zynesis, line 13: `export const BlockRewardDistributionPercentage: () => BlockRewardDistribution`

perhaps make it slightly easier for future upgrade, let say split percentage change after certain height.

ivan-zynesis (commented):
This is only calculating only single block instead of aggregate, looks good to me, we will see how well it works after whale adopt it.

packages/jellyfish-network/src/BlockSubsidy.ts
ivan-zynesis: But init on same `options` is simply wastage (if not difficult to avoid)

fuxingloh: Slightly confused about this statement, wastage in terms of?

packages/jellyfish-network/src/BlockRewardDistribution.ts
fuxingloh: Hmmm, that's true. 

Perhaps an issue for later? Not sure how to design it now since the change is unclear.

ivan-zynesis (approved):

PR #1051, opened 2022-02-14
feat(jellyfish-api-core): add blockchain getChainTxStats RPC

eli-lim:
/kind feature

Implemented type for RPC: #48
1. blockchain.getChainTxStats()

### DIFF
--- a/packages/jellyfish-api-core/__tests__/category/blockchain/getChainTxStats.test.ts
+++ b/packages/jellyfish-api-core/__tests__/category/blockchain/getChainTxStats.test.ts
@@ -1,5 +1,6 @@
 import { MasterNodeRegTestContainer } from '@defichain/testcontainers'
 import { Testing } from '@defichain/jellyfish-testing'
+import { Block } from '../../../src/category/blockchain'
 
 describe('ChainTxStats', () => {
   const testing = Testing.create(new MasterNodeRegTestContainer())
@@ -16,7 +17,6 @@ describe('ChainTxStats', () => {
     const blockHash = await testing.misc.waitForBlockHash(3)
     const stats = await testing.rpc.blockchain.getChainTxStats(2, blockHash)
 
-    expect(stats.time).toBeGreaterThan(1550000000)
     expect(stats.txcount).toStrictEqual(12)
     expect(stats.window_final_block_hash).toStrictEqual(blockHash)
     expect(stats.window_final_block_height).toStrictEqual(3)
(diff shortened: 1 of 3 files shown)


packages/jellyfish-api-core/src/category/blockchain.ts
```diff
@@ -214,6 +214,16 @@ export class Blockchain {
   async waitForBlockHeight (height: number, timeout: number = 30000): Promise<WaitBlockResult> {
     return await this.client.call('waitforblockheight', [height, timeout], 'number')
   }
+
+  /**
+   * Get statistics about the total number and rate of transactions in the chain.
+   *
+   * @param {number} nBlocks - size of the window in number of blocks
+   * @param {number} blockHash - the hash of the block that ends the window
```
chee-chyuan, line 222: {number} -> {string}

packages/jellyfish-api-core/__tests__/category/blockchain/getChainTxStats.test.ts
```diff
@@ -0,0 +1,90 @@
...
+
+  afterAll(async () => {
+    await testing.container.stop()
+  })
+
+  it('should getChainTxStats for selected interval', async () => {
+    const blockHash = await testing.misc.waitForBlockHash(3)
+    const stats = await testing.rpc.blockchain.getChainTxStats(2, blockHash)
+
+    expect(stats.time).toBeGreaterThan(1550000000)
```
chee-chyuan, line 19: personally i feel like testing it to be greater than a constant to be not very meaningful. how about we declare `const dtBefore = Date.now() / 1000` before we wait and we test against `dtBefore` ?

packages/jellyfish-api-core/__tests__/category/blockchain/getChainTxStats.test.ts
```diff
@@ -0,0 +1,90 @@
...
+    expect(stats.txrate).toStrictEqual(1)
+
+    expect(typeof stats.time).toStrictEqual('number')
+    expect(typeof stats.txcount).toStrictEqual('number')
+    expect(typeof stats.window_final_block_hash).toStrictEqual('string')
+    expect(typeof stats.window_final_block_height).toStrictEqual('number')
+    expect(typeof stats.window_block_count).toStrictEqual('number')
+    expect(typeof stats.window_tx_count).toStrictEqual('number')
+    expect(typeof stats.window_interval).toStrictEqual('number')
+    expect(typeof stats.txrate).toStrictEqual('number')
```
chee-chyuan, line 35: i believe since toStrictEqual was already used above, there is no need to check for the type again

packages/jellyfish-api-core/__tests__/category/blockchain/getChainTxStats.test.ts
```diff
@@ -0,0 +1,90 @@
...
+    await testing.container.stop()
+  })
+
+  it('should getChainTxStats for selected interval', async () => {
+    const blockHash = await testing.misc.waitForBlockHash(3)
+    const stats = await testing.rpc.blockchain.getChainTxStats(2, blockHash)
+
+    expect(stats.time).toBeGreaterThan(1550000000)
+    expect(stats.txcount).toStrictEqual(12)
+    expect(stats.window_final_block_hash.length).toStrictEqual(64)
```
chee-chyuan, line 21: i think we can compare this against `blockHash` since we already have this info?

chee-chyuan:
we can include another negative test for when the blockNumber exceeds the blockNumber of the blockHash
eg blockhash for block 5: abcde

and we try to `getChainTxStats(3, abcde)` which will result in an error

packages/jellyfish-api-core/__tests__/category/blockchain/getChainTxStats.test.ts
eli-lim: I believe depending on Date.now() generally introduces non-determinism / flakiness, so it should be avoided where possible - based on other tests I see this pattern too. Wdyt?

packages/jellyfish-api-core/__tests__/category/blockchain/getChainTxStats.test.ts
eli-lim: You're right! Mind stuck in JS 🤪

packages/jellyfish-api-core/__tests__/category/blockchain/getChainTxStats.test.ts
chee-chyuan: personally i do not see the non-determinism here as `waitForBlockHash` will definitely require some amount of time for the chain to generate 3 blocks

eli-lim: the _some amount of time_ is the non-deterministic part I think. Noobie request 😄 - I wanna know what others think too so that I know what the general strategy is when it comes to checking time fields

canonbrother (commented):
some failed test to test
```ts
getChainTxStats(num, '0') // expect 'blockhash must be of length 64'
getChainTxStats(num, '0',repeat(64)) // expect 'Block not found'
```

packages/jellyfish-api-core/src/category/blockchain.ts
```diff
@@ -214,6 +214,16 @@ export class Blockchain {
   async waitForBlockHeight (height: number, timeout: number = 30000): Promise<WaitBlockResult> {
     return await this.client.call('waitforblockheight', [height, timeout], 'number')
   }
+
+  /**
+   * Get statistics about the total number and rate of transactions in the chain.
+   *
+   * @param {number} nBlocks - size of the window in number of blocks
+   * @param {string} blockHash - the hash of the block that ends the window
```
canonbrother, line 222: we're referring jsdoc syntax (note: not google closure)
https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html

```suggestion
   * @param {number} [nBlocks] - size of the window in number of blocks
   * @param {string} [blockHash] - the hash of the block that ends the window
```

canonbrother: better to have default info as well
https://github.com/DeFiCh/ain/blob/612bc1a52f7ab9239767f52d0ac5bd5928979e35/src/rpc/blockchain.cpp#L1623-L1624

packages/jellyfish-api-core/__tests__/category/blockchain/getChainTxStats.test.ts
canonbrother: i think checking the tip block is more clarify
eg:
```ts
const bestHash = await testing.container.getBestBlockHash()
const bestBlock = await testing.container.call('getblock', [bestHash, 1])
expect(stats.time).toStrictEqual(bestBlock.mediantime)
```

canonbrother (approved):

packages/jellyfish-api-core/__tests__/category/blockchain/getChainTxStats.test.ts
```diff
@@ -0,0 +1,117 @@
...
+    const blockHash = await testing.misc.waitForBlockHash(3)
+    const stats = await testing.rpc.blockchain.getChainTxStats(2, blockHash)
+
+    expect(stats.txcount).toStrictEqual(12)
+    expect(stats.window_final_block_hash).toStrictEqual(blockHash)
+    expect(stats.window_final_block_height).toStrictEqual(3)
+    expect(stats.window_block_count).toStrictEqual(2)
+    expect(stats.window_tx_count).toStrictEqual(2)
+    expect(stats.window_interval).toStrictEqual(2)
+    expect(stats.txrate).toStrictEqual(1)
```
fuxingloh, line 26:
```ts
expect(stats).toStrictEqual({
  txcount: 12,
  ...
})
```

btw this works too and might be cleaner

fuxingloh (approved):

PR #1061, opened 2022-02-15
chore(ocean-api): initial nestjs monorepo setup

kodemon:

A pre-cursor to workflow and shift left adjustments in jellyfish to prepare for consolidation with playground and whale. Also preps for deployment tickets #942 #943


package.json
```diff
@@ -23,7 +23,12 @@
     "all:clean": "rm -rf ./packages/**/dist && rm -rf ./apps/**/dist && rm -rf ./packages/**/tsconfig.build.tsbuildinfo",
     "all:build": "lerna run build",
     "all:version": "lerna version $1 --yes --no-push --no-git-tag-version --exact",
-    "all:publish": "lerna exec --no-private -- npm publish --tag latest --access public"
+    "all:publish": "lerna exec --no-private -- npm publish --tag latest --access public",
+    "apps:ocean": "nest start ocean-api"
```
fuxingloh, line 27: should not have start function within root package, if we need to we can use `npm run start -workspace=ocean-api`.

package.json
```diff
@@ -23,7 +23,12 @@
     "all:clean": "rm -rf ./packages/**/dist && rm -rf ./apps/**/dist && rm -rf ./packages/**/tsconfig.build.tsbuildinfo",
```
fuxingloh, line 23: Largely this file should be untouched. Unless it concern all packages and apps.

nest-cli.json
```diff
@@ -0,0 +1,19 @@
+
+{
+  "collection": "@nestjs/schematics",
+  "monorepo": true,
+  "compilerOptions": {
+    "webpack": false
+  },
+  "projects": {
+    "ocean-api": {
+      "type": "application",
```
fuxingloh, line 10: hmm, I am wondering if we can forgo `nest-cli.json` and just use the barebone setup.

kodemon:
Thanks for the feedback @fuxingloh , I think with the realignment in thinking this PR does not serve a practical purpose. Will move onto the tickets and solve them from the `main` structure. 👍🏻

PR #1076, opened 2022-02-17
chore(ocean-api): temporary disable docker builds

kodemon:

Workflow addition in #1070 is bloating our build times up to 1 hour because of slow arm64 builds. This disables the workflow temporarily until we have a final solution for our docker release and sanity flow to keep our current build processes snappy.

#### Additional comments?:

Will re-enable a version of this with #943

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -62,24 +62,25 @@ jobs:
       - run: npm ci
       - run: npx --no-install eslint .
 
-  docker-buildx:
-    name: Docker Buildx
-    runs-on: ubuntu-latest
-    strategy:
-      matrix:
-        platform: [ linux/amd64, linux/arm64 ]
-        app: [ legacy-api, ocean-api ]
-    steps:
-      - name: Set up QEMU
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -62,24 +62,25 @@ jobs:
...
-    steps:
-      - name: Set up QEMU
-        uses: docker/setup-qemu-action@27d0a4f181a40b142cce983c5393082c365d1480
+  # TODO(kodemon): Re-enable in subsequent docker PRs
+  # docker-buildx:
+  #   name: Docker Buildx
+  #   runs-on: ubuntu-latest
+  #   strategy:
+  #     matrix:
+  #       platform: [ linux/amd64, linux/arm64 ]
```
fuxingloh, line 71:
```suggestion
       platform: [ linux/amd64 ]
```

We should just remove `arm64` build since that's the slow one.

kodemon: That's fair, removed `arm64` 👍🏻

fuxingloh (approved):

Issue #1060, opened 2022-02-15
`/apps/legacy-api` - HealthCheck via Actuator


fuxingloh:
#### What would you like to be added:

`_actuator/probes/...` health check. We can follow the ocean-api design towards how we can implement the health-check. Should also import `apps/ocean-api` files to keep it DRY.

#### Why is this needed:

To take down legacy-api nodes when they end up in a bad state.

/triage accepted
/assign @eli-lim
/area apps
/priority important-soon

fuxingloh:
Following this design: https://github.com/DeFiCh/jellyfish/blob/main/apps/ocean-api/src/controllers/ActuatorController.ts

eli-lim:
Update: simply importing the files from apps/ocean-api will cause nest to fail some dependency look-ups. Trying to find a way to share the `ActuatorModule` between the two (possibly more next time) apps

ivan-zynesis:
We will need something like `jellyfish-api-template`. All of our API going to need this implemented, I am copy pasting all those modules in the new rich-list-api at the moment and leaving them as it is and wait for the solution here.

fuxingloh:
> We will need something like `jellyfish-api-template`. All of our API going to need this implemented, I am copy pasting all those modules in the new rich-list-api at the moment and leaving them as it is and wait for the solution here.

@eli-lim worked on https://github.com/DeFiCh/jellyfish/pull/1078

Issue #1071, opened 2022-02-17
`apps/website` - move website out of apps


kodemon:
#### What would you like to be added:

Move `apps/website` out of `apps/` and into root, or a new folder which befits its scope more clearly.

#### Why is this needed:

In our ongoing consolidation efforts `apps` consists of containers that is to be part of a jellyfish services, and `website` falls into a different category in the upcoming scope of `apps`.

Since website is responsible for documentation and is built outside of this scope via `netlify` it will serve us well to have clear cut concerns between `apps` and `website`. This simplifies our other processes by scoping our structure more correctly making building docker containers for `apps` vs `website` via netlify straight forward without needing to add complex rules to our build processes going forward.

```
apps # Jellyfish services/apps/apis
  legacy-api
  ocean-api
  playground-api
  whale-api
website # Jellyfish Landing/Documentation
```

fuxingloh:
/area core workflow
/triage accepted
/assign @kodemon 

There is also another issue with our current approach that is pretty annoying. More info here: https://docs.netlify.com/configure-builds/common-configurations/monorepos/#declare-dependencies-at-the-root-level-and-within-a-subdirectory

Perhaps something to look into too.

kodemon:
Added follow up ticket to get website move done faster, added follow up #1087 for netlify adjustments

PR #1078, opened 2022-02-17
refactor(apps): Adopt Nest monorepo structure to allow module sharing

eli-lim:
TLDR: Adopt NestJS monorepo structure for legacy-api, ocean-api, rich-list-api and other future apps / services for code sharing

Context: was working on #1060 then dove into Nest's rabbit hole pretty hard. Tried to share a module between legacy-api and ocean-api and ended up here. 😂 Fully aware that this can be throwaway work, but we can refer to it in the future if the need arises.

- Original motivation - sharing the `ActuatorModule` for service health checks. Right now it's being ctrl+c ctrl-v'd everywhere: #1060 
- We can't just import modules from another workspace with its own `node_modules` as Nest will end up confused and throw some strange errors. Hence developing standalone Nest modules without the [Nest monorepo structure](https://docs.nestjs.com/cli/monorepo) is not possible **afaik** (<-- emphasis here).

Greatly reduce duplicate code in the apps

#### Additional comments?:
- To limit the amount of code changed from this 1 PR, let's continue consolidating duplicate code into `libs` in future smaller PRs.
- Jest should always be run from repository root, hence checked in Intellij `.run/all-tests.run.xml` - a template run config for Jest that will make this default.
- **Note**: the build output structure will change, so might have to adjust deployment code / procedure:
    | Apps & dependent packages | Apps expanded |
    |---------|------|
    |image|image|


canonbrother:
> I believe because of how Nest's IOC container works, you can't just import modules from another workspace. Hence developing standalone Nest modules without the [Nest monorepo structure](https://docs.nestjs.com/cli/monorepo) is not possible afaik (<-- emphasis here).

Ordinary nestjs mono structure only 1 node_modules
hence combine the node_modules to one then the dep can be shared to each other

kodemon:
I think as you mentioned in your descriptions we can keep the apis in the apps folder as the website is moving out as part of referenced ticket #1071 

I like the proposed structure here, so for this PR could just do

```ts
apps
  apis|apps|service|...
  lib
  website
  ...nest config
```

for now since website will move away from this folder once I get to the aforementioned ticket.

canonbrother (dismissed):

fuxingloh:
> I think as you mentioned in your description we can keep the apis in the apps folder as the website is moving out as part of referenced ticket #1071
> 
> I like the proposed structure here, so for this PR could just do
> 
> ```
> apps
>   apis|apps|services|...
>     legacy-api
>     ocean-api
>   libs
>     ...shared modules
>   website
>   ...config.files
> ```
> 
> for now since website will move away from this folder once I get to the aforementioned ticket.

Speaking of this PR, maybe we should move the website out first before we merge this in. Let's wait for https://github.com/DeFiCh/jellyfish/issues/1071

apps/tsconfig.json
```diff
@@ -0,0 +1,26 @@
...
+    "experimentalDecorators": true,
+    "outDir": "./dist",
+    "baseUrl": "./",
+    "paths": {
+      "@defichain-app-lib/actuator": [
+        "libs/actuator/src"
+      ],
+      "@defichain-app-lib/actuator/*": [
+        "libs/actuator/src/*"
+      ],
```
kodemon, line 20:
```suggestion
      "@defichain-app-lib/*": [
        "libs/*/src"
      ],
```
Wonder if we want to keep the simplicity of maintaining this file like we do with shared packages. Since this is within the scope of shared packages this space will dynamically grow as we abstract out shared functionality.

tsconfig.json
```diff
@@ -3,6 +3,12 @@
   "compilerOptions": {
     "baseUrl": "./",
     "paths": {
+      "@defichain-app-lib/actuator": [
+        "apps/libs/actuator/src"
+      ],
+      "@defichain-app-lib/actuator/*": [
+        "apps/libs/actuator/src/*"
+      ],
```
kodemon, line 11:
```suggestion
      "@defichain-app-lib/*": [
        "apps/libs/*/src"
      ],
```

eli-lim: image

Because jest is run at the project root, it works with `tsconfig.json` to figure out where this `@defichain-app-lib/actuator` actually is. Hence I don't think removing it here is an option.

apps/tsconfig.json
eli-lim: Yea I fully agree

But removing this causes `build` to fail - typescript needs to know how to resolve the imports
image

I was also trying to figure out how best to avoid duplicating this across `jellyfish/tsconfig.json` and `jellyfish/apps/tsconfig.json`, but couldn't find an elegant way so far 🤔

eli-lim: Woopsies - solution is for jellyfish/apps/tsconfig.json to extend jellyfish/tsconfig.json

eli-lim: Generalised to `'@defichain-app-lib/(.*)': '<rootDir>/apps/libs/$1/src'`
Thanks!

kodemon:
Added pull request for website move #1086

fuxingloh:
FYI conflicts due to #1086 merge

eli-lim:
All good

kodemon (approved):

canonbrother (approved):

lerna.json
```diff
@@ -2,7 +2,8 @@
   "version": "0.0.0",
   "packages": [
     "./",
-    "apps/*",
+    "apps",
+    "apps/website",
```
fuxingloh, line 6: I believe this has been moved out, we can remove it?

apps/package.json
```diff
@@ -0,0 +1,60 @@
...
+  "license": "MIT",
+  "scripts": {
+    "build": "nest build",
+    "start:legacy-api": "nest start legacy-api",
+    "start:ocean-api": "nest start ocean-api",
+    "start:rich-list-api": "nest start rich-list-api",
+    "clean": "rm -rf dist",
+    "start:dev": "nest start --watch",
+    "start:debug": "nest start --debug --watch",
+    "start:prod": "node dist/main"
```
fuxingloh, line 17: these scripts would not be applicable since it's mono?

fuxingloh:
@kodemon I assume this won't break the build process for docker?

.run/all-tests.run.xml
```diff
@@ -0,0 +1,12 @@
+<component name="ProjectRunConfigurationManager">
```
fuxingloh, line 1: Is it possible to have this be placed into the `.idea` directory since it's contextually a feature in IntelliJ?

apps/package.json
ivan-zynesis: Worked for me, I did similar when testing the rich-list-api but do not find it clean, would like to hear better solution.

tsconfig.json
```diff
@@ -3,6 +3,9 @@
   "compilerOptions": {
     "baseUrl": "./",
     "paths": {
+      "@defichain-app-lib/*": [
+        "apps/libs/*/src"
+      ],
```
fuxingloh, line 8:
```suggestion
      "@defichain-apps/libs-*": [
        "apps/libs/*/src"
      ],
```

Probably better to structure the import like this? To keep everything contextually related within the same org prefix.

Also; we are using organization prefix and `@defichain-app-lib` is not reserved while [`@defichain-apps`](https://www.npmjs.com/settings/defichain-apps/packages) is.

apps/package.json
fuxingloh: Since we have `start:*` e.g. `start:legacy-api`, we might not need these?

> Personally, when I develop, I never use `start:*`, I always set up tests to test my hypothesis.

kodemon:
> @kodemon I assume this won't break the build process for docker?

That is a good question, I will take a closer look at it to confirm.

tsconfig.json
eli-lim: Great idea thanks

fuxingloh (approved):

.run/all-tests.run.xml
fuxingloh: I assume this file got ignored by `.gitignore`, might need to `!` to unignore that section

jest.config.js
```diff
@@ -3,7 +3,10 @@ module.exports = {
   testRegex: '((\\.|/)(e2e|test|spec))\\.[jt]sx?$',
   testSequencer: require.resolve('./jest.sequencer'),
   moduleNameMapper: {
-    '@defichain/(?!whale-api-client)(.*)': '<rootDir>/packages/$1/src'
+    '@defichain/(?!whale-api-client)(.*)': '<rootDir>/packages/$1/src',
+
+    // apps
```
ivan-zynesis, line 8:
```suggestion
    // libs
```
Just to be clear non of the `app` component should be importable

ivan-zynesis (approved):

.run/all-tests.run.xml
fuxingloh: For this, open another PR instead after this get merged

eli-lim: Missed that, thanks for catch

canonbrother (approved):

Issue #1053, opened 2022-02-14
`/apps/rich-api` - Providing a Rich List Index


fuxingloh:
#### What would you like to be added:

Providing a Rich List index & endpoint (top 1000) service. This service doesn't need to be live but we need to be able to provide a token-rich list index. 

/triage accepted
/assign @ivan-zynesis 
/area apps
/priority important-soon

ivan-zynesis:
```md
# Jellyfish
## packages
- ### jellyfish-ops-adapter?
  - #### QueueModuleInterface
  - #### PersistentStorageInterface
  - #### RichListServer (with 2 of above injected)
    - create('channel', LIFO)
    - publish('channel', 'msg')
    - read('channel')
    - address watcher -> push address to queue
    - daemon: read from queue + existing -> update new rich list
- ### ocean-api-client
  - #### api/rich-list
    - /GET list

## apps
- ocean-api
  - controllers/rich-list
  - modules
    - QueueModule
    - PersistentStorageModule
    - RichListServerModule

 controller

# Jellyfish-ops?
- QueueModule: implement with SQS
- PersistentStorageModule: even simple text could work</br>
  (if only for rich list storage purpose, eg daemon run at daily/hourly basis only)
```

PR #1080, opened 2022-02-18
feat(`rich-list-api`): initialization

ivan-zynesis:
Narrative and initial app (server) structure for #1053 

Fixes #


kodemon (approved):

fuxingloh:
@kodemon I agree with the above but let's find a way to do https://github.com/DeFiCh/jellyfish/issues/1060#issuecomment-1043823664 before we move on that.

fuxingloh (approved):

PR #1081, opened 2022-02-18
feat(`rich-list-api`): prerequisite module / queue

ivan-zynesis:
Complementary module for #1053 
A queue service module implementation by using TypeOrmModule.
Instead of traditional redis + mq implementation, this reduce infrastructure requirement whenever RDB service already exists thus improve overall horizontal scalability.

Fixes #


chee-chyuan (approved):

apps/rich-list-api/src/modules/QueueModule.ts
```diff
@@ -0,0 +1,82 @@
...
+    const consumed = await this.repo.find({
+      where: {
+        queueName: this.queueName
+      },
+      take: maxMessage,
+      order: {
+        timestamp: this.mode === 'FIFO' ? 'ASC' : 'DESC'
+      }
+    })
+    await this.repo.delete({ id: In(consumed.map(i => i.id)) })
```
kodemon, line 53: Wanted to ask a quick question considering the case of horizontal scalability. Are these two operations part of a single `transaction`? If two or more nodes are pulling from a queue before the delete operation we will have multiple nodes processing the same items of the queue.

These considerations may be moot, and I don't have enough background knowledge to determine the strictness of the queue. But just wanted to drop a quick question on this in case it could be an actual issue.

kodemon (approved):
Had a comment for some background knowledge on queue requirements, otherwise PR looks good to me.

canonbrother (approved):

fuxingloh (approved):

PR #1084, opened 2022-02-21
feat(`rich-list-api`): rich-list-core

ivan-zynesis:
2 supplementary sub packages for #1053 

Fixes #


packages/rich-list-core/src/controller/AddressParser/dftx/accountToUtxos.ts
```diff
@@ -0,0 +1,16 @@
+import { fromScript } from '@defichain/jellyfish-address'
+import { AccountToUtxos, CAccountToUtxos, DfTx } from '@defichain/jellyfish-transaction'
+import { ActiveAddress } from '../ActiveAddress'
+import { DfTxAddressParser } from './_abstract'
+
+export class AccountToUtxosParser extends DfTxAddressParser<AccountToUtxos> {
```
ivan-zynesis, line 6: Checked the rpc behavior, cannot run away from this to extract all involved addresses. DfTx scriptPubKey will not include `addresses` field. This will become a long term effort to maintain rich list accuracy. On the optimistic side, this is not as complicated as calculating the dftx token output.

packages/rich-list-core/src/RichListCore.ts
```diff
@@ -0,0 +1,179 @@
...
+import { AddressParser } from './controller/AddressParser'
+import { NetworkName } from '@defichain/jellyfish-network'
+import { RichListItem } from '@defichain/rich-list-api-client'
+import { AccountAmount } from 'packages/jellyfish-api-core/src/category/account'
+import { ActiveAddressAccountAmount } from './controller/AddressParser/ActiveAddressAccountAmount'
+import { PersistentLinkedList } from './lib/PersistentLinkedList'
+
+export class RichListCore {
+  CATCHING_UP = false
+  RICH_LIST_LENGTH = 1000
```
kodemon, line 13: Quick question on variable naming here, usually all uppercase variables usually refers to a constant unchangeable assignment. Usually when I see variables in this format I think constants, thoughts?

ivan-zynesis: It started to be constant and I made it configurable later. Agree it should be in camel case.

packages/rich-list-core/src/RichListCore.ts
```diff
@@ -0,0 +1,179 @@
...
+
+  resume (): void {
+    if (this.CATCHING_UP) {
+      return
+    }
+    this.CATCHING_UP = true
+    void this._catchUp()
+  }
+
+  async _catchUp (): Promise<void> {
```
kodemon, line 40: Are `_` prefixed methods meant to represent `private`, in which case perhaps we should use the `private` class assignment for these methods to be more explicit?

kodemon: Actually looking through the file I see some `_` with preceding `private` assignment. What does the `_` represent in this class?

ivan-zynesis: All they `_` prefixed are meant to be private. But during the implementation I keep changing my mind for exposing them for functional test, ended up the inconsistency here. 

Will make them private first, more likely to have e2e instead of actually testing each single function.

packages/rich-list-core/src/RichListCore.ts
```diff
@@ -0,0 +1,174 @@
...
+    } else {
+      const queue = await this.queueClient.createQueueIfNotExist(this._queueName(), 'LIFO')
+
+      const addresses: string[] = []
+      for (const tx of nextBlock.tx) {
+        const addresses = await this.addressParser.parse(tx)
+        for (const a of addresses) {
+          addresses.push(a)
+          await queue.push(a)
+        }
```
canonbrother, line 61:
```suggestion
      const addresses: string[] = []
      for (const tx of nextBlock.tx) {
        const _addresses = await this.addressParser.parse(tx)
        for (const a of _addresses) {
          addresses.push(a)
          await queue.push(a)
        }
```
some readability imprv... it doubts me on ln:65

ivan-zynesis: Nice catch, I didn't even know our linter allow shadowed naming

packages/rich-list-core/src/RichListCore.ts
```diff
@@ -0,0 +1,174 @@
...
+  // private async _invalidate () {
+  //   /**
+  //    * TODO:
+  //    * 1. remove highest block data
+  //    * 2. get back dropped address
+  //    * 3. sort rich list again
+  //    */
+  // }
+
+  async get (token: string): Promise<RichListItem[]> {
```
canonbrother, line 93:
```suggestion
  async get (tokenId: string): Promise<RichListItem[]> {
```
better

packages/rich-list-core/src/lib/PersistentLinkedList.ts
```diff
@@ -0,0 +1,29 @@
...
+  }
+
+  async getLast (): Promise<T | undefined> {
+    return this.DATA[this.DATA.length - 1]
+  }
+
+  async size (): Promise<number> {
+    return this.DATA.length
+  }
+}
```
kodemon, line 29:
```suggestion
export class InMemoryLinkedListService<T> implements PersistentLinkedList<T> {
  private readonly data: T[] = []

  async append (data: T): Promise<void> {
    this.data.push(data)
  }

  async removeLast (): Promise<void> {
    this.data.pop()
  }

  async getLast (): Promise<T | undefined> {
    return this.data[this.data.length - 1]
  }

  async size (): Promise<number> {
    return this.data.length
  }
}
```
Based on previous insights I think this can be renamed to lowercased `data`

packages/rich-list-core/src/RichListCore.ts
```diff
@@ -0,0 +1,174 @@
...
+  }
+
+  /**
+   * Updated rich list with latest balance
+   * by sorting the existing rich list together with `queuedAddressLimit` number of recently active addresses
+   *
+   * @param queuedAddressLimit [5000]
+   */
+  async calculateNext (queuedAddressLimit = 5000): Promise<void> {
+    const tokens = await this._listTokens()
```
canonbrother, line 112: the `token` is a object for while i'm reading.. suggest to suffix `id` if its a tokenId
```suggestion
    const tokenIds = await this._listTokenIds()
```

packages/rich-list-core/src/lib/PersistentStorage.ts
```diff
@@ -0,0 +1,27 @@
...
+
+  async put (key: K, value: V): Promise<void> {
+    this.DATA[this.serializeKey(key)] = value
+  }
+
+  async delete (key: K): Promise<void> {
+    // eslint-disable-next-line @typescript-eslint/no-dynamic-delete
+    delete this.DATA[this.serializeKey(key)]
+  }
+}
```
kodemon, line 27:
```suggestion
export abstract class InMemoryStorageService<K, V> implements PersistentStorage<K, V> {
  private readonly data: { [key: string]: V } = {}

  abstract serializeKey (k: K): string

  async get (key: K): Promise<V | undefined> {
    return this.data[this.serializeKey(key)]
  }

  async put (key: K, value: V): Promise<void> {
    this.data[this.serializeKey(key)] = value
  }

  async delete (key: K): Promise<void> {
    // eslint-disable-next-line @typescript-eslint/no-dynamic-delete
    delete this.data[this.serializeKey(key)]
  }
}
```

packages/rich-list-core/src/lib/Queue.ts
```diff
@@ -0,0 +1,45 @@
...
+
+  getQueue (name: string): Queue<T> | undefined {
+    return this.DATA[name]
+  }
+
+  async createQueueIfNotExist (name: string, mode: Mode): Promise<Queue<T>> {
+    this.DATA[name] = new InMemoryQueueService()
+    return this.DATA[name]
+  }
+}
```
kodemon, line 45:
```suggestion
export class InMemoryQueueService<T> implements Queue<T> {
  private data: T[] = []

  async push (message: T): Promise<string> {
    this.data.push(message)
    return `${this.data.length - 1}`
  }

  async receive (maxMessage: number): Promise<T[]> {
    const startIndex = Math.min(this.data.length - maxMessage, 0)
    const consumed = this.data.slice(startIndex)
    this.data= this.data.slice(0, startIndex)
    return consumed
  }
}

/**
 * For unit test use.
 */
export class StubbedQueueClient<T> implements QueueClient<T> {
  private readonly data: { [key: string]: Queue<T> } = {}

  getQueue (name: string): Queue<T> | undefined {
    return this.data[name]
  }

  async createQueueIfNotExist (name: string, mode: Mode): Promise<Queue<T>> {
    this.data[name] = new InMemoryQueueService()
    return this.data[name]
  }
}
```

packages/rich-list-core/src/RichListCore.ts
```diff
@@ -0,0 +1,174 @@
...
+        }
+      }
+      await this.crawledBlocks.append({
+        hash: nextBlock.hash,
+        richListDroppedOut: addresses
+      })
+    }
+
+    return await this._catchUp()
+  }
```
kodemon, line 70:
```suggestion
      await this.crawledBlocks.append({
        hash: nextBlock.hash,
        richListDroppedOut: await this._getAddresses(nextBlock)
      })
    }

    return await this._catchUp()
  }
  
  private async _getAddresses(block: defid.Block<defid.Transaction>): Promise<string[]> {
    const queue = await this.queueClient.createQueueIfNotExist(this._queueName(), 'LIFO')
    const addresses: string[] = []
    for (const tx of block.tx) {
      const _addresses = await this.addressParser.parse(tx)
      for (const address of _addresses) {
        addresses.push(address)
        await queue.push(address)
      }
    }
    return addresses
  }
```
This one is more of a personal observation, looking at the overall method growing in size we may want to try to cut down the size and move some logic out into a separate method. So this is mostly just a personal suggestion than a blocker.

canonbrother (commented):
maybe move the address parser to `/parser/AddressParser`?

overall look goods to me
hmm... need to read the test `RichListApiClient.test.ts` execution...

packages/rich-list-core/src/RichListCore.ts
ivan-zynesis: Adopt the suggestion, also fixed a dumb mistake, the `droppedOutAddress` was a silly mistake, they are only computed after compared with previous rich list. 😓

packages/rich-list-core/src/lib/PersistentLinkedList.ts
ivan-zynesis: It is meant to be readonly, since no re-write needed. And this is a stubbed service for test use, I guess it is fine.

packages/rich-list-core/src/RichListCore.ts
```diff
@@ -0,0 +1,175 @@
...
+
+  resume (): void {
+    if (this.isCatchingUp) {
+      return
+    }
+    this.isCatchingUp = true
+    void this._catchUp()
+  }
+
+  async _catchUp (): Promise<void> {
```
kodemon, line 41:
```suggestion
  private async _catchUp (): Promise<void> {
```

ivan-zynesis:
> maybe move the address parser to `/parser/AddressParser`?
> 
> overall look goods to me hmm... need to read the test `RichListApiClient.test.ts` execution...

Those e2e won't be done until most of the pieces are in. TBH the e2e will extremely complicated, to setup the test env/fixture and handle all type of history
- utxo
- dftx
- rewards
- auctions result
- if anymore unaware no txid data

packages/rich-list-core/src/controller/AddressParser/dftx/accountToUtxos.ts
```diff
@@ -0,0 +1,12 @@
+import { fromScript } from '@defichain/jellyfish-address'
+import { AccountToUtxos, CAccountToUtxos, DfTx } from '@defichain/jellyfish-transaction'
+import { DfTxAddressParser } from './_abstract'
+
+export class AccountToUtxosParser extends DfTxAddressParser<AccountToUtxos> {
```
cwkang1998, line 5: Regarding the naming for this file, `accountToUtxos.ts` is used. But the test file for this is named as `AccountToUtxos.test.ts`. Should we change the name for this to `AccountToUtxos.ts` instead?

cwkang1998 (approved):

packages/rich-list-core/src/RichListCore.ts
```diff
@@ -0,0 +1,219 @@
...
+    if (this.isCatchingUp) {
+      return
+    }
+    this.isCatchingUp = true
+    void this._catchUp()
+  }
+
+  private async _catchUp (): Promise<void> {
+    const lastBlock = await this._getCrawledTip()
+    const nextBlockHeight = lastBlock?.data.height ?? 0
```
canonbrother, line 42: hmm.. looking for nextblock based on lastblock.. isn't it `nextblockheight = lastblockheight + 1`??

packages/rich-list-core/src/lib/Queue.ts
```diff
@@ -0,0 +1,45 @@
...
+/**
+ * For unit test use.
+ */
+export class InMemoryQueueService<T> implements Queue<T> {
+  private DATA: T[] = []
+
+  async push (message: T): Promise<string> {
+    this.DATA.push(message)
+    return `${this.DATA.length - 1}`
+  }
```
canonbrother, line 21:
```suggestion
  async push (message: T): number {
    this.DATA.push(message)
    return this.DATA.length - 1
  }
```

packages/rich-list-core/src/RichListCore.ts
ivan-zynesis: Yes you're right, it is a mistake. The previous implementation using a `LinkedList` like interface where `size()` is equivalent to previous height + 1. I should add +1 here.

packages/rich-list-core/__tests__/RichListCore.test.ts
```diff
@@ -0,0 +1,34 @@
...
+  })
+
+  describe('startOrResume - should start crawl blocks and push active addresses', () => {
+    it('should extract all addresses from transactions and push into queue', async () => {
+      richListCore.start()
+      await waitForCatchingUp(richListCore)
+
+      const queue = await richListCore.queueClient.createQueueIfNotExist('RichListCore_ACTIVE_ADDRESSES', 'LIFO')
+      const queuedItem = await queue.receive(Number.MAX_SAFE_INTEGER)
+      console.log(queuedItem)
```
cwkang1998, line 31: I think you meant to `expect` something here

packages/rich-list-core/src/lib/SingleIndexDb.ts
```diff
@@ -0,0 +1,72 @@
...
+  delete: (id: string) => Promise<void>
+  get: (id: string) => Promise<Schema<T> | undefined>
+  list: (filter: FindOptions) => Promise<Array<Schema<T>>>
+}
+
+/**
+ * For unit test use.
+ */
+export class InMemoryDatabase<T> implements SingleIndexDb<T> {
+  private data: { [key: string]: Array<Schema<T>> } = {}
```
cwkang1998, line 27: Thoughts about using the syntatic sugar `Record<string, Array<Schema<T>>>` for representing this? Curious about your ideas.

packages/rich-list-core/src/RichListCore.ts
```diff
@@ -0,0 +1,226 @@
...
+
+  private async _invalidateDroppedOutRichList (tokenId: string, height: number): Promise<void> {
+    const lastBlockRichList = await this._getDroppedOutRichList(tokenId, height)
+    await Promise.all(
+      lastBlockRichList.map(async rl => await this.droppedFromRichList.delete(rl.id))
+    )
+  }
+
+  private async _getDroppedOutRichList (tokenId: string, height: number): Promise<Array<Schema<string>>> {
+    return await this.droppedFromRichList.list({
```
cwkang1998, line 204: Should be equivalent to
```suggestion
    return this.droppedFromRichList.list({
```

packages/rich-list-core/src/RichListCore.ts
```diff
@@ -0,0 +1,226 @@
...
+      order: 'DESC',
+      limit: 1
+    })
+    return lastBlock
+  }
+
+  private async _invalidateDroppedOutRichList (tokenId: string, height: number): Promise<void> {
+    const lastBlockRichList = await this._getDroppedOutRichList(tokenId, height)
+    await Promise.all(
+      lastBlockRichList.map(async rl => await this.droppedFromRichList.delete(rl.id))
```
cwkang1998, line 199: Since you are using `Promise.all`, should be equivalent to 
```suggestion
      lastBlockRichList.map(rl => this.droppedFromRichList.delete(rl.id))
```

packages/rich-list-core/src/RichListCore.ts
```diff
@@ -0,0 +1,219 @@
...
+      const addresses = await this.addressParser.parse(tx)
+      _addresses.push(...addresses)
+    }
+    return _addresses
+  }
+
+  private async _getBlock (height: number): Promise<defid.Block<defid.Transaction> | undefined> {
+    try {
+      const bh = await this.apiClient.blockchain.getBlockHash(height)
+      return await this.apiClient.blockchain.getBlock(bh, 2)
```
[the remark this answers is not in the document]
ivan-zynesis: image

We have some linter rules in place since inherited since project beginning. I think either syntax has pros and cons, for I prefer still to return with await now.

packages/rich-list-core/src/lib/SingleIndexDb.ts
```diff
@@ -0,0 +1,64 @@
...
+      const index = partition.findIndex(d => d.id === id)
+      partition.splice(index, 1)
+    }
+  }
+
+  async get (id: string): Promise<Schema<T> | undefined> {
+    for (const partition of Object.values(this.data)) {
+      const found = partition.find(d => d.id === id)
+      if (found !== undefined) {
+        return found
```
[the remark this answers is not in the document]
ivan-zynesis: Nope, it need to search other "partition", only break if found or undefined until all iterated.

packages/rich-list-core/src/RichListCore.ts
kodemon: We are currently employing explicit `await` in Jellyfish so the above version is expected.

packages/rich-list-core/src/RichListCore.ts
kodemon: This may be another case of explicit requirements, can you confirm this case @fuxingloh ?

packages/rich-list-core/src/lib/Queue.ts
ivan-zynesis: Yup, it should be a promise, grabbing from an array is solely this test implementation.
And the returned handle to be string is more generic and conventional queue service like SQS will have that.

packages/rich-list-core/__tests__/RichListCore.test.ts
ivan-zynesis: My bad, the tests are just added, I should put this back to `Draft` first.
I do not like the `queueClient` being exposed, but consider this is private package, most like will do it for test purpose.

packages/rich-list-core/src/lib/Queue.ts
canonbrother: ohh.. missed that.. okay.. because it was used for in-memory ops.. so confuse the `async` here

packages/rich-list-core/src/RichListCore.ts
ivan-zynesis: Yup, my linter do complain after remove it which I actually prefer. It clearly differentiate on first glance.

cwkang1998: After pulling down the code, it seems that the eslint rule in place prefers having explicit async await. As such I think this is okay.

packages/rich-list-core/src/lib/SingleIndexDb.ts
ivan-zynesis: Practically no difference, but I do agree using `Record` will at least have better readability.

packages/rich-list-core/test/RichListCoreTest.ts
```diff
@@ -0,0 +1,20 @@
+import { JsonRpcClient } from '@defichain/jellyfish-api-jsonrpc'
+import { RichListItem } from '@defichain/rich-list-api-client'
+import { waitForCondition } from '@defichain/testcontainers'
+import { InMemoryDatabase, StubbedQueueClient } from '../src/lib'
+import { CrawledBlock, RichListCore } from '../src/RichListCore'
+
+export function RichListCoreTest (apiClient: JsonRpcClient): RichListCore {
```
canonbrother, line 7: maybe this can move to RichListCore static create.. :sweat_smile:

canonbrother (dismissed):

cwkang1998 (dismissed):

packages/rich-list-core/test/RichListCoreTest.ts
ivan-zynesis: Thought about it, but the file is (and going to be) quite huge. So decide to make another pure functional constructor

cwkang1998 (approved):

packages/rich-list-core/README.md
```diff
@@ -0,0 +1,9 @@
+# @defichain/rich-list-core
+Complementary library module for rich-list-api to compute and keep a rich list state in injected storage service.
+
+## Prerequisite
+Three dependencies to be injected
```
cwkang1998, line 5: I think you meant two?

packages/rich-list-core/src/RichListCore.ts
```diff
@@ -0,0 +1,232 @@
...
+      partition: tokenId,
+      order: 'DESC',
+      gt: height - 1,
+      lt: height + 1
+    })
+  }
+
+  private async _listTokenIds (): Promise<number[]> {
+    const tokens = await this.apiClient.token.listTokens()
+    return Object.keys(tokens).map(id => Number(id)).concat([-1])
```
cwkang1998, line 221: Regarding this, why are we concatenating -1? I don't quite understand this implementation. What does a token id of -1 indicate?

cwkang1998 (changes requested):

packages/rich-list-api-client/package.json
```diff
@@ -0,0 +1,23 @@
...
+  "scripts": {
+    "build": "tsc -b ./tsconfig.build.json"
+  },
+  "dependencies": {
+    "@defichain/ocean-api-client": "0.0.0"
+  },
+  "peerDependencies": {
+    "defichain": "^0.0.0"
+  }
+}
```
fuxingloh, line 23: Best to separate everything in `rich-list-api-client` as a barebone PR, easier to get them merged in iteratively.

packages/rich-list-core/README.md
ivan-zynesis: Yeah, didn't change it after change the lib components 😅

packages/rich-list-core/src/RichListCore.ts
ivan-zynesis: I am using `-1` to represent utxo now, depend if we want to combine it with DFI `0`

packages/rich-list-api-client/package.json
ivan-zynesis: Wanted to use the response interface from `api-client`. But literally only one, copied over and trim the package.

cwkang1998 (dismissed):

cwkang1998 (dismissed):

cwkang1998 (dismissed):

ivan-zynesis:
Need @fuxingloh 's review, blocking the rich-list app's PR.

fuxingloh:
There is a mix of different case sensitivity in this PR, not sure if you have seen this. But we are actively trying to move to standard to `PascalCase`

https://github.com/DeFiCh/jellyfish/blob/main/CONTRIBUTING.md#use-pascalcase-and-period-not-underscores-or-dashes-in-filenames

ivan-zynesis:
> There is a mix of different case sensitivity in this PR, not sure if you have seen this. But we are actively trying to move to standard to `PascalCase`
> 
> https://github.com/DeFiCh/jellyfish/blob/main/CONTRIBUTING.md#use-pascalcase-and-period-not-underscores-or-dashes-in-filenames

Will check through again

cwkang1998 (approved):

packages/rich-list-core/src/lib/Queue.ts
```diff
@@ -0,0 +1,51 @@
...
+  receive: (maxMessage: number) => Promise<T[]>
+}
+
+export interface QueueClient<T> {
+  createQueueIfNotExist: (name: string, mode: Mode) => Promise<Queue<T>>
+}
+
+/**
+ * For unit test use.
+ */
```
fuxingloh, line 14: Since all these are for testing purposes, may be better to put them in `/test` directory for better contextual relevance?

packages/rich-list-core/src/RichListCore.ts
```diff
@@ -0,0 +1,231 @@
...
+  private async _getCrawledTip (): Promise<Schema<CrawledBlock> | undefined> {
+    const [lastBlock] = await this.crawledBlockHashes.list({
+      partition: 'NONE',
+      order: 'DESC',
+      limit: 1
+    })
+    return lastBlock
+  }
+
+  private async _invalidateDroppedOutRichList (tokenId: string, height: number): Promise<void> {
```
fuxingloh, line 201:
```suggestion
  private async invalidateDroppedOutRichList (tokenId: string, height: number): Promise<void> {
```

In the world of typescript, I think the use of `_` is redundant since you already declare it's private. What do you think?

packages/rich-list-core/__tests__/controller/AddressParser/dftx/accountToUtxos.test.ts
```diff
@@ -0,0 +1,51 @@
...
+
+  beforeAll(async () => {
+    await container.start()
+    await container.waitForWalletCoinbaseMaturity()
+    apiClient = new JsonRpcClient(await container.getCachedRpcUrl())
+
+    sender = await container.getNewAddress()
+    rec1 = await container.getNewAddress()
+    rec2 = await container.getNewAddress()
+    rec3 = await container.getNewAddress()
```
fuxingloh, line 24: https://github.com/DeFiCh/jellyfish/blob/a76864a9dbf966e688827d9df22d48e531b7ce2a/packages/jellyfish-testing/src/testing.ts#L33

Awhile back this feature was introduced in `jellyfish-testing` to provide the ability to use named address for testing.

packages/rich-list-core/src/controller/AddressParser/index.ts
```diff
@@ -0,0 +1,58 @@
+import { OP_DEFI_TX, OPCode, DfTx, toOPCodes } from '@defichain/jellyfish-transaction'
+import { ApiClient, blockchain as defid } from '@defichain/jellyfish-api-core'
+import { SmartBuffer } from 'smart-buffer'
+import { DfTxAddressParser } from './dftx/_abstract'
+import { NetworkName } from '@defichain/jellyfish-network'
+import { AccountToUtxosParser } from './dftx/AccountToUtxos'
+import { UtxoAddressParser } from './utxo'
+
```
fuxingloh, line 8: Technically they can be called "saga" rather than "controller" for a more accurate description of what they do.

packages/rich-list-core/src/controller/AddressParser/utxo.ts
```diff
@@ -0,0 +1,30 @@
+import { ApiClient, blockchain as defid } from '@defichain/jellyfish-api-core'
+
```
fuxingloh, line 2: Rename to `UtxoAddressParser.ts`?

packages/rich-list-core/src/RichListCore.ts
```diff
@@ -0,0 +1,231 @@
...
+const DEFAULT_RICH_LIST_LENGTH = 1000
+
+export class RichListCore {
+  isCatchingUp = false
+  richListLength = DEFAULT_RICH_LIST_LENGTH
+  readonly addressParser: AddressParser
+
+  constructor (
+    private readonly network: NetworkName,
+    private readonly apiClient: ApiClient,
```
fuxingloh, line 19: Hmmm, it's much more scalable to expose a WhaleClient than a RpcClient. RpcClient we need to replicate an Ocean like design for the `rich-list-api` service with a snapshot provider. It might not be easy. Maybe we can test with WhaleApiClient instead?

fuxingloh (changes requested):

packages/rich-list-core/src/RichListCore.ts
ivan-zynesis: I have no preference.

packages/rich-list-core/test/StubbedWhaleClient.ts
```diff
@@ -0,0 +1,27 @@
...
+ */
+export class StubbedWhaleApiClient extends WhaleApiClient {
+  constructor () {
+    super({ url: 'stubbed' })
+  }
+
+  async requestAsString (method: Method, path: string, body?: string): Promise<ResponseAsString> {
+    const pathComponent = path.split('/')
+
+    if (method === 'GET' && pathComponent[0] === 'address' && pathComponent[2] === 'balance') {
```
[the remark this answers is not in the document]
ivan-zynesis: Change to stub every expected to be call method instead of this

packages/rich-list-core/src/types.ts
```diff
@@ -1,10 +1,9 @@
...
-  [key: string]: AccountAmount
+export interface AccountAmount {
+  [id: string]: BigNumber
 }
 
-export interface AddressBalance {
-  address: string
-  amount: number
+export interface ActiveAddressAccountAmount {
+  [key: string]: AccountAmount
```
cwkang1998, line 8: I think we can move all this into `RichListCore.ts` since its the only file that uses these data structure.

cwkang1998, line 8: I think we can move all this into `RichListCore.ts` since its the only file that uses these data structure.

packages/rich-list-core/src/RichListCore.ts
```diff
@@ -1,29 +1,34 @@
...
-import { QueueClient, Queue } from './lib/Queue'
-import { SingleIndexDb, Schema } from './lib/SingleIndexDb'
-import { AddressParser } from './controller/AddressParser'
+import { BigNumber, blockchain as defid } from '@defichain/jellyfish-api-core'
+import { WhaleApiClient, WhaleRpcClient } from '@defichain/whale-api-client'
+import { QueueClient, Queue } from './persistent/Queue'
+import { SingleIndexDb, Schema } from './persistent/SingleIndexDb'
+import { AddressParser } from './saga/AddressParser'
 import { NetworkName } from '@defichain/jellyfish-network'
 import { AccountAmount } from '@defichain/jellyfish-api-core/src/category/account'
```
cwkang1998, line 7: I think you meant to import AccountAmount from `./types`

cwkang1998, line 7: I think you meant to import AccountAmount from `./types`

cwkang1998 (approved):

cwkang1998 (changes requested):

packages/rich-list-core/src/RichListCore.ts
```diff
@@ -0,0 +1,236 @@
+import { BigNumber, blockchain as defid } from '@defichain/jellyfish-api-core'
+import { WhaleApiClient, WhaleRpcClient } from '@defichain/whale-api-client'
+import { QueueClient, Queue } from './persistent/Queue'
+import { SingleIndexDb, Schema } from './persistent/SingleIndexDb'
+import { AddressParser } from './saga/AddressParser'
+import { NetworkName } from '@defichain/jellyfish-network'
+import { AccountAmount } from '@defichain/jellyfish-api-core/src/category/account'
```
cwkang1998, line 7: I think you meant to import `AccountAmount` from `./types`

packages/rich-list-core/src/types.ts
ivan-zynesis: Renamed the type to be less confusing.

The logic was correct, it is for the rpc response.

cwkang1998 (approved):

packages/rich-list-core/src/RichListCore.ts
```diff
@@ -0,0 +1,236 @@
...
+  ) {
+    this.addressParser = new AddressParser(whaleRpcClient, network)
+  }
+
+  setRichListLength (length: number): void {
+    this.richListLength = length
+  }
+
+  /**
+   * Expected to be called on application bootstarapped.
```
eli-lim, line 39:
```suggestion
   * Expected to be called on application bootstrapped.
```

packages/rich-list-core/__tests__/RichListCore.test.ts
```diff
@@ -0,0 +1,111 @@
...
+        lt: new BigNumber(richList[2].amount).times('1e8').dividedToIntegerBy(1).toNumber()
+      })
+      expect(excludedFromTopThree.length).toStrictEqual(EXPECTED_RICH_LIST_ADDRESSES.length - 3)
+
+      for (let i = 0; i < excludedFromTopThree.length; i++) {
+        expect(excludedFromTopThree[i].data.amount).toBeLessThanOrEqual(richList[2].amount)
+      }
+    })
+  })
+})
```
eli-lim, line 111: Probably can add a test for block invalidation + rich list updating logic? Or can do in a follow-up PR

I tried the following, but the balance doesn't get updated when fundAddress is called. I might be missing something / misunderstanding.
```suggestion
})


  describe('invalidation', () => {
    it('should update balances when block is invalidated', async () => {
      // Given some addresses having balances
      richListCore.start()
      await waitForCatchingUp(richListCore)
      await richListCore.calculateNext()

      // When one of the rich gets a lot of DFI
      await container.fundAddress('2NCWAKfEehP3qibkLKYQjXaWMK23k4EDMVS', 9_000_000)
      await container.generate(1)

      // And the address queue is processed
      await waitForCatchingUp(richListCore)
      await richListCore.calculateNext()
      const queue = await richListCore.queueClient.createQueueIfNotExist('RichListCore_ACTIVE_ADDRESSES', 'LIFO')
      const anythingRemainedInQ = await queue.receive(Number.MAX_SAFE_INTEGER)
      expect(anythingRemainedInQ.length).toStrictEqual(0)

      // Then rich list is updated
      const richList = await richListCore.get(UTXO)
      expect(richList.length).toStrictEqual(EXPECTED_RICH_LIST_ADDRESSES.length)
      expectCorrectRichListOrderedByBalance(richList)

      // When block is invalidated
      // ...
      // Then rich list is updated again
    })
  })
})

function expectCorrectRichListOrderedByBalance(richList: AddressBalance[]) {
  for (let i = 0; i < richList.length - 1; i++) {
    const current = richList[i]
    const next = richList[i + 1]
    expect(EXPECTED_RICH_LIST_ADDRESSES).toContain(current.address)
    expect(EXPECTED_RICH_LIST_ADDRESSES).toContain(next.address)
    expect(current.amount).toBeGreaterThanOrEqual(next.amount)
  }
}
```

eli-lim (commented):
Some minor stuff, and preferably add a test for invalidation logic (unless that's going to be tackled in a follow-up PR)

packages/rich-list-core/package.json
```diff
@@ -0,0 +1,36 @@
...
+  "main": "dist/index.js",
+  "types": "dist/index.d.ts",
+  "files": [
+    "dist"
+  ],
+  "scripts": {
+    "build": "tsc -b ./tsconfig.build.json"
+  },
+  "dependencies": {
+    "@defichain/jellyfish-api-core": "0.0.0",
```
eli-lim, line 25:
```suggestion
    "@defichain/jellyfish-api-core": "^0.0.0",
```

packages/rich-list-core/package.json
```diff
@@ -0,0 +1,36 @@
...
+    "@defichain/jellyfish-transaction": "0.0.0",
+    "@defichain/whale-api-client": "0.0.0",
```
eli-lim, line 27:
```suggestion
    "@defichain/jellyfish-api-core": "^0.0.0",
    "@defichain/jellyfish-transaction": "^0.0.0",
    "@defichain/whale-api-client": "^0.0.0",
```

ivan-zynesis:
> Some minor stuff, and preferably add a test for invalidation logic (unless that's going to be tackled in a follow-up PR)

Good stuffs, will revise next week.

packages/rich-list-core/__tests__/RichListCore.test.ts
ivan-zynesis: The logic makes sense, technically the `invalidate()` execution itself can be tested. But since it is a private func,  should be handled in some newer PR with e2e.

packages/rich-list-core/src/RichListCore.ts
```diff
@@ -0,0 +1,237 @@
...
+    const balances: AddressAccountAmount = {}
+    for (const a of addresses) {
+      const nonZeroBalances = await this.whaleRpcClient.account.getTokenBalances(
+        { limit: Number.MAX_SAFE_INTEGER },
+        true,
+        { symbolLookup: false }
+      )
+      balances[a] = this.appendZeroBalances(nonZeroBalances, tokens)
+      // TBD: should be combine utxo and DFI rich list
+      balances[a]['-1'] = new BigNumber(await this.whaleApiClient.address.getBalance(a))
```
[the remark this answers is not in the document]
ivan-zynesis: There is newer [ticket](https://github.com/DeFiCh/jellyfish/issues/1260) going to merge utxo and DFI token into single list, so not going to make this change.

packages/rich-list-core/src/RichListCore.ts
```diff
@@ -0,0 +1,236 @@
...
+  }
+
+  /**
+   * Expected to be called on application bootstarapped.
+   * Start sync-ing up rich list up to tip and stop.
+   * Application layer should also consume this on interval basis to allow rich list updates again
+   *
+   * @returns {void}
+   */
+  start (): void {
```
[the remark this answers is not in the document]
ivan-zynesis: Similar to dex-swap-indexer, rich-list also meant to be a standalone service. I think it was designed to be independent and simpler as they do not affect the core and not likely they will require much updates.

Even the logic/result might be outdated and less accurate in future will not affect any critical operation (they are for show only), we want minimal effort for maintenance.

packages/rich-list-core/__tests__/RichListCore.test.ts
eli-lim: One more thing to check in this test - address '2NCWAKfEehP3qibkLKYQjXaWMK23k4EDMVS' should be first in the list

Yea let's have a separate PR with more tests then :shipit:

eli-lim (approved):

cwkang1998 (dismissed):

eli-lim (approved):

ivan-zynesis:
Thanks for all the reviews, it has been iterated few times, should be ready for final review.
@canonbrother @kodemon @fuxingloh

packages/rich-list-core/__tests__/RichListCore.test.ts
cwkang1998: The tests is currently being added at #1224, along with tests for rich-list-api

canonbrother (approved):

kodemon (approved):

ivan-zynesis:
@cwkang1998 found a bug while adding e2e, holding back until patch submitted

cwkang1998 (approved):

kodemon (approved):

eli-lim (approved):

PR #1089, opened 2022-02-23
fix(apps/legacy-api): reject with 404 for queries with invalid network url param

eli-lim:
- Since WhaleApiClientProvider caches clients by network, currently it naively creates new clients based on the 'network' param. 
- Potential memory issues if the endpoints are spammed with invalid networks - the cache will get filled with clients that will never be cleared
- Solution: reject all invalid `network` values, as the WhaleApiClient doesn't validate the network value when instantiated. The only way it _detects_ an invalid network is when it tries to make the request and fails.

#### Additional comments?:
- Added a test to guard all registered routes in `legacy-api`
- Also refactored `'mainnet' | 'testnet' | 'regtest'` to use `NetworkName`

### DIFF
--- a/apps/legacy-api/src/common/networks.ts
+++ b/apps/legacy-api/src/common/networks.ts
@@ -1,0 +1,1 @@
+export type SupportedNetwork = 'mainnet' | 'testnet'
--- a/apps/legacy-api/src/common/networks.ts
+++ b/apps/legacy-api/src/common/networks.ts
@@ -1,1 +1,0 @@
-export type SupportedNetwork = 'mainnet' | 'testnet'
--- a/apps/legacy-api/src/pipes/NetworkValidationPipe.ts
+++ b/apps/legacy-api/src/pipes/NetworkValidationPipe.ts
@@ -4,15 +4,14 @@ import {
   Injectable,
   PipeTransform
 } from '@nestjs/common'
-import { NetworkName } from '@defichain/jellyfish-network'
+import { SupportedNetwork } from '../common/networks'
 
 @Injectable()
 export class NetworkValidationPipe implements PipeTransform {
-  private static readonly VALID_NETWORKS: Set<undefined | NetworkName> = new Set([
+  private static readonly VALID_NETWORKS: Set<undefined | SupportedNetwork> = new Set([
     undefined, // defaults to 'mainnet'
     'mainnet',
-    'testnet',
-    'regtest'
+    'testnet'
   ])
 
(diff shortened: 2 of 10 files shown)


apps/legacy-api/src/pipes/NetworkValidationPipe.ts
```diff
@@ -0,0 +1,33 @@
...
+} from '@nestjs/common'
+import { NetworkName } from '@defichain/jellyfish-network'
+
+@Injectable()
+export class NetworkValidationPipe implements PipeTransform {
+  private static readonly VALID_NETWORKS: Set<undefined | NetworkName> = new Set([
+    undefined, // defaults to 'mainnet'
+    'mainnet',
+    'testnet',
+    'regtest'
```
fuxingloh, line 15:
```suggestion
```

let's remove this too

apps/legacy-api/testing/LegacyStubServer.ts
```diff
@@ -21,10 +23,36 @@ export class LegacyStubServer extends RootServer {
...
+  private recordAllRegisteredRoutes (app: NestFastifyApplication): void {
+    app.getHttpAdapter()
+      .getInstance()
+      .addHook('onRoute', (opts: RegisteredRoute) => {
+        this.allRoutes.push(opts)
+      })
+  }
+}
+
+// See https://www.fastify.io/docs/latest/Reference/Hooks/#onroute
```
fuxingloh, line 51:
```suggestion
/**
 * @see https://www.fastify.io/docs/latest/Reference/Hooks/#onroute
 */
```

should stick with JSDoc format when possible

fuxingloh (changes requested):

apps/legacy-api/src/common/networks.ts
```diff
@@ -0,0 +1 @@
+export type SupportedNetwork = 'mainnet' | 'testnet'
```
fuxingloh, line 1: I think it would make sense to put this in [NetworkValidationPipe.ts](https://github.com/DeFiCh/jellyfish/pull/1089/files/75245ce621d2cc98ecea38979424ca401bc2b359..2c5cec3130c7328738b3f57e2e54d3880d14ad5f#diff-3053d4c9b6e6b66c60ae4ed501681937046f662c90d30b50ed74832ece2a04e3) since the concerns are tightly related.

eli-lim: yup makes sense

fuxingloh (approved):

PR #1085, opened 2022-02-22
feat(docker): add docker release workflow

kodemon:
#### What this PR does / why we need it:

Shift left release flow for turning our apps into docker containers.

#### Which issue(s) does this PR fixes?:

Partially Fixes #943

#### Additional comments?:

Since we need to target `linux/arm64` docker build times are substantially increasing build times from a few minutes to an hour. 

Still looking into `arm64` build performance.

PR #1092, opened 2022-02-23
feat(apps): add apps release workflow

### LINKED ISSUE #943: `/apps/ocean-api`: CD Publish Docker Image

kodemon:

Add shift left automation for application containers.

Fixes #943

#### Additional comments?:

This is part 2 following up on

 - https://github.com/DeFiCh/jellyfish/pull/1085

This consolidates all application release logic into a single yml workflow file, and adds staging builds when merging to the main branch and dev builds when creating PRs against main.


canonbrother (dismissed):

kodemon:
I want to emphasize that the build times for our docker containers are ~1 hour long because of the `arm64` requirement. This workflow triggers on release, main and PR to main. I think for release and main the build time should be fine. But for PRs they may be a little intrusive, so depending on how things goes we may want to be more selective with the PR triggers.

We can be more targeted I think, maybe do manual build triggers for PRs so that we can opt in on container releases for branches that see a lot of activity?

@fuxingloh thoughts?

fuxingloh:
> @fuxingloh thoughts?

I think it's fine because we don't need to wait for them to pass before merging - furthermore our PR takes way longer than 1 hour to review.

By taking on the shift left mentality, we can have visibility of failures earlier which is good!

.github/scripts/release-tags.js
```diff
@@ -0,0 +1,65 @@
...
+function getReleaseTag(app, context) {
+  const semver = context.payload.release.tag_name.replace('v', '')
+  if (semver.match(/^[0-9]+\.[0-9]+\.[0-9]+$/) === false) {
+    throw new Error(`Release Violation: Provided version '${semver}' is not valid semver.`)
+  }
+  return `ghcr.io/defich/${app}:latest,ghcr.io/defich/${app}:${semver}`
+}
+
+function getStagingTag(app, context) {
+  return `ghcr.io/defich/${app}:main-${context.sha.substr(0, 12)}`
```
fuxingloh, line 60:
```suggestion
  const sha = context.sha.substr(0, 12)
  return `ghcr.io/defich/${app}:main,ghcr.io/defich/${app}:main-${sha}`
```

.github/workflows/release-apps.yml
```diff
@@ -0,0 +1,52 @@
...
+      
+      - name: Build & Publish
+        uses: docker/build-push-action@7f9d37fa544684fb73bfe4835ed7214c255ce02b
+        with:
+          push: true
+          build-args: APP=${{ matrix.app }}
+          platforms: |
+            linux/amd64
+            linux/arm64
+          tags: ${{ steps.tags.outputs.result }}
```
fuxingloh, line 52: For PR we should add something like this, easily allow the reader to grab the hash for testing.

via: `if: github.event_name == 'pull_request'`

https://github.com/DeFiCh/wallet/blob/df1696ccb922541c9b279944abb39d3bf4ee8b10/.github/workflows/expo-preview.yml#L35-L44

fuxingloh:
A few minor QOL suggestions.

.github/workflows/release-apps.yml
kodemon: yeah, that makes sense, will update.

.github/scripts/release-tags.js
kodemon: updated, since we are using it in two places I made a `getShortSha` method

.github/workflows/release-apps.yml
```diff
@@ -0,0 +1,52 @@
...
+
+jobs:
+  build:
+    name: Release Apps
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        app: [ legacy-api, ocean-api ]
+    steps:
+      - uses: actions/checkout@v2
```
canonbrother, line 20:
```suggestion
      - uses: actions/checkout@230611dbd0eb52da1e1f4f7bc8bb0c3a339fc8b7
```
better
https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions

.github/scripts/release-tags.js
fuxingloh: To add another `ghcr.io/defich/${app}:main` too

kodemon: done, I actually had the tagging a little wrong so I have updated to adhere to the request in the ticket

.github/workflows/release-apps.yml
```diff
@@ -0,0 +1,71 @@
+name: Release Apps
+
+on:
+  release:
+    types: [ published ]
+  push:
+    branches: [ main ]
+  pull_request_target:
```
fuxingloh, line 8:
```suggestion
  pull_request:
```

I think it should be `pull_request`, since `_target` is a privileged environment.

kodemon: changed to use `pull_request`

.github/workflows/release-apps.yml
```diff
@@ -0,0 +1,70 @@
...
+        uses: marocchino/sticky-pull-request-comment@39c5b5dc7717447d0cba270cd115037d32d28443
+        if: github.event_name == 'pull_request'
+        with:
+          header: apps-preview
+          message: |
+            Build preview for jellyfish apps is ready!
+            
+            Built with commit ${{ github.sha }}
+            
+            ${{ needs.build.outputs.tags }}
```
kodemon, line 70: This is not yet outputting the way we want, still working on this.

.github/workflows/release-apps.yml
```diff
@@ -0,0 +1,74 @@
...
+            linux/arm64
+          tags: ${{ steps.tags.outputs.result }}
+
+  report:
+    name: Report
+    runs-on: ubuntu-latest
+    needs: build
+    if: github.event_name == 'pull_request'
+    steps:
+      - uses: actions/checkout@230611dbd0eb52da1e1f4f7bc8bb0c3a339fc8b7
```
fuxingloh, line 59: Minor suggestion for optimization, since you are checking out the entire workflow with this step. Would it make sense to just have GitHub scripts with the embedded script since it's pretty short - to also keep it close and contextual together?

Just a suggestion, I am on the fence here. Either way works imo.

.github/workflows/release-apps.yml
```diff
@@ -0,0 +1,74 @@
+name: Release Apps
+
+on:
+  release:
+    types: [ published ]
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
+
```
fuxingloh, line 10: For security hardening should add permission.

https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs
https://docs.github.com/en/rest/reference/permissions-required-for-github-apps

.github/workflows/release-apps.yml
```diff
@@ -0,0 +1,74 @@
...
+jobs:
+  build:
```
fuxingloh, line 12: `if: github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository`

To prevent all workflow from running if it's coming from a forked pull request. Since it can't run without failing as they can't get the access to write into ghcr.io

fuxingloh (changes requested):
A few more minor suggestions and changes

.github/workflows/release-apps.yml
kodemon: While I would definitely advocate keeping things as close as possible, I think it this case it bloats the file too much. If the scripts themselves were smaller i'd probably leave them in though. In this case though I have emphasized writing it up so that any future additions to the yml stays isolated to the yml while the scripts are more abstract.

I haven't seen any performance concerns with the checkout yet, but if we see it as an issue we could move the scripts inline.

I will defer to your judgment here, but personally I think its at least more readable this way.

.github/workflows/release-apps.yml
kodemon: After a little thought I updated the script to use the github org/username as deploy target so that we don't have to worry about hard coded endpoints when people fork the repository.

.github/workflows/release-apps.yml
kodemon: I have been wondering what sort of permissions we may want to add here.

I assume we add permissions so that any third party package we are using is not able to perform operations on our repository they are not permitted to?

If my assumption is correct I will work to figure out what the minimum permissions are to fulfill our workflow.

.github/workflows/release-apps.yml
```diff
@@ -0,0 +1,75 @@
...
+          script: return require('./.github/scripts/release-tags.js')({ context })
+          result-encoding: string
+      
+      - name: Build & Publish
+        uses: docker/build-push-action@7f9d37fa544684fb73bfe4835ed7214c255ce02b
+        with:
+          push: true
+          build-args: APP=${{ matrix.app }}
+          platforms: |
+            linux/amd64
```
fuxingloh, line 50: This should be separated into a matrix because we want to run them concurrently so that it's faster in parallel.

For speed to deployment - we should not need to wait for both platforms to be built and then pushed.

.github/workflows/release-apps.yml
```diff
@@ -0,0 +1,75 @@
...
+      - uses: actions/checkout@230611dbd0eb52da1e1f4f7bc8bb0c3a339fc8b7
+
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@27d0a4f181a40b142cce983c5393082c365d1480
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@94ab11c41e45d028884a99163086648e898eed25
+
+      - name: Login to GitHub Container Registry
+        uses: docker/login-action@6af3c118c8376c675363897acf1757f7a9be6583
```
fuxingloh, line 29:
```suggestion
        uses: docker/login-action@bb984efc561711aaa26e433c32c3521176eae55b
```

A recent update in https://github.com/DeFiCh/jellyfish/pull/1123

.github/workflows/release-apps.yml
kodemon: Makes sense, will update.

.github/workflows/release-apps.yml
fuxingloh: To create another PR for this, since it's low priority.

fuxingloh: Created a new ticket to track it: https://github.com/DeFiCh/jellyfish/issues/1151

fuxingloh:
@kodemon can you create another PR to add `playground-api` in the matrix?

PR #1093, opened 2022-02-23
feat(jellyfish-testing):Add TestingWrapper as an abstraction layer

surangap:
Tries a different approach to address https://github.com/DeFiCh/jellyfish/issues/637 and https://github.com/DeFiCh/jellyfish/issues/965 instead of #1035

Here we introduce  `TestingWrapper` class to create a new layer of abstraction with the goal of being able to create a test /testGroup instance in a standardized manner.

Also adds the capability to add NonMasternode RegTestContainers to the tests. For this, `NonMNTesting` class has been introduced.

Again this is not the most perfect way to do this. But given the restrictions discussed here -> https://github.com/DeFiCh/jellyfish/pull/1035#discussion_r808761851, this way seems not ugly and at the same time serves the purpose.

Fixes https://github.com/DeFiCh/jellyfish/issues/637
Fixes https://github.com/DeFiCh/jellyfish/issues/965

### DIFF
--- a/packages/jellyfish-testing/src/testingwrapper.ts
+++ b/packages/jellyfish-testing/src/testingwrapper.ts
@@ -1,0 +1,57 @@
+import { RegTestFoundationKeys } from '@defichain/jellyfish-network'
+import { ContainerGroup, MasterNodeRegTestContainer, RegTestContainer } from '@defichain/testcontainers'
+import { Testing, TestingGroup, NonMNTesting } from '.'
+
+type InitMNRegContainerFn = (index: number) => MasterNodeRegTestContainer
+function defaultInitMNRegContainer (index: number): MasterNodeRegTestContainer {
+  return new MasterNodeRegTestContainer(RegTestFoundationKeys[index])
+}
+
+type InitNonMNRegContainerFn = (index: number) => RegTestContainer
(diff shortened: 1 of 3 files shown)


packages/jellyfish-testing/src/testingwrapper.ts
```diff
@@ -0,0 +1,57 @@
...
+
+  create (n?: number, init: InitMNRegContainerFn = defaultInitMNRegContainer): Testing | TestingGroup {
+    if (n === undefined || n <= 1) {
+      return Testing.create(init(0))
+    }
+
+    return TestingGroup.create(n, init)
+  }
+
+  createNonMN (): NonMNTesting
```
kodemon, line 29: Wanted to start a quick discussion on naming since it will be harder to change later once this starts spreading into our testing structures.

From initial reading `NonMN` twists the tongue a little. Wondering if perhaps we want to create a proper name for a reg container that is not a MasterNode. I was thinking perhaps PartialNode?

And then since we are creating individual methods with expected outcomes perhaps the wrapper exposes.

`createMasterNode` and `createPartialNode`

It creates a little more uniformity in naming and I think more clear in their meaning, thoughts?

Issue #967, opened 2022-01-12
documentation operations and workflow


fuxingloh:
#### What would you like to be added:

Shift Left Documentation and Examples by
- Importing Codeblock from shipped code instead of duplicating it
- Importing Jest Test suite as an example instead of duplicating it
- Examples to be run as automated E2E test

Documentation is part of the living standard of the SDK they should be tested, shift-left. The above recommendation suggests ways where we can improve developer operations by making documentation be part of the workflow.

/area docs examples
/sig ecosystem

PR #1104, opened 2022-02-25
chore(doc): setup shift left documentation

cwkang1998:
Attempts to setup shift left documentation, according to #967.

#### Additional comments?:
Currently there are 2 methods that are considered for this implementation.

1. Importing the code directly from github.

> Importing Codeblock from shipped code instead of duplicating it

Taking from this literally, this approach attempts to reduce the amount of duplication by allowing documentation to be imported using github links. This approach is done via the use of https://github.com/saucelabs/docusaurus-theme-github-codeblock.

Pros:
- Straight forward and easy to extend and implement
- Allows flexibility in current style of documentation
- Allow importing whatever code you desire in the code base since its just a github link to the code lines.

Cons: 
- Still manual work in managing the docs
- No way of managing the changes that happen out of the box for the github links, meaning that on releases that changes the code mentiioned in documentation we have to make sure it points to the same area. (Pining the docs via commit hashes would be a way to semi solve this)

2. Automating code generation

This approach removes the need to maintaining the documentation by relying on documentation generation tools. This can be done via a variety of tools, but so far the tool that I find the most success in using and provides the most information for typescript is [api-extractor](https://api-extractor.com/).

Some nice examples:
- https://rushstack.io/pages/api/
- https://faastjs.org/docs/api/faastjs

Pros:
- Encourages docs to be written with code as Tsdoc
- Automates a lot of documentation writing process as a lot of information can be inferred from code.

Cons:
- Require the use of [Tsdoc](https://tsdoc.org/) standard to make it work best. (Although it works okay without it)
- Need some work to get it working with current docusaurus setup (generated markdown is not directly compatible with current site)
- Specific syntax requirements for typescript, `export * as somethingElse from 'something` is not supported yet. instead they require something like `import * as somethingElse from 'something'` then only `export {something}`. Although this is only a problem in `jellyfish-api-core`, and can be fixed easily.

A screenshot from testing, you can see it does generate docs successfully from current code, just that its gonna need some configuration work and styling work to be done to make it look pretty. Screenshot from 2022-02-26 01-34-32

This PR right now is only working on method 1, as method 2 produces a lot of markdown files, and as such I refrained from pushing to reduce noise in commit changes.


cwkang1998:
@fuxingloh Would love to hear your thoughts on this.

cwkang1998:
Right now I've only applied this to `docs/node/CATEGORIES/01-blockchain.md` as a prove of concept.  Any feedback is welcomed.

docs/node/CATEGORIES/01-blockchain.md
```diff
@@ -17,138 +17,44 @@ const something = await client.blockchain.method()
...
 
 Get block header data with particular header hash.
 
-```ts title="client.blockchain.getBlockHeader()"
+```ts reference title="client.blockchain.getBlockHeader()"
 interface blockchain {
-  getBlockHeader (hash: string, verbosity: true): Promise<BlockHeader>
-  getBlockHeader (hash: string, verbosity: false): Promise<string>
+  i('https://github.com/DeFiCh/jellyfish/blob/main/packages/jellyfish-api-core/src/category/blockchain.ts#L89')
+  i('https://github.com/DeFiCh/jellyfish/blob/main/packages/jellyfish-api-core/src/category/blockchain.ts#L89')
```
canonbrother, line 54: i think i'm outdated.. so we will be using `i(url:line)` to indicate the interface??
and these 2 lines are same.. i think the line should be diff

cwkang1998: The idea is using `i(url:line)` to directly import code (not only interface) from the file into the code block. It's still quite limiting actually so I am exploring other options.
 Regarding the 2 lines, I think i made a mistake, will change it.

fuxingloh: Interesting approach, hmm would it be possible to do symbol import instead? Since line items get shifted around when we edit the source code.

ivan-zynesis: The approach only looks clean but effectively made it even harder to maintain.
It still requires manual maintenance regardless and copy pasting the interface is easier than find out the correct GH link.

cwkang1998:
Screenshot 2022-03-09 at 12 37 34 PM

Currently got typedocs working with docusaurus, without the need of strictly adhering to tsdocs, and at the same time able to read signature from the code itself. The docs are now auto gen from the `all:docs` command, and would be served at docusaurus at the `/packages` path.

PR #1118, opened 2022-02-28
feat(rich-list-core): implement `poolpair` transactions parsers

cwkang1998:

This is built upon #1084, so it should only be merged after its parent is resolved. Related to #1053,

Implements `poolpair` transaction parsers (i.e. `PoolAddLiquidity`, `PoolRemoveLiquidity`, etc.) for address parser.


packages/jellyfish-api-core/src/category/poolpair.ts
```diff
@@ -162,17 +162,17 @@ export class PoolPair {
...
+   * @param {string} address defi address containing the liquidity tokens
+   * @param {string} amount 'liquidityAmount@liquidity' pool symbol, e.g. BTC-DFI
    * @param {PoolLiquidityOptions} [options]
    * @param {UTXO[]} [options.utxos] utxos array of specific UTXOs to spend
    * @param {string} [options.utxos.txid]
    * @param {number} [options.utxos.vout]
    * @return {Promise<string>}
    */
-  async removePoolLiquidity (address: string, poolAccount: string, options: PoolLiquidityOptions = {}): Promise<string> {
+  async removePoolLiquidity (address: string, amount: string, options: PoolLiquidityOptions = {}): Promise<string> {
```
ivan-zynesis, line 173: I prefer the old naming, `123@BTC-DFI` is not an "amount".

If we were going to overhaul things, I will go for amount and pair ID as 2 individual argument, but not likely for sake of backward compatibility.

cwkang1998: Got it. I think we don't need to seperate things as well. I'll just add these information to the comments.

cwkang1998:
Closing this as combining this with #1099

PR #1125, opened 2022-03-01
feat(apps/legacy-api): stats jsonpath query param

### LINKED ISSUE #1117: legacy-api: `/stats` to support q=json.path and return as text format

`/stats` to support `q=json.path` and return as text format

/triage accepted
/assign @eli-lim 
/area apps
/priority urgent-now

eli-lim:
Support `q` json path query param for the legacy-api /stats endpoint

Fixes #1117

#### Additional comments?:
- Added lodash for json path lookup (_.get)


apps/package.json
```diff
@@ -32,6 +32,7 @@
     "class-transformer": "^0.5.1",
     "class-validator": "^0.13.2",
     "joi": "^17.6.0",
+    "lodash.get": "^4.4.2",
```
fuxingloh, line 35: Since `lodash` is likely already an internal dependency, maybe it's better to just import the entire `lodash` package?

When compiled it will be tree shaken out, giving the same benefit of using `lodash.get.

apps/legacy-api/src/controllers/stats/LegacyStatsProvider.ts
```diff
@@ -51,7 +52,7 @@ export class MainnetLegacyStatsProvider {
     this.api = clientProvider.getClient('mainnet')
   }
 
-  async getStats (): Promise<LegacyStats> {
+  async getStats (jsonPath: string): Promise<LegacyStats> {
```
fuxingloh, line 55:
```suggestion
  async getStats (jsonPath?: string): Promise<LegacyStats> {
```

apps/package.json
eli-lim: Hmm I don't see lodash being used anywhere in jellyfish right now (ctrl shift f 'lodash'), don't see it in root package.json either

fuxingloh: `npm list lodash`, it's a internal dependencies, when you add it to the root `apps/package.json`, it will be deduped

fuxingloh (approved):

PR #1127, opened 2022-03-01
fix(docker): don't need to specify ${APP} twice in RUN ln due to nest-apps mono repo design update

fuxingloh:

`Dockerfile` build was failing as `${APP}` is now redundant due to nestjs mono repo structural changes.

> #1073 will shift-left migrated invalid artifact in the future.


eli-lim:
LGTM

PR #1145, opened 2022-03-02
refactor(jellyfish-transaction): reorganize code in `dftx_loans` into sections

### LINKED ISSUE #811: Organize Loan & Vault `DfTx` into their sections
> Can do code organization after development. I think we have some rearrangements to do in others as well.

Loan | Vault

_Originally posted by @surangap in https://github.com/DeFiCh/jellyfish/pull/731#pullrequestreview-799761520_

/triage accepted
/area jellyfish-transaction jellyfish-transaction-builder

cwkang1998:
Reorganize code per linked issue.

Fixes #811


chee-chyuan (approved):

canonbrother (dismissed):

packages/jellyfish-transaction/src/script/dftx/dftx_loans.ts
```diff
@@ -271,6 +271,30 @@ export class CWithdrawFromVault extends ComposableBuffer<WithdrawFromVault> {
   }
 }
 
+/**
+ * CloseVault DeFi Transaction
+ */
+export interface CloseVault {
```
ivan-zynesis, line 277: I think what @surangap mean in the original comment is to move these into `dftx/vault`

Can refer to https://github.com/DeFiCh/ain/tree/master/src/masternodes (which IMO, some of them are not quite in the right place)

ivan-zynesis (changes requested):

packages/jellyfish-transaction/src/script/dftx/dftx_loans.ts
cwkang1998: For reference, what are the things that you think are not in the right place, and where do you think they should be?

canonbrother:
> Regarding `jellyfish-api-core`, where vault operations are still accessible via `rpc.loan.createVault` etc., should we also refactor that? It'll be a backwards breaking change, but just making the change here and not there kind of does not makes a lot of sense

yea i think reorganize code according to ain structure
https://github.com/DeFiCh/ain/blob/bb394bec2b08087885b008e00e6178fc41b87caa/src/masternodes/rpc_loan.cpp#L1428-L1444
https://github.com/DeFiCh/ain/blob/bb394bec2b08087885b008e00e6178fc41b87caa/src/masternodes/rpc_vault.cpp#L1778-L1791

cwkang1998:
This release causes a lot of breaking changes since there's a major changes on the current loan api

ivan-zynesis:
> This release causes a lot of breaking changes since there's a major changes on the current loan api.

Can deprecate the old one without remove them. Call `vault.xxx()` in `loan.xxx()`

fuxingloh:
> > This release causes a lot of breaking changes since there's a major changes on the current loan api.
> 
> Can deprecate the old one without remove them. Call `vault.xxx()` in `loan.xxx()`

I realize there is quite a significant amount of change going on in this PR. To reduce the impact of the PR and its' releases. Maybe we can make small efforts and create smaller PR bit by bit? Breaking it down, 1 by 1 before going all in.

surangap:
> > > This release causes a lot of breaking changes since there's a major changes on the current loan api.
> > 
> > 
> > Can deprecate the old one without remove them. Call `vault.xxx()` in `loan.xxx()`
> 
> I realize there is quite a significant amount of change going on in this PR. To reduce the impact of the PR and its' releases. Maybe we can make small efforts and create smaller PR bit by bit? Breaking it down, 1 by 1 before going all in.

might need `TxnBuilderLoans` also with deprecated ones.

cwkang1998:
> > > This release causes a lot of breaking changes since there's a major changes on the current loan api.
> > 
> > 
> > Can deprecate the old one without remove them. Call `vault.xxx()` in `loan.xxx()`
> 
> I realize there is quite a significant amount of change going on in this PR. To reduce the impact of the PR and its' releases. Maybe we can make small efforts and create smaller PR bit by bit? Breaking it down, 1 by 1 before going all in.

Closing as its divided into 3 different PR for different packages for a smaller release.

Issue #1024, opened 2022-02-04
`/apps/legacy-api` - Remaps old stats API


joeldavidw:
#### What would you like to be added:
Implementation of `legacy-api` to remap the legacy stats API (https://api.defichain.com/v1/*). 

Legacy API Endpoints

- [x] `gettoken`
- [x] `listtokens`
- [x] `getpoolpair`
- [x] `listpoolpairs`
- [x] `getblockcount`
- [x] `validateAddress` #1090
- [x] `getgov` #1090 
- [x] `listyieldfarming` #1115 
- [x] `listswaps` #1102
- [ ] `getswaptransaction`
- [x] `getsubgraphswaps` WIP -> #1119
- [ ] `gettotalvolumestats`
- [ ] `getliquiditytotalvolumedata`
- [x] `stats` #1100 
- [ ] `stats/tickers`

Low Priority
- [ ] `getanchorlist`
- [ ] `getaccount`
- [ ] `getcustomtx`
- [ ] `listaccounthistory`
- [ ] `gettokenrichlist`
- [ ] `getblockactivity`
- [ ] `getpoolpairlplist`

### Notes
- http://json2ts.com/ will be helpful - copy the response from https://api.defichain.com/v1/* and paste it in, generating typescript interface that will greatly reduce the burden to figure out the correct json shape and types (especially for large ones)

---

/sig ecosystem
/area apps
/triage accepted

fuxingloh:
Since #1011 is merged, need your help to update all the endpoints. What's missing, what's done. What can't be done, etc.

fuxingloh:
I assume we can ignore:
- getanchorlist
- getaccount
- getcustomtx
- listaccounthistory
- gettokenrichlist
- getblockactivity
- getpoolpairlplist

fuxingloh:
We need to prioritize all pool related API.

eli-lim:
Just wanna check - are we maintaining the types too?
```javascript
{
  // ...
  "burnInfo": {
    "address": "8defichainBurnAddressXXXXXXXdRQkSm",
    "amount": "156002224.91060823", // ------------- string
    "tokens": [
      "7260.95539278@DFI",
      "6550.00000000@ETH",
      "2476.86380002@BTC",
      "14032000.00000000@USDT",
      "16820000.00000000@DOGE",
      "19051.00000000@LTC",
      "3131.00000000@BCH",
      "4850000.00000000@USDC"
    ],
[...]

PR #1148, opened 2022-03-03
feat(apps/legacy-api): getsubgraphswaps

eli-lim:
Implementation of https://api.defichain.com/v1/getsubgraphswaps in legacy-api

Fixes part of https://github.com/DeFiCh/jellyfish/issues/1024

### Note
- query performance bottleneck is getaccounthistory

### DIFF
--- a/apps/legacy-api/__tests__/controllers/PoolPairController.test.ts
+++ b/apps/legacy-api/__tests__/controllers/PoolPairController.test.ts
@@ -158,8 +158,7 @@ describe('getsubgraphswaps', () => {
   it('/v1/getsubgraphswaps', async () => {
     const res = await apiTesting.app.inject({
       method: 'GET',
-      // TODO(eli-lim): temporary deployed TestNet for testing, to switch to mainnet
-      url: '/v1/getsubgraphswaps?network=testnet'
+      url: '/v1/getsubgraphswaps'
     })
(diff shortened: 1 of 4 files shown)


apps/legacy-api/__tests__/controllers/PoolPairController.test.ts
```diff
@@ -153,3 +153,64 @@ it('/v1/listyieldfarming', async () => {
...
+            decimals: 8,
+            symbol: expect.any(String),
+            tradeVolume: expect.stringMatching(/^[0-9]+(\.[0-9]{8})?$/)
+          }
+        },
+        timestamp: expect.stringMatching(/\d+/),
+        fromAmount: expect.stringMatching(ONLY_DECIMAL_NUMBER_REGEX),
+        toAmount: 'TODO' // TODO(eli-lim)
+      })
+    }
```
fuxingloh, line 187: shouldn't skip, should test it?

Speaking of this mapping format, should we map to follow this standard as well? On top of what we have

eli-lim: Woops - skipped it originally because of local jest timing out. unskipped and changed the response format.

fuxingloh (approved):

PR #1165, opened 2022-03-04
feat(workflow): skip container builds for dependabot PRs

kodemon:

Dependabot does not have access to publish containers so we skip PR container builds in those circumstances:

 - https://github.com/DeFiCh/jellyfish/pull/1156#issuecomment-1058802501

### DIFF
--- a/.github/workflows/release-apps.yml
+++ b/.github/workflows/release-apps.yml
@@ -10,7 +10,7 @@ on:
 
 jobs:
   build:
-    if: github.actor != "dependabot[bot]"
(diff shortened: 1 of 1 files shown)


fuxingloh (approved):

fuxingloh:
Doesn't seem like it's reporting or building

kodemon:
Yeah, will investigate further 🤔

kodemon:
@fuxingloh seems to be building now, wondering if the `"` to `'` was all that was needed.

fuxingloh (approved):

PR #1168, opened 2022-03-04
feat(whale): migrate `apps/whale-api`

kodemon:
#### What this PR does / why we need it:

Consolidation of our public apis into the jellyfish ecosystem.

#### Which issue(s) does this PR fixes?:

Fixes #977

#### Additional comments?:

Migrate `whale/src` to `apps/whale-api`, restructures and renames files and folders based on the other apps in the ecosystem.

Efforts to split up this PR into smaller digestible pieces and introducing tests:

 - https://github.com/DeFiCh/jellyfish/pull/1201

Depends on following PRs:

 - https://github.com/DeFiCh/jellyfish/pull/1166
 - https://github.com/DeFiCh/jellyfish/pull/1167

Related whale migration PRs:

 - https://github.com/DeFiCh/jellyfish/pull/1170

PR #1169, opened 2022-03-04
feat(packages): add additional exports to support `apps/whale-api`

### LINKED ISSUE #977: `/apps/whale-api` - Consolidation & Migration

kodemon:

Since we cannot rely on `/dist` imports within jellyfish we need to provide the expected values out of our core packages for use in our apps. These are plucked exports as `export * from '...'` results in naming conflicts.

Partially Fixes #977

#### Additional comments?:

This is a precursor PR for:

 - https://github.com/DeFiCh/jellyfish/pull/1168

### DIFF
--- a/packages/jellyfish-api-core/src/index.ts
+++ b/packages/jellyfish-api-core/src/index.ts
@@ -35,6 +35,25 @@ export * as icxorderbook from './category/icxorderbook'
 export * as misc from './category/misc'
 export * as loan from './category/loan'
 
+export { AccountHistory } from './category/account'
+export { BlockchainInfo } from './category/blockchain'
+export {
+  AuctionPagination,
+  CollateralTokenDetail,
+  GetLoanSchemeResult,
+  LoanSchemeResult,
+  LoanTokenResult,
+  VaultActive,
+  VaultLiquidation,
+  VaultLiquidationBatch,
+  VaultPagination,
+  VaultState
+} from './category/loan'
+export { EstimateMode } from './category/mining'
+export { PoolPairInfo } from './category/poolpair'
+export { TokenInfo, TokenResult } from './category/token'
+export { MasternodeInfo, MasternodeState, MasternodeTimeLock } from './category/masternode'
+
 /**
  * A protocol agnostic DeFiChain node client, RPC calls are separated into their category.
  */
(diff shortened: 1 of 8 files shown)


packages/jellyfish-api-core/src/index.ts
```diff
@@ -35,6 +35,25 @@ export * as icxorderbook from './category/icxorderbook'
...
+  VaultActive,
+  VaultLiquidation,
+  VaultLiquidationBatch,
+  VaultPagination,
+  VaultState
+} from './category/loan'
+export { EstimateMode } from './category/mining'
+export { PoolPairInfo } from './category/poolpair'
+export { TokenInfo, TokenResult } from './category/token'
+export { MasternodeInfo, MasternodeState, MasternodeTimeLock } from './category/masternode'
```
fuxingloh, line 55: I am thinking whether:

1. we should just export everything and see how well it can deconflict?
2. used named exports? e.g: 
   1. `export * as account from './category/account'`
   2. `import account from '...'`
   3.  `account.AccountHistory`


Since this is explicitly declared just for whale-api, might seem arbitrary and hard to maintain?

kodemon: We could try going for `option 1` and see if we can't deconflict.

From memory the conflicts are usually related to the same interface being declared in different files. So I think it should be possible to DRY it up and create shared interfaces. I didn't do this as part of the original whale migration PR because it will require a lot more unrelated refactoring of jellyfish core packages. And the current implementation of this is the path of least resistance to get whale migrated without having to refactor any code.

I would say its probably the right direction to head though so we can correct some past mistakes in terms of our export structures.

kodemon: Perhaps we can get this in with a `todo` ticket for refactoring export structures in jellyfish-core?

This way we may prevent bloating time frames for whale consolidation so we can continue whale work within jellyfish and shift the ongoing refactoring to the right?

kodemon: Did an initial pass at deconflicting the exports so we can do a `export *` approach. All of them except for 1 was `UTXO` duplication. The one I am unsure about is `MasterNodeInfo`, for now I added the info coming from the master node file itself as the source of truth.

kodemon:
Setting the PR back to draft as I am trying out another path of least resistance suggested by @fuxingloh

kodemon:
Closing this as its not longer required, in migration changing `/dist` to `/src` in the imports seems to do the trick.

PR #1166, opened 2022-03-04
feat(whale): migrate `packages/whale-api-client`

kodemon:
#### What this PR does / why we need it:

Migrate `whale-api-client` from whale to jellyfish repo in consolidation efforts of the whale api.

#### Which issue(s) does this PR fixes?:

Partially Fixes #977

#### Additional comments?:

Related to porting whale RawTx to ocean:

 - https://github.com/DeFiCh/jellyfish/pull/948
 - https://github.com/DeFiCh/jellyfish/pull/949

This is a leading PR for whale api consolidation and is followed by:

 - https://github.com/DeFiCh/jellyfish/pull/1168
 - https://github.com/DeFiCh/jellyfish/pull/1167
 - https://github.com/DeFiCh/jellyfish/pull/1170

PR #1170, opened 2022-03-04
feat(legacy-api): update `whale-api-client` imports

kodemon:

Since we are no longer targeting external package of `whale-api-client` we cannot use `/dist` targets. This PR removes the dependency on `/dist` and replaces it with `/src` as seen in other areas of apps in jellyfish. [Sample](https://github.com/DeFiCh/jellyfish/blob/main/apps/legacy-api/src/controllers/PoolPairController.ts#L16)

#### Additional comments?:

This PR depends on:

This is part of whale migration efforts:

 - https://github.com/DeFiCh/jellyfish/pull/1168

### DIFF
--- a/apps/legacy-api/__tests__/controllers/PoolPairController.test.ts
+++ b/apps/legacy-api/__tests__/controllers/PoolPairController.test.ts
@@ -1,5 +1,5 @@
 import { LegacyApiTesting } from '../../testing/LegacyApiTesting'
-import { PoolPairData } from '@defichain/whale-api-client/src/api/PoolPairs'
+import { PoolPairData } from '@defichain/whale-api-client/dist/api/poolpairs'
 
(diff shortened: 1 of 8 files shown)


fuxingloh (dismissed):

fuxingloh:
I assume we need to target this for `main`?

kodemon:
@fuxingloh when api client goes into main this PR should then point back to main with the required dependency and run tests. Once dependency PR goes in it should kick of tests in this PR and I can mark it ready for review.

eli-lim (dismissed):

eli-lim:
Do we wanna update package-lock.json?

kodemon:
> Do we wanna commit the change to package-lock.json?

Updated, thanks for the catch 👍🏻

eli-lim (approved):

fuxingloh (approved):

PR #1176, opened 2022-03-07
refactor(apps): abstract root server

### LINKED ISSUE #1171: `apps/libs/server` abstract RootServer into a shared class

There are multiple `RootServer` that can be abstracted and de-duplicated across `apps`.

Also to set logging to default true: `const adapter = new FastifyAdapter({logger: true})`

/triage accepted
/area apps
/assign @canonbrother

canonbrother:

Fixes #1171


kodemon:
Looking at the initial abstraction [here](https://github.com/DeFiCh/jellyfish/pull/1176/files#diff-5200803d072a0debee2a9e93c1692a9d57907510bbe31fa7f5b9af6090568e70), it seems that we still require to write all the individual methods out in each app. Which may render the shared abstraction redundant apart from providing a shared interface _(which we sort of already get from nest perhaps?)_

I was expecting something in the line of potentially take anything that is custom and configure it through the `constructor` and having most the abstracted methods have shared implementation details in the RootServer class. On initial look I think `create,init,start,stop` could be unified and perhaps a default override-able `configuration`.

So we would end up with something like the following for the apps, removing potential duplication:

```ts
export class OceanApiServer extends RootServer {
  constructor() {
    super(RootModule)
  }
}
```

I might not be aware of a nest.js detail that prevents this though so these are base observations, thoughts?

ivan-zynesis:
> Looking at the initial abstraction [here](https://github.com/DeFiCh/jellyfish/pull/1176/files#diff-5200803d072a0debee2a9e93c1692a9d57907510bbe31fa7f5b9af6090568e70), it seems that we still require to write all the individual methods out in each app. Which may render the shared abstraction redundant apart from providing a shared interface _(which we sort of already get from nest perhaps?)_
> 
> I was expecting something in the line of potentially take anything that is custom and configure it through the `constructor` and having most the abstracted methods have shared implementation details in the RootServer class. On initial look I think `create,init,start,stop` could be unified and perhaps a default override-able `configuration`.
> 
> So we would end up with something like the following for the apps, removing potential duplication:
> 
> ```ts
> export class OceanApiServer extends RootServer {
>   constructor() {
>     super(RootModule)
>   }
> }
> ```
> 
> I might not be aware of a nest.js detail that prevents this though so these are base observations, thoughts?

Agree. So far this PR added only an interface/protocol for it, no implementation is abstracted out yet.

canonbrother:
> > Looking at the initial abstraction [here](https://github.com/DeFiCh/jellyfish/pull/1176/files#diff-5200803d072a0debee2a9e93c1692a9d57907510bbe31fa7f5b9af6090568e70), it seems that we still require to write all the individual methods out in each app. Which may render the shared abstraction redundant apart from providing a shared interface _(which we sort of already get from nest perhaps?)_
> > I was expecting something in the line of potentially take anything that is custom and configure it through the `constructor` and having most the abstracted methods have shared implementation details in the RootServer class. On initial look I think `create,init,start,stop` could be unified and perhaps a default override-able `configuration`.
> > So we would end up with something like the following for the apps, removing potential duplication:
> > ```ts
> > export class OceanApiServer extends RootServer {
> >   constructor() {
> >     super(RootModule)
> >   }
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I might not be aware of a nest.js detail that prevents this though so these are base observations, thoughts?
> 
> Agree. So far this PR added only an interface/protocol for it, no implementation is abstracted out yet.

hmm... yea.. the `abstract` is closed similar with `interface`.. the `RootServer` is just a restricted class for now..

apps/libs/rootserver/src/index.ts
```diff
@@ -0,0 +1,17 @@
...
+
+  abstract create (): Promise<NestFastifyApplication>
+
+  abstract configure (app: NestFastifyApplication, config: ConfigService): Promise<void>
+
+  abstract init (app: NestFastifyApplication, config: ConfigService): Promise<void>
+
+  abstract start (): Promise<void>
+
+  abstract stop (): Promise<void>
```
fuxingloh, line 16: I think a few methods can be implemented since they are not shared?

canonbrother: yea earlier i was thinking they might be vary in future and easy to fix... btw its' good to move them to root as well...

apps/libs/rootserver/src/index.ts
```diff
@@ -0,0 +1,35 @@
...
+    })
+  }
+
+  async init (app: NestFastifyApplication, config: ConfigService): Promise<void> {
+    const port = config.get<number>('PORT', 3000)
+    await app.listen(port, '0.0.0.0')
+  }
+
+  async start (): Promise<void> {
+    this.app = await this.create()
```
kodemon, line 25: Since we are assigning a new application instance to this variable. Should we be running `stop` before we create a new instance, in case another instance is already running?
```suggestion
  async start (): Promise<void> {
    await this.stop()
    this.app = await this.create()
```

apps/ocean-api/src/index.ts
```diff
@@ -1,56 +1,32 @@
...
-    const adapter = new FastifyAdapter()
-    return await NestFactory.create<NestFastifyApplication>(RootModule, adapter)
+    return await NestFactory.create<NestFastifyApplication>(RootModule, this.adapter)
   }
 
   async configure (app: NestFastifyApplication, config: ConfigService): Promise<void> {
+    await super.configure(app, config)
+
     const version = config.get<string>('API_VERSION') as string
     const network = config.get<string>('API_NETWORK') as string
```
ivan-zynesis, line 16: I think is okay to include this to abstract layer too, also `setGlobalPrefix` if version and network envvar available.

canonbrother: i implemented this but find out not so suitable :sweat:

cwkang1998 (approved):

eli-lim (approved):

ivan-zynesis (approved):

eli-lim:
:shipit:

kodemon (approved):

PR #1193, opened 2022-03-08
chore(tsconfig): add `lib: ["dom"]` globally in tsconfig.base.json

fuxingloh:

Due to the use of `cross-fetch` on `jellyfish-api-core`, we might need to add `lib: ["dom"]` so that is can build without error.

@kodemon this might be related to you.

### DIFF
--- a/tsconfig.base.json
+++ b/tsconfig.base.json
@@ -1,7 +1,7 @@
 {
   "compilerOptions": {
     "lib": [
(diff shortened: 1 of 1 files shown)


kodemon (approved):
Can confirm this fixes build issue in https://github.com/DeFiCh/jellyfish/pull/1166

Issue #1150, opened 2022-03-03
centralised Dex Swap indexer


eli-lim:
#### What would you like to be added:
- A centralised indexer for `AccountHistory` and possibly more.

#### Why is this needed:
- Legacy-api `/v1/getsubgraphswaps` endpoint is very slow due to getaccounthistory rpc bottlenecking.

### Design
- a **centralised** indexing model is needed, where only 1 worker will source accounthistory data from the blocks and write the relevant data to a central, horizontally scalable database.
- The data can then be exposed via a service layer, greatly improving query performance.
- This system / approach can be leveraged for other bottlenecks if they arise.

#### Components
Similar to CQRS
- ETL worker to read from chain and write to db
- Managed db
- Service to expose the data

#### Build items
- [ ] Scaffold of the new app - #1197
- [ ] Implementation of the dex swap indexer - #1203

fuxingloh:
A better title would be Dex Swap indexer?

Since account history involve everything, might be overkill?

PR #1197, opened 2022-03-09
feat(apps/centralised-indexer): initial baseline

eli-lim:
Scaffolds a solution for #1150 

- A new app - `CentralisedIndexer` - essentially an ETL worker 
- This indexer will be composed of many workers, e.g. Dex Swap Indexing worker - whose only responsibility is to index data it cares about. This indexed data is then exposed by a separate multi-instance read-only service.

#### Additional comments?:
- This PR only contains a simple rpc block poller (similar to whale's) to persist blocks into a local DynamoDB container.
- This PR demonstrates:
    - how dynamodb key schemas and indexes are created via Dynamoose ORM
    - how dynamodb read / write will be performed via Dynamoose ORM
    - how the code of the centralised indexer (etl worker) will be structured
- After this is merged, will begin work on the dex swap indexing
- To support various _globally-sorted-list_ queries, we create a global secondary index with a **fixed** partition key and a sort key. For those coming from the SQL world like me, this is the most reliable (eventually consistent) way to support queries that are akin to SQL queries like `SELECT MAX(height) FROM block`.

## Follow-up PRs
For the app to be production ready (but leaving them for future PRs so this one doesn't become excessively complex):
- [x] block invalidation (triggering deletes on the tables when a block is invalidated) -> #1218
- [ ] indexing `status` tracking for exception handling / recovery
- [ ] service to expose the dex swap data


eli-lim:
Some dev notes:
- Fixed tests depending on dynamodb container failing due to aws credentials missing: solution is to hard-code some dummy values in the tests. Tests were passing on my mac because aws-sdk was reading them from disk in `~/.aws/credentials`
- Removed hard-coded dynamodb container ports to support concurrent testing

canonbrother (commented):
should we standardize the design by following the ocean-api design?? 
eg: defid == blockchaincpp

apps/centralised-indexer/src/indexer/index.ts
```diff
@@ -0,0 +1,56 @@
+import { IndexerModule } from './IndexerModule'
+import { FastifyAdapter, NestFastifyApplication } from '@nestjs/platform-fastify'
+import { NestFactory } from '@nestjs/core'
+import { ConfigService } from '@nestjs/config'
+
+export class RootServer {
```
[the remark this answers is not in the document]
eli-lim: this is the entry point of the app - follows the current pattern in all the other apps

apps/centralised-indexer/src/indexer/IndexerModule.ts
```diff
@@ -0,0 +1,56 @@
...
+@Module({
+  imports: [
+    ConfigModule.forRoot({
+      isGlobal: true,
+      load: [AppConfiguration]
+    }),
+    ActuatorModule,
+    ScheduleModule.forRoot(),
+    BlockchainCppModule,
+    IndexerModule,
```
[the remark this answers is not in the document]
eli-lim: That was a mistake due to renaming - thanks for the catch
I'm surprised it works too :P

apps/centralised-indexer/src/indexer/RootIndexer.ts
```diff
@@ -0,0 +1,39 @@
...
+export class RootIndexer implements Indexer {
+  private readonly indexers: Indexer[]
+
+  constructor (
+    private readonly blockIndexer: BlockIndexer,
+    private readonly dexSwapIndexer: DexSwapIndexer
+  ) {
+    this.indexers = [
+      blockIndexer,
+      dexSwapIndexer
```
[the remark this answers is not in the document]
eli-lim: > Probably make these 2 have hierarchy instead of same level. `blockIndexer` is story some raw with some some fundamental index. `transactionIndexer` will then derived into `dexSwapIndexer` etc

Sorry, what do you mean by hierarchy?

> Note that `TransactionIndexer.index(transaction)` (instead of block) could be better idea, since we have to process the data transaction by following the `n` chronological order as much as possible.

Ah yes - that's in a separate PR (#1203). Basically to have a `MainTransactionIndexer` similar to Whale?
I didn't include any `Transaction` stuff in this PR to keep it smaller and easier to review 😆

eli-lim:
Updated to include a working implementation of dex swap indexing.

However still have the following required for the app to be production ready (but leaving them for future PRs so this one doesn't become excessively complex):
- [] block invalidation (triggering deletes on the tables)
- [] indexing `status` tracking for exception handling / recovery

apps/centralised-indexer/testing/containers/DynamoDbContainer.ts
```diff
@@ -0,0 +1,134 @@
+import Dockerode, { ContainerInfo, DockerOptions } from 'dockerode'
+import { DockerContainer } from '@defichain/testcontainers'
```
cwkang1998, line 2: Instead of exposing`DockerContainer` from `@defichain/testcontiners` , would moving this into `@defichain/testcontainer` instead be a better call?

apps/centralised-indexer/testing/CentralisedIndexerTesting.ts
```diff
@@ -0,0 +1,89 @@
+import { Testing, TestingGroup } from '@defichain/jellyfish-testing'
+import { MasterNodeRegTestContainer } from '@defichain/testcontainers'
+import { ApiClient } from '@defichain/jellyfish-api-core'
+import { NestFastifyApplication } from '@nestjs/platform-fastify'
+import { CentralisedIndexerStubServer } from './CentralisedIndexerStubServer'
+import { DynamoDbContainer } from './containers/DynamoDbContainer'
+
+/**
+ * OceanApi Testing framework.
```
cwkang1998, line 9: Might want to change this to reflect class.

apps/package.json
```diff
@@ -19,8 +19,8 @@
     "@defichain/jellyfish-api-jsonrpc": "^0.0.0",
     "@defichain/jellyfish-testing": "^0.0.0",
     "@defichain/ocean-api-client": "^0.0.0",
-    "@defichain/playground-api-client": "^0.0.0",
```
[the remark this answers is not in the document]
eli-lim: Thanks for reviewing!

And thanks for catching - removed it mistakenly during a rebase

apps/centralised-indexer/testing/containers/DynamoDbContainer.ts
eli-lim: Hmm don't think DynamoDB container should be an offering of the core package, since it's for a single specific app's use-case

apps/centralised-indexer/testing/CentralisedIndexerTesting.ts
eli-lim: Thanks for catching!

apps/centralised-indexer/src/models/dftx/Token.ts
```diff
@@ -0,0 +1,156 @@
...
+  id: {
+    type: Number,
+    hashKey: true
+  },
+  symbol: String,
+  name: String,
+  decimal: Number,
+  limit: String,
+  isDAT: Boolean,
+  isLPS: Boolean,
```
canonbrother, line 44:
```suggestion
  isLPS: Boolean,
  isLoanToken: Boolean,
```
i think we need `isLoanToken` property as well

apps/centralised-indexer/__tests__/model/DexSwapService.test.ts
```diff
@@ -0,0 +1,68 @@
...
+    toSymbol: 'DFI',
+    block: {
+      hash: '',
+      height: 0,
+      medianTime: 0,
+      time: 0
+    }
+  }
+  await dexSwapService.upsert(dexSwap)
+  await dexSwapService.delete('aaa124')
```
canonbrother, line 66: here needs to prove dexSwap is not undefined first then delete

ivan-zynesis (commented):
Shall we break this down to smaller PR, the dynamo module have quite a bit to be tested.

eli-lim:
> Shall we break this down to smaller PR, the dynamo module have quite a bit to be tested.

Yea lemme do that

PR #1212, opened 2022-03-15
chore(package.json): add subpath export for typescript

fuxingloh:

Experiment with subpath exports for `jellyfish-api-core`, new in typescript 4.5, this allows imports of category rpc without using declaring `@defichain/jellyfish-api-core/dist/*` -> `@defichain/jellyfish-api-core/*`.

### DIFF
--- a/packages/jellyfish-api-core/package.json
+++ b/packages/jellyfish-api-core/package.json
@@ -18,6 +18,10 @@
   "files": [
     "dist"
   ],
+  "exports": {
+    ".": "./dist/index.js",
+    "./category/*": "./dist/category/*.js"
+  },
   "scripts": {
     "build": "tsc -b ./tsconfig.build.json"
   },


kodemon:
Will do some local testing of this PR to see how this works before approving, looks neat!

fuxingloh:
> Will do some local testing of this PR to see how this works before approving, looks neat!

If this works flawlessly, we need to update typescript to 4.5 and update the config as well in another PR.

kodemon:
@fuxingloh I went through the release notes for 4.5 and couldn't find mention of node exports support _(which I believe is what we call subpaths?)_

Then did a quick search and came over this ticket

 - https://github.com/microsoft/TypeScript/issues/33079
 
 Which through reading seems to indicate that there is still no stable support for this functionality in TS yet.

If I am looking in the wrong place, could you point me in the right direction?

PR #1221, opened 2022-03-16
refactor(packages/jellyfish-wallet): replace deprecated DeFiAddress.from() in WalletAccount

fuxingloh:

`DeFiAddress.from()` in `packages/jellyfish-address` has been deprecated for some time now. This PR replace the use of it in `WalletAccount.addressToScript(address: string)` with the latest version.

### DIFF
--- a/packages/jellyfish-wallet/src/wallet_account.ts
+++ b/packages/jellyfish-wallet/src/wallet_account.ts
@@ -2,7 +2,7 @@ import { OP_CODES, Script, Transaction, TransactionSegWit, Vout } from '@deficha
 import { WalletEllipticPair } from './wallet_elliptic_pair'
 import { Bech32, HASH160 } from '@defichain/jellyfish-crypto'
 import { Network } from '@defichain/jellyfish-network'
-import { DeFiAddress } from '@defichain/jellyfish-address'
+import { fromAddress } from '@defichain/jellyfish-address'
 
 /**
  * An HDW is organized as several 'accounts'.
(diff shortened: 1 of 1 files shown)


kodemon (approved):

ivan-zynesis (approved):

packages/jellyfish-wallet/src/wallet_account.ts
```diff
@@ -45,10 +45,14 @@ export abstract class WalletAccount implements WalletEllipticPair {
    *
    * @param {string} address to parse into script
    * @return {Script} parsed from address
+   * @throws {Error} if address or network is invalid
    */
   addressToScript (address: string): Script {
-    const parsed = DeFiAddress.from(this.network.name, address)
-    return parsed.getScript()
+    const decodedAddress = fromAddress(address, this.network.name)
+    if (decodedAddress?.script === undefined) {
```
cwkang1998, line 52: This can possibly be simplified.
```suggestion
    if (!decodedAddress?.script) {
```
I believe from implementation, it can only be `undefined` or `DecodedAddress`, thus there should be no issue with falsy error when comparing.

Another more explicit way.
```suggestion
    if (decodedAddress? .script) {
        return decodedAddress.script
    }
    throw new Error('address is invalid')
```
Just preferences, not a must.

cwkang1998 (approved):

packages/jellyfish-wallet/src/wallet_account.ts
fuxingloh: Actually, this is enforced by the linting engine, explicit over implicit, and no implicit coercion. Because this synthetic sugar provides somewhat of a greater DX - it can introduce ambiguity and unnecessary confusion in a large codebase.

fuxingloh: Also on the topic of why I prefer returning errors early over returning objects. It's another explicit DX where you move explicit errors up to terminate early while leaving one true condition that it can allow the function to complete. 

This makes the code base more manageable and maintainable as you allow more injection of errors checking through adding more early termination. See this from a code complexity standpoint and long-term maintainability.

cwkang1998: Ah got it. Thanks for the explaination.

PR #1222, opened 2022-03-16
feat(packages/jellyfish-wallet): expose deriveNode and derivePath to allow better abstraction of Wallet

fuxingloh:

Node and path derivation were previously embedded within `wallet.get()` routine. This makes it difficult for downstream implementation to use Wallet as an Abstraction to build custom patterns. This PR exposes those two functions as separate methods.

This PR also updates all wallet class parameters from `private readonly` to `protected readonly` for the same effect.

### DIFF
--- a/packages/jellyfish-wallet/src/wallet.ts
+++ b/packages/jellyfish-wallet/src/wallet.ts
@@ -31,21 +31,37 @@ export class JellyfishWallet<Account extends WalletAccount, HdNode extends Walle
    * @param {number} [purpose=0] PURPOSE_LIGHT_WALLET
    */
   constructor (
-    private readonly nodeProvider: WalletHdNodeProvider<HdNode>,
-    private readonly accountProvider: WalletAccountProvider<Account>,
-    private readonly coinType: number = JellyfishWallet.COIN_TYPE_DFI,
-    private readonly purpose: number = JellyfishWallet.PURPOSE_LIGHT_WALLET
+    protected readonly nodeProvider: WalletHdNodeProvider<HdNode>,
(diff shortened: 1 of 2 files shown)


packages/jellyfish-wallet/src/wallet.ts
```diff
@@ -31,23 +31,39 @@ export class JellyfishWallet<Account extends WalletAccount, HdNode extends Walle
...
+  deriveNode (account: number): HdNode {
+    const path = this.derivePath(account)
+    return this.nodeProvider.derive(path)
+  }
+
+  /**
+   * @param {number} account number to get
+   * @return {string} HD path
+   */
+  derivePath (account: number): string {
```
ivan-zynesis, line 63:
```suggestion
  derivePath (account: number, purpose?: number): string {
```

Should we?

packages/jellyfish-wallet/src/wallet.ts
```diff
@@ -31,23 +31,39 @@ export class JellyfishWallet<Account extends WalletAccount, HdNode extends Walle
...
+    return `${this.coinType}/${this.purpose}/0/${account}`
```
ivan-zynesis, line 64:
```suggestion
  derivePath (account: number, something?: number): string {
    return `${this.coinType}/${this.purpose}/${something}/${account}`
```

Should we?

fuxingloh: Probably not so that we don't deviate from the standard set by this internal wallet design. The idea is more towards exposing the path for other cosmetics or UI use.

ivan-zynesis (approved):

cwkang1998 (approved):

PR #1225, opened 2022-03-16
feat(packages/jellyfish-wallet-mnemonic): validate 24 words optionally

fuxingloh:

While BIP39 standard doesn't enforce validation of words to generate the HD seed, this implementation optionally allows you to validate the BIP39 word list before generating the MnemonicProviderData.

```ts
/**
 * @param {boolean} [validate=false] optionally validate mnemonic words. While BIP39 standard doesn't enforce
 * validation of words to generate the HD seed, this implementation optionally allow you to validate the BIP39 
 * word list before generating the MnemonicProviderData.
 *
 * @throws {Error} if mnemonic sentence checksum invalid, if validate=true
 */
static wordsToData (words: string[], options: Bip32Options, validate: boolean = false): MnemonicProviderData {
  if (validate && !validateMnemonicSentence(words)) {
    throw new Error('mnemonic sentence checksum invalid')
  }
  // ...
}
```

### DIFF
--- a/packages/jellyfish-wallet-mnemonic/src/hd_node.ts
+++ b/packages/jellyfish-wallet-mnemonic/src/hd_node.ts
@@ -4,7 +4,7 @@ import { WalletHdNode, WalletHdNodeProvider } from '@defichain/jellyfish-wallet'
 import { DERSignature } from '@defichain/jellyfish-crypto'
 import { SIGHASH, Transaction, TransactionSegWit, Vout } from '@defichain/jellyfish-transaction'
 import { TransactionSigner } from '@defichain/jellyfish-transaction-signature'
-import { generateMnemonicWords, mnemonicToSeed } from './mnemonic'
+import { generateMnemonicWords, mnemonicToSeed, validateMnemonicSentence } from './mnemonic'
 
 /**
  * Bip32 Options, version bytes and WIF format. Unique to each chain.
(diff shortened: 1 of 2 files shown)


packages/jellyfish-wallet-mnemonic/src/hd_node.ts
```diff
@@ -136,22 +136,36 @@ export class MnemonicHdNodeProvider implements WalletHdNodeProvider<MnemonicHdNo
...
   }
 
   /**
    * @param {string[]} words to convert into MnemonicProviderData
    * @param {Bip32Options} options
+   * @param {boolean} [validate=false] optionally validate mnemonic words. While BIP39 standard doesn't enforce
+   * validation of words to generate the HD seed, this implementation optionally allow you to validate the BIP39 word
+   * list before generating the MnemonicProviderData.
    * @return MnemonicProviderData
+   * @throws {Error} if mnemonic sentence checksum invalid, if validate=true
```
cwkang1998, line 156: It's out of the scope of this PR, but should we create custom errors  instead of relying on the base error class? Something like

```typescript
class MnemonicValidationError extends Error {
    message = "mnemonic sentence checksum invalid"
}
```

so we can use it in later error handling with something like:

```typescript
try {
 // ... do something ...
}catch(err){
    if(err instanceof MnemonicValidationError){
        //  handle it somehow
    }
}
```
```typescript
expect(() => {
        MnemonicHdNodeProvider.fromWords(words, regTestBip32Options, false)
      }).not.toThrow(MnemonicValidationError)
```
This is probably something that needs a lot of discussion from the team and is more of a change in convention than something related to this PR.

cwkang1998 (approved):

surangap (approved):

canonbrother (approved):

PR #1236, opened 2022-03-18
feat(apps/dex-swap-indexer): scaffold

eli-lim:
Scaffolds a solution for #1150 

- A new app - `DexSwapIndexer` - essentially an ETL worker 
- This indexer will be composed of many workers, e.g. Block indexer, Create Token indexer, Dex Swap indexer - whose only responsibility is to index data it cares about. This indexed data is then exposed by a separate multi-instance read-only service.

#### Additional comments?:
- This PR only contains the scaffolding - testing framework and `RootServer` boilerplate


canonbrother (dismissed):

eli-lim:
@canonbrother just renamed some folders and classes from `centralised-indexer` to `dexswap-indexer` to scope down

canonbrother (approved):

PR #1241, opened 2022-03-21
chore(whale-api-client): update from latest whale changes

kodemon:
#### What this PR does / why we need it:

Whale migration is reliant on the latest changes to `whale-api-client`. To keep this update separate from `apps/whale-api` migration we introduce these updates in a smaller PR.

#### Additional comments?:

Required updates for upcoming updates to PR:

 - https://github.com/DeFiCh/jellyfish/pull/1168

PR #1239, opened 2022-03-21
chore(apps/whale-api): database module

kodemon:
#### What this PR does / why we need it:

Adds database module for apps/whale-api as part of larger consolidation effort of whale into jellyfish.

#### Which issue(s) does this PR address?:

 - https://github.com/DeFiCh/jellyfish/issues/977

#### Additional comments?:

Depends on PR:

 - https://github.com/DeFiCh/jellyfish/pull/1201
   - _This PR will stay in draft mode until parent branch is merged to master._

An ongoing effort to split up the larger `apps/whale-api` PR into smaller digestible pieces while providing incremental testing:

 - https://github.com/DeFiCh/jellyfish/pull/1168

PR #1201, opened 2022-03-11
chore(apps/whale-api): base structure

canonbrother:
#### Which issue(s) does this PR fixes?:

Fixes #

#### Additional comments?:
splited from https://github.com/DeFiCh/jellyfish/pull/1168

PR #1167, opened 2022-03-04
feat(whale): migrate `packages/whale-api-wallet`

kodemon:
#### What this PR does / why we need it:

Migrate `whale-api-wallet` from whale to jellyfish repo in consolidation efforts of the whale api.

#### Which issue(s) does this PR fixes?:

Partial #977

#### Additional comments?:

This PR follows up which is a precursor and dependency of this package:

 - https://github.com/DeFiCh/jellyfish/pull/1166
   - Wallet has a direct dependency on `whale-api-client` so this trails after that.

This PR will be in draft until #1166 is merged to main and will be followed up by:

 - https://github.com/DeFiCh/jellyfish/pull/1168

PR #1248, opened 2022-03-22
chore(whale-api): migrate whale to apps/whale

### LINKED ISSUE #977: `/apps/whale-api` - Consolidation & Migration

kodemon:

Final part of whale migration to jellyfish moving [`whale/src`](https://github.com/DeFiCh/whale/tree/main/src) to `apps/whale`.

Fixes #977

#### Additional comments?:

Please note the following required changes to make this migration work. While this PR aims to change as little as possible to avoid exhaustive review requirements some differences between whale and jellyfish requires changes to be made. I will make an effort here to point out changes that required these minor changes to occur.

First change is to update all imports pointing to:

 - `@src` => `../relative/path`
 - `@whale-api-client` => `@defichain/whale-api-client`
 - `@defichain/*/dist` => `@defichain/*/src`

This is done through a migration script I quickly wrote so that we could quickly port any ongoing updates to whale into jellyfish without having to go file by file each time. You can see the migration script used [here](https://gist.github.com/kodemon/81da53ea3bd69654c64a9468567862e9).

This change is needed because whale was doing its own internal pathing as [seen here](https://github.com/DeFiCh/whale/blob/main/tsconfig.json#L19-L20). This is an issue because you can't do these sort of sub pathing in individual application projects. While it can technically work from a typescript build perspective, it starts conflicting with the pathing requirements in our [jest setup](https://github.com/DeFiCh/jellyfish/blob/main/jest.config.js). Trying to consolidate the differences between whale and jellyfish in this manner would be too cumbersome and require us to make root jellyfish changes to adhere to a single application package. So the optimal solution here is simply to just fix the imports, which is done in this PR.

The second change that has occurred in this PR is rooted in the different typescript rules between jellyfish and whale. In jellyfish all `catch (err)` casts `err` as `unknown`. In whale `err` is `any`, this requirement forces us to cast all the errors caught in whale to be cast to a correct error instance. The following files has been modified for this requirement.

 - module.api/loan.collateral.controller.e2e.ts:218:14
 - module.api/loan.scheme.controller.e2e.ts:138:14
 - module.api/loan.token.controller.e2e.ts:187:14
 - module.api/loan.vault.controller.e2e.ts:130:14
 - module.api/loan.vault.controller.e2e.ts:141:14
 - module.api/masternode.controller.e2e.ts:97:14
 - module.api/poolpair.controller.e2e.ts:270:14
 - module.api/poolpair.service.ts:64:11
 - module.api/poolpair.service.ts:75:11
 - module.api/rawtx.controller.e2e.ts:58:14
 - module.api/rawtx.controller.e2e.ts:76:14
 - module.api/rawtx.controller.e2e.ts:122:14
 - module.api/rawtx.controller.e2e.ts:140:14
 - module.api/rawtx.controller.ts:58:11
 - module.api/rawtx.controller.ts:61:11
 - module.api/rawtx.controller.ts:66:40
 - module.api/rawtx.controller.ts:85:11
 - module.api/rawtx.controller.ts:88:11
 - module.api/rawtx.controller.ts:92:40
 - module.api/transaction.controller.e2e.ts:88:14
 - module.database/provider.level/level.database.ts:77:11
 - module.indexer/rpc.block.provider.ts:86:11

I considered changing the typescript rule to cast errors back to `any`, but since that would be specifically to fix whale I decided against this. It's better to do these very minor updates to whale rather than changing all our jellyfish approaches to adhere to whale which is being deprecated.

Third change was fixing JSON import in the `app.module.ts` as apps are now sharing the same package.json file.

 - https://github.com/DeFiCh/whale/blob/main/src/app.module.ts#L1

---

That should conclude the changes made to the original whale `src` code. Outside of this nothing else has been modified and is a 1:1 representation of the original whale implementation. This should ensure that we do not need to review the code, but simply ensure that it builds and tests reliably. And any future refactoring and restructuring efforts can be delegated as tasks within jellyfish going forward.

---

Non `src` changes:

 - https://github.com/DeFiCh/jellyfish/pull/1248/files#diff-b55cdbef4907b7045f32cc5360d48d262cca5f94062e353089f189f4460039e0R13
   - To support functionality that resolves `this` in class variable assignments, whale uses es2019 but I see no reason not to bump this to `es2020` since we need to change it regardless.
  - https://github.com/DeFiCh/jellyfish/pull/1248/files#diff-b55cdbef4907b7045f32cc5360d48d262cca5f94062e353089f189f4460039e0R18
    - To support json imports used in whale for package versioning.
  - https://github.com/DeFiCh/jellyfish/pull/1248/files#diff-1e058ca1442e46581b13571fb8d261f9e1f5657e26c96634d4c1072f0f0347f1R6
    - To support previous /dist targets to /src which is used extensively in whale.
      - _Note that this was not actually needed in #1005 but seems to break with the latest main and whale src. So this additional rule was added to fix pathing errors in jest._

This PR is a updated version of:

 - https://github.com/DeFiCh/jellyfish/pull/1005

This PR outdates the following PRs:

_These PRs can be made into refactoring tickets post migration_

 - https://github.com/DeFiCh/jellyfish/pull/1201
 - https://github.com/DeFiCh/jellyfish/pull/1239

This PR builds on the following PRs:

 - https://github.com/DeFiCh/jellyfish/pull/1167
 - https://github.com/DeFiCh/jellyfish/pull/1170
 - https://github.com/DeFiCh/jellyfish/pull/1241

This PR is followed up by:

 - https://github.com/DeFiCh/jellyfish/pull/1250


eli-lim:
thanks for the detailed write-up! will do some sanity checks:
- [ ] tests pass locally
- [ ] contents are indeed the same, apart from the minor changes mentioned

eli-lim (changes requested):
A minor suggestion:
1. rename `apps/whale-api` to `apps/whale` as it encompasses more than just the api layer

Needs syncing up:
- poolpair.service.ts

Some notes:
- the relative imports can be shortened in a subsequent PR
- just FYI - building and running whale's entrypoint main.js throws: `node:internal/modules/cjs/loader:936 ; Error: Cannot find module '@defichain/jellyfish-transaction/src/script/_buffer'`, also to be fixed in a subsequent PR

eli-lim:
# Checks

### 1. Tests all pass
| whale $ `jest src` | jellyfish $ `jest whale-api` |
| - | - |
|image|image|

### 2. Same tests
image
> No diff in jest --listTests

### 3. `src/` content
- Manually diffed `jellyfish/apps/whale-api/src` against `whale/src` locally with intellij [diff viewer](https://www.jetbrains.com/help/webstorm/comparing-files-and-folders.html#cd16dbb0), and can confirm that only (1) imports are changed and (2) error type assertions are added
- only exception is `poolpairs.service.ts`, which needs syncing with whale's latest changes

Nice! @kodemon

ivan-zynesis:
We need to bump the CI test instance count for this soon.

kodemon:
whale-api-client unit tests from #1005 has been moved to follow up PR and placed in its original location from where they originated in whale repository.

 - https://github.com/DeFiCh/jellyfish/pull/1250

kodemon:
> A minor suggestion:
> 
> 1. rename `apps/whale-api` to `apps/whale` as it encompasses more than just the api layer
> 
> Needs syncing up:
> 
> * poolpair.service.ts
> 
> Some notes:
> 
> * the relative imports can be shortened in a subsequent PR
> * just FYI - building and running whale's entrypoint main.js throws: `node:internal/modules/cjs/loader:936 ; Error: Cannot find module '@defichain/jellyfish-transaction/src/script/_buffer'`, also to be fixed in a subsequent PR

Updated as per your suggestions, thanks for the review efforts thus far 👍🏻

eli-lim (approved):
LGTM!

Some follow-ups apart from the styling / convention stuff:
- the relative imports can be shortened
- building and running whale's entrypoint main.js throws: node:internal/modules/cjs/loader:936 ; Error: Cannot find module '@defichain/jellyfish-transaction/src/script/_buffer'
- jest tests currently also get included into the build dir `apps/dist`

ivan-zynesis (approved):

canonbrother (approved):

PR #1249, opened 2022-03-22
chore(packages/testcontainer): migration from dockerrode to testcontainers

### LINKED ISSUE #887: Move `@defichain/testcontainers` to use npm testcontainers exclusively

The current `@defichain/testcontainers` implementation uses `Dockerode` for Docker API interfacing. We should move to use npm [testcontainers](https://www.npmjs.com/package/testcontainers).

#### Why is this needed:

Tap onto battle-tested testcontainers ecosystem design from the open source community.

/triage accepted
/area testcontainers

cwkang1998:

This PR aims to migrate the current dockerrode implementation of `package/testcontainers` to using npm testcontainers.

Closes #887

#### Additional comments?:
There are some specific constraints put forth by npm testcontainers' maintainers, which conflicts with some of our previous implementation and design choices of `packages/testcontainers`. As such, there are some functionality that might still rely a bit on the use of dockerrode, although the author of this PR tries to avoid it as much as possible.

### DIFF
--- a/packages/testcontainers/src/containers/DeFiDContainer.ts
+++ b/packages/testcontainers/src/containers/DeFiDContainer.ts
@@ -1,7 +1,7 @@
 import fetch from 'cross-fetch'
 import { DockerContainer, DockerOptions } from './DockerContainer'
 import { waitForCondition } from '../utils'
-import Dockerode from 'dockerode'
+import { dockerClient } from 'testcontainers/dist/docker/docker-client'
 /**
  * Types of network as per https://github.com/DeFiCh/ain/blob/bc231241/src/chainparams.cpp#L825-L836
  */
@@ -234,8 +234,8 @@ export abstract class DeFiDContainer extends DockerContainer {
    * This will stop the container and start it again with old data intact.
    * @param {number} [timeout=30000] in millis
    */
-  async restart (timeout: number = 60000): Promise<void> {
-    const dockerrode = new Dockerode()
+  async restart (timeout: number = 30000): Promise<void> {
+    const dockerrode = (await dockerClient).dockerode
     const dockerrodeContainer = dockerrode.getContainer(this.requireContainer().getId())
     await dockerrodeContainer.restart()
     await this.waitForRpc(timeout)
--- a/packages/testcontainers/src/containers/DeFiDContainer.ts
+++ b/packages/testcontainers/src/containers/DeFiDContainer.ts
@@ -1,7 +1,9 @@
+import { dockerClient } from 'testcontainers/dist/docker/docker-client'
+import { inspectContainer } from 'testcontainers/dist/docker/functions/container/inspect-container'
+import { BoundPorts } from 'testcontainers/dist/bound-ports'
 import fetch from 'cross-fetch'
 import { DockerContainer, DockerOptions } from './DockerContainer'
 import { waitForCondition } from '../utils'
-import { dockerClient } from 'testcontainers/dist/docker/docker-client'
 /**
  * Types of network as per https://github.com/DeFiCh/ain/blob/bc231241/src/chainparams.cpp#L825-L836
  */
@@ -238,6 +240,13 @@ export abstract class DeFiDContainer extends DockerContainer {
     const dockerrode = (await dockerClient).dockerode
     const dockerrodeContainer = dockerrode.getContainer(this.requireContainer().getId())
     await dockerrodeContainer.restart()
+    // Hacky solution to go pass changing ports after restart
+    // And since testcontainers doesn't update, we do it here
+    const container: any = this.requireContainer()
+    const inspectRes = await inspectContainer(dockerrodeContainer)
+    const boundPorts: BoundPorts = (BoundPorts as any).fromInspectResult(inspectRes).filter(DeFiDContainer.DefiDPorts[this.network])
+    container.inspectResult = inspectRes
+    container.boundPorts = boundPorts
     await this.waitForRpc(timeout)
   }
 }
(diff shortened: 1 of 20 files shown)


packages/testcontainers/src/containers/DeFiDContainer.ts
```diff
@@ -246,7 +235,9 @@ export abstract class DeFiDContainer extends DockerContainer {
    * @param {number} [timeout=30000] in millis
    */
   async restart (timeout: number = 30000): Promise<void> {
-    await this.container?.restart()
+    const dockerrode = (await dockerClient).dockerode
+    const dockerrodeContainer = dockerrode.getContainer(this.requireContainer().getId())
+    await dockerrodeContainer.restart()
```
canonbrother, line 240: seems you're trying use `dockerode` attach the running docker to execute `restart` since testcontainer doesn't support `restart` <-- gosh!!

canonbrother: and seems we get advised to use `stop and start` instead of `restart` in testcontainer
https://github.com/testcontainers/testcontainers-node/issues/83#issuecomment-638379724

cwkang1998: Yes I saw this, but for this case, where its setting the `defi.conf` file for masternode, the restart is required to retain state inside the container, and stopping and then starting it would cause it to lose all state. I also tried working with stopping and starting with dockerrode instead of testcontainer, but for some reason it doesn't work.

cwkang1998: Yes I saw this, but for this case, where its setting the `defi.conf` file for masternode, the restart is required to retain state inside the container, and stopping and then starting it would cause it to lose all state. I also tried working with stopping and starting with dockerrode instead of testcontainer, but for some reason it doesn't work.

canonbrother: hmm... perhaps we need docker volume to retain the state??

cwkang1998: default api provided by `testcontainer-node` does not support volumnes, but does support bind mount and tmpfs. I think if we need access to volumes we'll have to use the docker-compose apis provided by them, which needs a docker-compose file to be provided.

canonbrother: hmm... do you think `testcontainer` suit our case?? i feel its a little bit `workaround`.. :thinking: 
cc @fuxingloh

cwkang1998:
Looking at the code from both `testcontainers-node` and `dockerrode`, the current restart code **should** work, but its causing waitForRpc to timeout, which indicates it does not work, and by ssh-ing into the container, it shows that the container does restart, but defid somehow does not restart with it (?). 

However, the previous dockerrode based code works as expected, and looking at the implementation for testcontainers, it doesn't seem like there's a lot of difference handling the state of the container (unless i missed something)

Below is the test output for reference:
```bash
 FAIL  packages/testcontainers/__tests__/containers/ContainerRestart.test.ts (111.307 s)
  container restart with setDeFiConf
    ✕ should create a masternode and mint some blocks with setDeFiConf masternode_operator= (51791 ms)
  container group restart
    ✕ test container group restart (31143 ms)

  ● container restart with setDeFiConf › should create a masternode and mint some blocks with setDeFiConf masternode_operator=

    waitForRpc is not ready within given timeout of 30000ms.

      16 |         resolve()
      17 |       } else if (expiredAt < Date.now()) {
    > 18 |         reject(new Error(`${message} is not ready within given timeout of ${timeout}ms.`))
         |                ^
      19 |       } else {
      20 |         setTimeout(() => void checkCondition(), interval)
      21 |       }

      at checkCondition (packages/testcontainers/src/utils.ts:18:16)
          at runMicrotasks (<anonymous>)

  ● container group restart › test container group restart

    waitForRpc is not ready within given timeout of 30000ms.

      16 |         resolve()
      17 |       } else if (expiredAt < Date.now()) {
    > 18 |         reject(new Error(`${message} is not ready within given timeout of ${timeout}ms.`))
         |                ^
      19 |       } else {
      20 |         setTimeout(() => void checkCondition(), interval)
      21 |       }

      at checkCondition (packages/testcontainers/src/utils.ts:18:16)
          at runMicrotasks (<anonymous>)
```

Since `container.restart()` api is not provided, I've tried to use both Dockerrode directly or use the testcontainers Dockerrode instance for this, with both ending up in failure. Stopping and restarting simply does not work since testcontainers' stop and start method removes the stopped container. As for doing it using dockerrode, it yields the same result as restarting so far.

Another method I've done is through using the command line interface via exec, to call `defi-cli stop` and then starting `defid` again, but that also does not seem to work, possibly due to permissions or users based configuration. I did not investigate very deeply for this but instead opted for docker based approach since this seems too much of a workaround, but its a method that might be able to work.

ivan-zynesis:
I think the implementation is offed track from the original motivation (from my understanding).

The purpose is to allow easier bridging between containers (GenericContainer) thus any downstream project will be able to setup complicated test env by mix and match multiple containers

without doing this - https://github.com/DeFiCh/oracle-adapters/blob/main/packages/salmon-testing/src/WhaleMasternodeRegTestContainer.ts
able to easily setup - https://github.com/DeFiCh/wallet/blob/main/docker-compose.yml within unit test

cwkang1998:
Managed to get restart working, main issue was that after restart using dockerrode, the container would reassign external ports. As such extra effort needs to be done to force a refresh or update of the cached ports for the container.

PR #1250, opened 2022-03-23
chore(whale-api-client): migrate unit tests

kodemon:

Moving the original unit tests from whale-api-client to jellyfish. This is done as a follow up to whale migration since the tests has a direct dependency on whale app code for the client stub. This is why we had to wait until the application layer was migrated before we could re-introduce these tests.

#### Additional comments?:

Test imports has been updated to support the new structure in jellyfish.

Depends on revert wrapper PR:

 - https://github.com/DeFiCh/jellyfish/pull/1284

This is a follow up to whale migration PR:

 - https://github.com/DeFiCh/jellyfish/pull/1248
   - Has a direct dependency on `apps/whale`

### DIFF
--- a/packages/whale-api-client/__tests__/api/account.history.test.ts
+++ b/packages/whale-api-client/__tests__/api/account.history.test.ts
@@ -3,6 +3,7 @@ import { Testing } from '@defichain/jellyfish-testing'
 import { StubWhaleApiClient } from '../stub.client'
 import { StubService } from '../stub.service'
 import { WhaleApiClient, WhaleApiException } from '../../src'
+import { RegTestFoundationKeys } from '@defichain/jellyfish-network'
 
 let container: MasterNodeRegTestContainer
 let service: StubService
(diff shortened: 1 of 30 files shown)


canonbrother:
hmm... that's one failed test.. does the changes affect that?? :thinking:

ivan-zynesis:
> hmm... that's one failed test.. does the changes affect that?? 🤔

The composite swap path calculation test failed. It is some heavy computation but 180sec do not really make any sense, could it be the logic itself is outdated (using some old unoptimized version)?

eli-lim:
very strange.. can't replicate the failure locally
this PR doesn't change anything related to `WhaleApiClient` either haha

canonbrother:
ohh.. all passed now.. seems just flaky

hmm.. btw.. preferably we can migrate code through git history commit
https://github.com/JellyfishSDK/jellyfish/pull/1267
a fancy job by @eli-lim

ivan-zynesis (approved):

eli-lim:
> hmm.. btw.. preferably we can migrate code through git history commit #1267 a fancy job by @eli-lim

haha considered this, but whale-api-client was initially migrated without the original git history. So adding the git history will require a few PRs that purge the existing files, adding them in + renaming, etc. - but we prefer to quickly finish up the migration instead 😅

eli-lim (approved):

canonbrother (approved):

PR #1284, opened 2022-03-29
chore(whale-api-client): revert to original whale wrappers

kodemon:

Reverts some initial decisions during migration to adhere to the correct expectations of the original whale api client. This is mainly because of breaking expectations in tests compared to the original version.

Any further refactoring needs to be accompanied through a more complete onset of changes to every whale layer going forwards.

This is however a rather small revert simply adding back in what was removed during previous migration.

#### Additional comments?:

This is a precursor PR for whale api unit test migration:

 - https://github.com/DeFiCh/jellyfish/pull/1250

Original migration PR:

This PR carries over changes made to whale-api-client in whale:

 - https://github.com/DeFiCh/whale/commit/46698bfcf814219af673e709bc817bcc345cbf07#diff-0e10f6cd3807333a55b6653be4aca13ef10ce9ae90b88a542084ef13d234012b

### DIFF
--- a/packages/whale-api-client/src/api/PoolPairs.ts
+++ b/packages/whale-api-client/src/api/PoolPairs.ts
@@ -1,5 +1,5 @@
-import { ApiPagedResponse } from '@defichain/ocean-api-client'
 import { WhaleApiClient } from '../WhaleApiClient'
+import { ApiPagedResponse } from '../WhaleApiResponse'
 
 /**
  * DeFi whale endpoint for poolpair related services.
@@ -93,6 +93,14 @@ export class PoolPairs {
    */
   async getAllPaths (fromTokenId: string, toTokenId: string): Promise<SwapPathsResult> {
     return await this.client.requestData('GET', `poolpairs/paths/from/${fromTokenId}/to/${toTokenId}`)
+  }
+
+  /**
+   * Get all dex prices denominated in a given token
+   * @param {string} [denomination='dUSD'] denomination
+   */
+  async listDexPrices (denomination: string): Promise<DexPricesResult> {
+    return await this.client.requestData('GET', `poolpairs/dexprices?denomination=${denomination}`)
   }
 }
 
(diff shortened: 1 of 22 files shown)


packages/whale-api-client/src/api/PoolPairs.ts
```diff
@@ -229,7 +237,7 @@ export interface BestSwapPathResult {
   fromToken: TokenIdentifier
   toToken: TokenIdentifier
   bestPath: SwapPathPoolPair[]
-  estimatedReturn: string
+  estimatedReturn: string // BigNumber
```
cwkang1998, line 240: For this, we can also do something like
```typescript
type BigNumberString = string
```
just to indicate that its a big number string. Not sure if it fits our current code convention though.

kodemon: Yeah, no sure what the preference here would be. Perhaps something we can spawn a ticket for and discuss. This is a change that is carrying over from latest [whale changes](https://github.com/DeFiCh/whale/commit/46698bfcf814219af673e709bc817bcc345cbf07#diff-0e10f6cd3807333a55b6653be4aca13ef10ce9ae90b88a542084ef13d234012b). 

I will update the PR message reflect this carry over.

ivan-zynesis: I did something similar before like `type Address = string` but could have some annoying glitch depend on lint engine, give `type string not assignable to BigNumberString`

cwkang1998: If there are inconsistencies in linting engine then it'll be definitely less than ideal, but I think this is something we can probably discuss in a ticket and maybe test to see how inconsistent it is, or if its (hopefully) been fixed if its not. 

I think this PR is good to go.

cwkang1998 (approved):

eli-lim (approved):

weiyuan95 (approved):

ivan-zynesis (approved):

canonbrother (approved):

PR #1299, opened 2022-03-30
feat(apps/playground): setgov for enabling payback to all tokens

### LINKED ISSUE #1297: Token Ids are changing on /v0/regtest/tokens?size=200 with mocked playground
#### What happened:

The token ids returned from `/v0/regtest/tokens?size=200` keeps changing when re-running `docker-compose` on light wallet
On the screenshots below:
**e.g DUSD = id: 13 and DUSD= id: 15** also happens in other tokens

```
# docker-compose.yml
whale:0.32.1
playground-api:2.32.1
```
```
jellyfish: 2.32.1
```

The token id should be consistent on mocked playground regardless if `docker-compose` is re-run multiple times

- Rerunning the docker twice then the said API endpoint tokenIds will change

#### Anything else we need to know?:

image
image

### LINKED ISSUE #1298: `bot` to automate setting of GovVar to access new hard fork features
Need more info @fuxingloh

canonbrother:

Fixes #1298
Fixes #1297

#### Additional comments?:


apps/playground-api/src/modules/PlaygroundModule.ts
```diff
@@ -66,8 +65,7 @@ export class PlaygroundModule implements OnApplicationBootstrap {
       loanScheme,
       loanToken,
       loanCollateral,
-      dex,
-      gov
+      dex
```
fuxingloh, line 68: Actually, why do we need it to run after `SetupLoanToken`, rather we can have this run as its' own "Setup" since the setup service run sequentially?

apps/playground-api/src/bots/GovBot.ts
```diff
@@ -0,0 +1,48 @@
...
+import { TokenResult } from '@defichain/jellyfish-api-core/src/category/token'
+import { Injectable } from '@nestjs/common'
+import { Interval } from '@nestjs/schedule'
+
+@Injectable()
+export class GovBot {
+  constructor (readonly client: ApiClient) {
+  }
+
+  @Interval(6000)
```
fuxingloh, line 11:
```suggestion
```

We don't need to run this in intervals?

apps/playground-api/src/bots/GovBot.ts
```diff
@@ -0,0 +1,48 @@
...
+  async run (): Promise<void> {
+    const tokens = await this.client.token.listTokens()
+    for (const id of Object.keys(tokens)) {
+      if (tokens[id].isLoanToken) {
+        await this.enablePayback(id, tokens)
+      }
```
fuxingloh, line 17: I think we should only enable a few scenarios to keep it saner. Easier for @thedoublejay to design different scenarios without this capability too.

apps/playground-api/src/modules/PlaygroundModule.ts
canonbrother: yes that's few approaches.. early was thinking to take tokenId from memory instead of rpc.. but find out its not possible..
yea will move back to `setup` flow

apps/playground-api/src/bots/GovBot.ts
canonbrother: yea we don't that.. moved to setup flow

apps/playground-api/src/bots/GovBot.ts
canonbrother: yea done

apps/playground-api/src/setups/setup.loan.token.ts
```diff
@@ -42,6 +42,7 @@ export class SetupLoanToken extends PlaygroundSetup<SetLoanToken> {
 
   async create (each: SetLoanToken): Promise<void> {
     await this.client.loan.setLoanToken(each)
+    await this.generate(1)
```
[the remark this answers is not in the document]
canonbrother: yes.. its actually just random loan token id issue
because they were generated within a same block, the id will be randomly assigned by miner

random-loan-token-id

fuxingloh (approved):
LGTM!

PR #1300, opened 2022-03-30
chore(whale): add build support

kodemon:

This enables whale to be built successfully

A few changes has to be made to make this work as expected:

 - Changed all `/src` imports to `/dist` imports.
   - The reason for this is because `/src` does not exist after running lerna build, and since the imports themselves drilled into non existent `/src` import errors occurred after build process completed.
   - Now builds run fine with `/dist`, but since `/dist` does not exist during testing we have to perform another work around by pathing any packages with `/dist` import to `/src`. Hence the added bespoke `/dist` targets in `tsconfig` and `jest.config`
 - Changed `src/main` to `src/index`
   - Since docker builds targets the package build src and not a specific file index has to be used for automatic resolution.
   - See [Dockerfile](https://github.com/DeFiCh/jellyfish/blob/main/Dockerfile#L26)

Partially solves 

 - https://github.com/DeFiCh/jellyfish/issues/977

#### Additional comments?:

Follow up for PR: 

 - https://github.com/DeFiCh/jellyfish/pull/1272

The custom pathing for whale `/dist` targets is not ideal but until there is better support for this sort of behavior inherent in jest and typescript this is the only way I see this working for now. Another potential work around would be to refactor how imports are used to remove the `/dist` requirement, such as:

```ts
import { address } from '@defichain/whale-api-client'

// access namespace functionality from exported (* as address)
// address.AddressToken
// address.AddressHistory
```

or change our packages to export without casting

```ts
export * from './api/Address'
// instead of
export * as address from './api/Address'
```

I think until we have a defined way we want to approach this, the following work around should suffice to get the beginning of our whale build steps in jellyfish operational.


eli-lim (dismissed):

apps/tsconfig.json
```diff
@@ -15,6 +15,18 @@
       "@defichain-apps/libs/*": [
         "libs/*/src"
       ],
+      "@defichain/jellyfish-api-core/dist/*": [
```
ivan-zynesis, line 18: would this work? (sometimes the mapper just weird)
```
"@defichain/*/dist/*": [
  "../packages/*/src/*"
]
```

kodemon: typescript only allows for 1 `*` wildcard unfortunately. If you try with multiple you will get a ts error. ☹️

eli-lim (dismissed):
Biggest concern is the mapper stuff which we can refactor after unblocking the build & release process and migrating everyone from whale into jellyfish/apps/whale.

kodemon:
> Biggest concern is the mapper stuff which we can refactor after unblocking the build & release process and migrating everyone from whale into jellyfish/apps/whale.

Yes, to recap some conversation I am having with @eli-lim we are thinking:

 - Merge this PR with mapped workarounds to get builds successfully working.
 - Create release workflow PR to make whale part of jellyfish releases.
 - Officially move all development efforts from whale to jellyfish/whale.
 - Start further refactoring of jellyfish whale, restructuring and fixing imports _(and in turn remove mapping workarounds)_

This should get us fully migrated.

canonbrother (approved):
> Start further refactoring of jellyfish whale, restructuring and fixing imports (and in turn remove mapping workarounds)

this is the only issue for me but since its on plan then :runner:

apps/whale/docker-compose.yml
```diff
@@ -0,0 +1,49 @@
+version: '3.7'
```
ivan-zynesis, line 1: Is this compose for doing test in local?

It is a mixed of production (port) and test (regtest flag) configuration, not sure what's the purpose.

eli-lim: Woopsies - meant to update that
Thanks for the catch

kodemon: yes this is currently purely for local sanity testing https://github.com/DeFiCh/whale/blob/main/.github/workflows/ci/docker-compose.yml

This will most likely be re-worked in upcoming PRs in regards to sanity testing introductions.

eli-lim: And yes, it's meant for local run only like rich-list-api's

eli-lim (approved):

ivan-zynesis (approved):

cwkang1998 (approved):

PR #1301, opened 2022-03-30
feat(apps/whale): improve getsubgraphswaps query perf with inmemory indexer

eli-lim:
- Introduces an inmemory indexer largely following whale's indexer design, to eagerly fetch and cache blocks
- When catching up far from the chain tip (> 10 blocks away), the indexer will run in 'concurrent' mode, where 50 requests are dispatched concurrently to fetch the blocks. When one request resolves, another is dispatched straightaway. The blocks are then enqueued and sorted in memory.
   - This concurrent behaviour allows the indexer to catch up **significantly** faster.
   - Any failed request will be added back into the requests queue.
- The blocks are combed through to look for dexswaps, enqueued into an inmemory queue sorted by timestamp
- Drastically improves performance of `/getsubgraphswaps`

Part of #1024 

#### Additional comments?:
- When the inmemory indexer is catching up, controller will rely on the old logic of brute force search for swaps starting from the chain tip
- Indexer will take over all requests when it's caught up with chain tip

```
[Nest] 26390  - 03/31/2022, 1:05:16 AM     LOG [NestFactory] Starting Nest application...
...
[Nest] 26390  - 03/31/2022, 1:05:17 AM     LOG [RPCBlockProvider:mainnet] Cached block 1752086, hash=15f572483fda92cf442c5f3671747440c8c00ef75bccd8533a6834bff5832e8a. BlockQueue[size=2, blockHeights=[1752086 ... 1752098]]
...
[Nest] 26390  - 03/31/2022, 1:05:41 AM     LOG [DexSwapQueue:mainnet] Removed 6 swaps from cache. Remaining: 4171
[Nest] 26390  - 03/31/2022, 1:05:41 AM     LOG [RPCBlockProvider:mainnet] Cached block 1753094, hash=bcb37b19dcdfb9b8b4355bb0e6826997e6515288f88cc6b2fdd5a1678a9f568f. BlockQueue[size=1003, blockHeights=[1752091 ... 1753094]]
[Nest] 26390  - 03/31/2022, 1:05:41 AM     LOG [RPCBlockProvider:mainnet] Indexer caught up with chain tip!
```

Takes about 20s - 30s to index 1000 blocks and their contained dex swaps.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -35,7 +35,8 @@
       }
     ],
     "import/no-default-export": "error",
-    "no-else-return": "error"
+    "no-else-return": "error",
+    "no-labels": "off"
   },
   "overrides": [
     {
--- a/apps/legacy-api/__tests__/controllers/PoolPairController.test.ts
+++ b/apps/legacy-api/__tests__/controllers/PoolPairController.test.ts
@@ -1,9 +1,13 @@
 import { LegacyApiTesting } from '../../testing/LegacyApiTesting'
 import { PoolPairData } from '@defichain/whale-api-client/src/api/PoolPairs'
+import { verifySwapsOrdering, verifySwapsShape } from '../providers/DexSwapQueue.e2e'
 
 const ONLY_DECIMAL_NUMBER_REGEX = /^[0-9]+(\.[0-9]+)?$/
 
-const apiTesting = LegacyApiTesting.create()
+const apiTesting = LegacyApiTesting.create({
+  mainnetBlockCacheCount: 100,
+  testnetBlockCacheCount: 10
+})
 
 beforeAll(async () => {
   await apiTesting.start()
@@ -217,7 +221,7 @@ it('/v1/listyieldfarming', async () => {
   })
 })
 
-describe('getsubgraphswaps', () => {
+describe('getsubgraphswaps - without waiting for inmemory indexer', () => {
   it('/v1/getsubgraphswaps', async () => {
     const res = await apiTesting.app.inject({
       method: 'GET',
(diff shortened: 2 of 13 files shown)


.eslintrc
```diff
@@ -35,7 +35,8 @@
       }
     ],
     "import/no-default-export": "error",
-    "no-else-return": "error"
+    "no-else-return": "error",
+    "no-labels": "off"
```
fuxingloh, line 39: what's does this change do? A separate PR?

eli-lim: Allows [labels](https://eslint.org/docs/rules/no-labels) to be used

it makes the swap search loop more readable imo

fuxingloh:
Just throwing out my thoughts - I was thinking whether this might be too complex and about whether we can reduce the complexity.

Currently, the implementation is somewhat of a push and pull design both running at the same time. Pulling when the indexer hasn't caught up. Pushing it so that it catches up. This implementation creates cognitive complexity since the implementation is convoluted to understand.

The other way we could do it is having it as "pull" only, where you cache "Most Recent Swap" via a timestamp. Using `txid` as the key since `txid` lookup is the most expensive part of the logic (Block -> Txn in Block -> Vout -> History). You could effectively cut down on the lookup cost by caching "Txn in Block" -> onwards. So we cache last 10k records of the most expensive part?

I am not sure if this makes sense, but might reduce complexity since you remove the "indexer" part of the equation by relying on a pull design when controller load. We can also run the pull logic on a interval to keep caching.

eli-lim:
> The other way we could do it is having it as "pull" only, where you cache "Most Recent Swap" via a timestamp. Using `txid` as the key since `txid` lookup is the most expensive part of the logic (Block -> Txn in Block -> Vout -> History). You could effectively cut down on the lookup cost by caching "Txn in Block" -> onwards. So we cache last 10k records of the most expensive part?
> 
> I am not sure if this makes sense, but might reduce complexity since you remove the "indexer" part of the equation by relying on a pull design when controller load. We can also run the pull logic on a interval to keep caching.

Yea agree that it's somewhat convoluted, although it's following whale's indexer design with the concurrent requests tweak for performance.
Your approach sounds a lot simpler - lemme try it out and see if we can achieve similar performance gain

eli-lim:
Closing in favour of #1309

Issue #1271, opened 2022-03-28
`apps/status-api`: Implement Blockchain Status


joeldavidw:
#### What would you like to be added:

An endpoint (`/blockchain`) to provide the status of the blockchain.

Statuses:
- operational: < 30 mins
- degraded: 30-45 mins
- outage: > 45 mins

/triage accepted
/area apps

PR #1308, opened 2022-04-01
feat(status-api): Status API for Blockchain 

nichellekoh:
This includes the status api to return the status of the blockchain
operational: < 30 mins
degraded: 30-45 mins
outage: > 45 mins

Fixes https://github.com/DeFiCh/jellyfish/issues/1271

### DIFF
--- a/apps/status-api/__test__/controllers/BlockchainController.test.ts
+++ b/apps/status-api/__test__/controllers/BlockchainController.test.ts
@@ -22,7 +22,8 @@ describe('BlockchainController - Status test', () => {
     })
 
     expect(res.statusCode).toStrictEqual(200)
-    expect(res.body).toContain('operational')
+    const body = JSON.parse(res.body)
+    expect(body.status).toStrictEqual('operational')
   })
 
   it('/blockchain/status - should get degraded', async () => {
@@ -36,7 +37,8 @@ describe('BlockchainController - Status test', () => {
     })
 
     expect(res.statusCode).toStrictEqual(200)
-    expect(res.body).toContain('degraded')
+    const body = JSON.parse(res.body)
+    expect(body.status).toStrictEqual('degraded')
   })
 
   it('/blockchain/status - should get outage', async () => {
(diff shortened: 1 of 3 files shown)


apps/status-api/src/controllers/BlockchainController.ts
```diff
@@ -0,0 +1,37 @@
...
+
+    const nowEpoch = Date.now()
+    const latestBlockTime = blocks[0].time * 1000
+    const timeDiff = nowEpoch - latestBlockTime
+
+    const statusMap = new Map<string, number>([
+      ['outage', (45 * 60 * 1000)],
+      ['degraded', (30 * 60 * 1000)],
+      ['operational', 0]
+    ])
```
cwkang1998, line 22: I think something like this

```typescript
type BlockchainStatus = 'outage' | 'degraded' | 'operational'
const StatusMap = new Map<BlockchainStatus, number>([
      ['outage', (45 * 60 * 1000)],
      ['degraded', (30 * 60 * 1000)],
      ['operational', 0]
    ])
```
would help constraint this even further.

Personally, would prefer using normal records for `StatusMap` instead of a map object, as we are not really putting anything into map and does not as such need its utilities. Its more explicit as well.
```typescript
const StatusMap: Record<BlockchainStatus, number> = {
    'outage': 45 * 60 * 1000,
    'degraded': 30 * 60 * 1000,
    'operational': 0
}
```

Pulling it outside of the controller class would also probably be ideal since its suppose to be a constant.

apps/status-api/__test__/controllers/BlockchainController.test.ts
```diff
@@ -0,0 +1,82 @@
...
+      .spyOn(apiTesting.app.get(WhaleApiClient).blocks, 'list')
+      .mockReturnValueOnce(getBlockResponseWithPresetTime(25))
+
+    const res = await apiTesting.app.inject({
+      method: 'GET',
+      url: '/blockchain/status'
+    })
+
+    expect(res.statusCode).toStrictEqual(200)
+    expect(res.body).toContain('operational')
```
cwkang1998, line 25:
```suggestion
    expect(res.body.status).toStrictEqual('operational')
```

eli-lim (changes requested):

eli-lim (approved):

cwkang1998 (approved):

joeldavidw (approved):

PR #1309, opened 2022-04-01
feat(apps/legacy-api): perf improvement with caching

eli-lim:
- Introduces block-level + transaction-level swap caching for getsubgraphswaps query path
- Recent blocks (within chain height - 2) are not cached as they could be invalidated. However, transaction-level swap cache enables us to call AccountHistory rpc at most once for any given txn, hence largely side-stepping the performance bottleneck.

#### 1. Block-level caching of pool swaps needed to get response time in the milliseconds range
- Unfortunately, it's not enough to just cache by txid of the pool swap. Although the bottleneck of accounthistory rpc is side-stepped with this, the times from traversing blocks and their transactions add up.
- Hence, I tried caching the swaps by block hash instead, although we probably shouldn't cache recent blocks since they could be invalidated / more swaps could be added to them. 
- This was great for performance!

#### 2. Fixing pagination and making it work with this caching approach
- Pagination is currently wrongly implemented. Every returned `next` token when used to hit the endpoint again will **skip the block** even if there are pool swaps remaining in the block.
  ```
  blockchain = [ 
    block 2 [{ swap D }, ..., { swap C }, ..., { swap B }]
    block 1 [{ swap A }, ...]
  ]

  /getsubgraphswaps?limit=2 
    - gives: { swaps: [{ swap D }, { swap C }], page: XXX }

  /getsubgraphswaps?limit=2&next=XXX
    - should give:  [{ swap B }, { swap A }]
    - but gives:  [{ swap A }]
  ```
- This is due to the `next.height` value being returned to the client as the current block's height, say 2. So in the next pagination call, the provided `next.height` of 2 is passed to the node rpc call, thereby skipping block 2 (this is expected behaviour).
- We therefore should +1 to `next.height` if the swap **isn't the last swap in the block's swaps**.

- Part of https://github.com/DeFiCh/jellyfish/issues/1024

# Sanity checks
- [x] Responses are the same as https://api.defichain.com/v1/getsubgraphswaps
- [x] Pagination is correct

### DIFF
--- a/apps/legacy-api/__tests__/controllers/PoolPairController.test.ts
+++ b/apps/legacy-api/__tests__/controllers/PoolPairController.test.ts
@@ -289,7 +289,7 @@ describe('getsubgraphswaps', () => {
     expect(response.data.swaps.length).toStrictEqual(0)
   })
 
-  it('/v1/getsubgraphswaps?limit=101 - limited to 30', async () => {
+  it('/v1/getsubgraphswaps?limit=101 - limited to 100', async () => {
     const res = await apiTesting.app.inject({
       method: 'GET',
       url: '/v1/getsubgraphswaps?limit=101'
@@ -298,7 +298,7 @@ describe('getsubgraphswaps', () => {
     expect(response.data.swaps.length).toStrictEqual(100)
   })
 
-  it('should paginate', async () => {
+  it('/v1/getsubgraphswaps?limit=X&next=Y - should paginate', async () => {
     const [swap1And2, swap1]: any = await Promise.all([
       apiTesting.app.inject({
         method: 'GET',
@@ -326,7 +326,7 @@ describe('getsubgraphswaps - relying on caching', () => {
     // Given caching enabled and cache is filled sufficiently
     await waitForCondition(
       async () => apiTesting.app.get(SwapCacheFiller).isReady,
-      60_000 // 60s
+      30_000 // 30s
     )
   })
 
--- a/apps/legacy-api/__tests__/controllers/PoolPairController.test.ts
+++ b/apps/legacy-api/__tests__/controllers/PoolPairController.test.ts
@@ -326,7 +326,7 @@ describe('getsubgraphswaps - relying on caching', () => {
     // Given caching enabled and cache is filled sufficiently
     await waitForCondition(
       async () => apiTesting.app.get(SwapCacheFiller).isReady,
-      30_000 // 30s
+      60_000 // 60s
     )
   })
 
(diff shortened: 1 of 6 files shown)


fuxingloh (commented):
looks simple 🚀

How's the performance?

eli-lim:
> looks simple 🚀
> 
> How's the performance?

Pretty good - about 200ms for 100 swaps (once cache is warmed up)

apps/legacy-api/src/controllers/PoolPairController.ts
```diff
@@ -513,3 +647,13 @@
 
   return undefined
 }
+
+export function encodeBase64 (next: NextToken): string {
+  return Buffer.from(JSON.stringify(next), 'utf8').toString('base64url')
```
weiyuan95, line 652: `toString(base64url) was introduced in v14.18` ref: https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V14.md#14.18.0 so we'll need to set a dependency on this version in `jellyfish/package.json` (don't ask me how I realised this)

weiyuan95 (changes requested):

apps/legacy-api/src/controllers/PoolPairController.ts
```diff
@@ -287,6 +373,54 @@
   }
 }
 
+/**
+ * Hits PoolPairController with requests to enable pre-fetching behaviour,
+ * improving performance on queries
+ */
+@Injectable()
+export class SwapCacheFiller {
```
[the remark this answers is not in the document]
eli-lim: Same area of concern, hence keeping it where it is

apps/legacy-api/__tests__/controllers/PoolPairController.test.ts
```diff
@@ -262,12 +295,192 @@ describe('getsubgraphswaps', () => {
...
+
+    expect(swap1.data.swaps[0]).toStrictEqual(swap1And2.data.swaps[0])
+    expect(swap2.data.swaps[0]).toStrictEqual(swap1And2.data.swaps[1])
+  })
+})
+
+describe('getsubgraphswaps - relying on caching', () => {
+  beforeAll(async () => {
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    async function relyOnProduction (): Promise<void> {
```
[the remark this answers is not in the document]
eli-lim: No - but it's useful for local sanity testing to actually see the production data being served.

apps/legacy-api/src/controllers/PoolPairController.ts
eli-lim: Lemme guess - you were running with < v14.18 locally? LOL
So you're suggesting we bump https://github.com/DeFiCh/jellyfish/blob/ee2016fd86e172a53fd5ef5a2393173ce8fd9807/package.json#L17 to `>= 14.18` ?

If you have `nvm` installed, IDEA ide should automagically pick the node runtime when running anything in the project

weiyuan95 (approved):
:shipit:

apps/legacy-api/src/controllers/PoolPairController.ts
weiyuan95: We don't have a .npmrc which enforces the version of node, so marking this as resolved

fuxingloh (approved):

PR #1314, opened 2022-04-04
chore(apps): add sanity testing containers

kodemon:

We want to be able to run a separate suite of sanity tests against our applications being released as docker containers in our ecosystem. This is a part of the shift left automation effort of providing more tooling to quality test our releases before they hit production.

We also want to be able to write sanity tests in the format that we are already familiar with through our standard unit tests instead of the previous format in whale which were running through postman. This unifies our test paradigm into a single workflow which lessens the cognitive impact of the testing part of our workflows.

It's set up to support running tests in parallel for vertical scaling capabilities and I have written some minor reasoning for key methods in this PR through code comments. Please read those for additional details on this solution.

Partially Resolves
 - https://github.com/DeFiCh/jellyfish/issues/942

#### Additional comments?:

This is part of the whale consolidation efforts noting the final stretch tasks that we need to finish before being 100% migrated. Remaining efforts trailing of this PR will after some internal communication be an approximate of:

 - Port sanity tests from whale postman to new sanity test files in jellyfish.
 - Move over any changes to whale that has not been ported since last update.
 - Add whale app, whale-api-client, and whale-api-wallet to the release workflow of jellyfish and decommission whale.


packages/testcontainers/src/containers/SanityContainer/index.ts
```diff
@@ -0,0 +1,120 @@
...
+ * and running instances on random ports to support parallel testing allowing for
+ * vertical scaling in our test environments.
+ */
+
+export abstract class SanityContainer extends DockerContainer {
+  public readonly name = this.generateName()
+
+  constructor (
+    public readonly app: string,
+    public readonly port = Math.floor(Math.random() * (5000 - 3000 + 1) + 3000),
```
ivan-zynesis, line 32: Looks like a mistake.  `floor ( 0.xx + 2001 + 3000 )`, the random will be always ignored

ivan-zynesis (dismissed):

ivan-zynesis:
But looking at the structure, doesn't seems easy to further extend it for playground use. (for wallet side e2e use)

packages/testcontainers/src/containers/SanityContainer/index.ts
kodemon: The randomized results does work as intended, so if the random was ignored I think this would always return a static result no?

ivan-zynesis: Oops, misread it. Anyway, just put `5001` will do 😄

kodemon:
> But looking at the structure, doesn't seems easy to further extend it for playground use. (for wallet side e2e use)

Could you elaborate on the points you think makes it hard to extend, ?

I think from a container perspective the only thing the solution does here is build the image you need and enables you to run an instance from it. Providing a base which enables tooling for the things that are repeatable between each individual requirement, then creating the specific requirements in their own classes such as this initial PR introduces through `WhaleSanityContainer`.

That said if we identify thing in the base that aren't shared across all our applications then we would want to move them into further abstractions. The idea is definitely that this will require some adjustments when new requirements come in that doesn't meet expectations of our other apps.

Happy to improve this with more contributions as I am sure there may be requirements in our other applications not being considered in this initial solution.

ivan-zynesis:
> > But looking at the structure, doesn't seems easy to further extend it for playground use. (for wallet side e2e use)
> 
> Could you elaborate on the points you think makes it hard to extend?
> 
> I think from a container perspective the only thing the solution does here is build the image you need and enables you to run an instance from it. Providing a base which enables tooling for the things that are repeatable between each individual requirement, then creating the specific requirements in their own classes such as this initial PR introduces through `WhaleSanityContainer`.
> 
> That said if we identify things in the base that aren't shared across all our applications then we would want to move them into further abstractions. The idea is definitely that this will require some adjustments when new requirements come in that doesn't meet expectations of our other apps.
> 
> Happy to improve this with more contributions as I am sure there may be requirements in our other applications not being considered in this initial solution.

Perhaps it's alright for playground, since the block minting bot is done via RPC.

kodemon:
> Perhaps it's alright for playground, since the block minting bot is done via RPC.

Also if there are anything we need to adjust for when we start adding sanity testing for our other apps, then it should be pretty straight forward to do so. Initially as mentioned this PR currently is just setting up a strategy to build our app containers and being able to run and test against their endpoints for now, and there is definitely room for modifications as we need to expand its capabilities.

Thanks for the feedback 👍🏻

packages/testcontainers/src/containers/SanityContainer/index.ts
```diff
@@ -0,0 +1,127 @@
...
+ * unit tests to create uniformity in our approach to testing. Reducing the
+ * cognitive complexity of having to run multiple automated solutions to ensure
+ * release quality.
+ *
+ * This solution works by building a new docker image using our root Dockerfile
+ * and running instances on random ports to support parallel testing allowing for
+ * vertical scaling in our test environments.
+ */
+
+export abstract class SanityContainer extends DockerContainer {
```
eli-lim, line 27:
```suggestion
 */
export abstract class SanityContainer extends DockerContainer {
```

eli-lim (approved):

packages/testcontainers/src/containers/SanityContainer/index.ts
kodemon: This was originally deliberate but seeing its content description it makes sense to connect it directly to the class 👍🏻

kodemon:
I have some minor changes as I ran into a issue with having the images being built at jest runtime. Will update description once I push the updated version.

apps/tsconfig.json
```diff
@@ -2,7 +2,8 @@
   "extends": "../tsconfig.json",
   "include": [
     "./*/src",
-    "./libs/*/src"
+    "./libs/*/src",
+    "./*/__sanity__"
```
eli-lim, line 6:
```suggestion
```
Is there a reason why we're including `__sanity__` tests in the production build?

eli-lim (changes requested):

eli-lim (dismissed):
More sanity tests to follow after this is merged

apps/whale/__sanity__/WhaleSanityContainer.sanity.ts
```diff
@@ -0,0 +1,59 @@
+import { WhaleSanityContainer } from '@defichain/testcontainers'
+import waitForExpect from 'wait-for-expect'
+
+const whale = new WhaleSanityContainer()
+
+// TODO(kodemon/eli-lim):
+//   would it make sense to use WhaleApiClient for sanity tests,
+//   instead of raw http requests?
```
eli-lim, line 8: IMO, we probably should test both raw http and via WhaleApiClient, as it'll allow us to sanity test WhaleApiClient code too

jingyi2811 (dismissed):

package.json
```diff
@@ -19,6 +19,7 @@
     "prepare": "husky install",
     "lint": "eslint . --fix",
     "test": "jest --maxWorkers=100%",
+    "sanity": "jest --maxWorkers=100% --config=jest.sanity.js",
```
jingyi2811, line 22: Do you want to add --testTimeout=<number>?
I run your test and I get this
Screenshot 2022-05-18 at 1 02 13 AM

jingyi2811:
I feel like the randomized number used in this PR should be replaced by an incremental number 

3000, 3001, 3002 ... 5000
1, 2, 3 ... 100000

packages/testcontainers/src/containers/AppContainer/index.ts
```diff
@@ -0,0 +1,95 @@
...
+   * reduce the amount of lingering docker containers as a result of running tests.
+   */
+  public async stop (): Promise<void> {
+    await this.container?.stop()
+    await this.container?.remove({ v: true })
+    await this.blockchain.stop()
+  }
+
+  public generateName (): string {
+    const rand = Math.floor(Math.random() * 10000000)
```
weiyuan95, line 63: I think we should use `uuid` to generated this random string instead of using `Math.random` to avoid clashing container names? As per https://github.com/JellyfishSDK/jellyfish/pull/1332

weiyuan95 (commented):
One quick question about the random container name, not sure if it's a blocker to this PR. If it isn't, i'll approve

packages/testcontainers/src/containers/AppContainer/index.ts
jingyi2811: Not a blocker. I have approved

weiyuan95 (dismissed):

package.json
kodemon: Build times can definitely be excessive in some cases, what do we think is a sensible timeout value?

eli-lim (approved):
had to update:
- testTimeout to 5 mins (docker build takes awhile)
- change container name generation to use `uuidv4()`

weiyuan95 (approved):

jingyi2811 (approved):

cwkang1998:
Not sure why runnning locally doesn't work for me on my linux machine.
Works for mac right now though.

```bash
 FAIL  apps/whale/__sanity__/WhaleSanityContainer.sanity.ts (51.611 s)
  /_actuator
    /_actuator/probes/liveness
      ✕ Status in JSON body is ok (1 ms)
    /_actuator/probes/readiness
      ✕ Status code is 503 (1 ms)

  ● /_actuator › /_actuator/probes/liveness › Status in JSON body is ok

    waitForCondition is not ready within given timeout of 30000ms.

      16 |         resolve()
      17 |       } else if (expiredAt < Date.now()) {
    > 18 |         reject(new Error(`${message} is not ready within given timeout of ${timeout}ms.`))
         |                ^
      19 |       } else {
      20 |         setTimeout(() => void checkCondition(), interval)
      21 |       }

      at checkCondition (packages/testcontainers/src/utils.ts:18:16)
          at runMicrotasks (<anonymous>)

  ● /_actuator › /_actuator/probes/readiness › Status code is 503

    waitForCondition is not ready within given timeout of 30000ms.

      16 |         resolve()
      17 |       } else if (expiredAt < Date.now()) {
    > 18 |         reject(new Error(`${message} is not ready within given timeout of ${timeout}ms.`))
         |                ^
      19 |       } else {
      20 |         setTimeout(() => void checkCondition(), interval)
      21 |       }

      at checkCondition (packages/testcontainers/src/utils.ts:18:16)
          at runMicrotasks (<anonymous>)

Test Suites: 1 failed, 1 total
Tests:       2 failed, 2 total
Snapshots:   0 total
Time:        51.701 s
```

cwkang1998 (commented):
Other than those concerns, I think should be okay. We can move debugging why it doesn't work on linux and setting it up for ci to another PR if there's a need for speed.

Issue #1274, opened 2022-03-28
`apps/status-api`: Implement Overall Status


joeldavidw:
#### What would you like to be added:
An endpoint (`/overall`) to provide the aggregated status of all services required by LW & Scan (Blockchain & Ocean)

Priority of status
1. Blockchain Status
2. Ocean

Status:
- operational: All running fine
- degraded: 
   - Blockchain: degraded
   - Ocean: < 3 regions down
- outage:
   - Blockchain: outage
   - Ocean: All region or MainNet v0 down 

/area apps

PR #1315, opened 2022-04-04
feat(status-api): Status API for Oracles

nichellekoh:
This includes the status api to return the status of each oracles
operational: <= 15 mins
outage: > 15 mins

Part of  https://github.com/DeFiCh/jellyfish/issues/1274


apps/status-api/src/controllers/OracleStatusController.ts
```diff
@@ -0,0 +1,42 @@
...
+    private readonly client: WhaleApiClient,
+    protected readonly cache: SemaphoreCache
+  ) {
+  }
+
+  @Get('oracles')
+  async getOracleStatus (
+    @Query('address') oracleAddress: string
+  ): Promise<{ [key: string]: string }> {
+    const oracle: Oracle = await this.cachedGet('ORACLE', async () => {
```
joeldavidw, line 18:
```suggestion
    const oracle: Oracle = await this.cachedGet(`oracle-${oracleAddress}`, async () => {
```
Cache id has to be unique to the oracle.

apps/status-api/src/controllers/OracleStatusController.ts
```diff
@@ -0,0 +1,43 @@
...
+      const oracle = await this.client.oracles.getOracleByAddress(oracleAddress)
+      return (await this.client.oracles.getPriceFeed(oracle.id, oracle.priceFeeds[0].token, oracle.priceFeeds[0].currency, 1))[0]
+    }, 5000)
+
+    const nowEpoch = Date.now()
+    const latestPublishedTime = oraclePriceFeed.block.medianTime * 1000
+    const timeDiff = nowEpoch - latestPublishedTime
+
+    return {
+      status: timeDiff <= (45 * 60 * 1000) ? 'operational' : 'outage'
```
ivan-zynesis, line 28: How about ticker "holiday"?

nichellekoh: Hmm how would we determine if the ticker is on 'holiday'?

ivan-zynesis: Probably is too much to be able to precisely determine holiday. Perhaps can check for weekend (might have to make assumption for ticker time zone).

Any thoughts for not showing `outage` when it is not meant to? @fuxingloh @joeldavidw

nichellekoh: The providers will be repeatedly return with the last known price when the market closes over the weekends to keepalive. So in the event that there are no published time, it will be an outage.

ivan-zynesis: 😅 then can ignore the question in the first place, the source is up all time regardless "holiday"

ivan-zynesis (changes requested):
Minor change suggestion for more explicit api response

apps/status-api/src/controllers/OracleStatusController.ts
```diff
@@ -0,0 +1,43 @@
...
+  constructor (
+    private readonly client: WhaleApiClient,
+    protected readonly cache: SemaphoreCache
+  ) {
+  }
+
+  @Get()
+  async getOracleStatus (
+    @Query('address') oracleAddress: string
+  ): Promise<{ [key: string]: string }> {
```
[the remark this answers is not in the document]
nichellekoh: Introduced a type OracleStatus to ensure this

joeldavidw (approved):

joeldavidw (approved):

kodemon (approved):

apps/status-api/src/modules/ControllerModule.ts
```diff
@@ -3,6 +3,8 @@ import { ActuatorController } from '@defichain-apps/libs/actuator'
 import { BlockchainStatusController } from '../controllers/BlockchainStatusController'
 import { WhaleApiClient } from '@defichain/whale-api-client'
 import { ConfigService } from '@nestjs/config'
+import { SemaphoreCache } from '../../../whale/src/module.api/cache/semaphore.cache'
```
cwkang1998, line 6: This should probably be extracted out of `whale` as a `lib` module.

cwkang1998 (approved):

apps/status-api/src/modules/ControllerModule.ts
nichellekoh: https://github.com/DeFiCh/jellyfish/issues/1358 created

ivan-zynesis (approved):

apps/status-api/README.md
```diff
@@ -8,12 +8,29 @@ DeFiChain Status API, providing the statuses of different DeFiChain services
...
+
+| Status             | Threshold Time    |
+|--------------------|-------------------|
+| `operational`      | `< 30 minutes`    |
+| `degraded`         | `30 - 45 minutes` |
+| `outage`           | `> 45 minutes`    |
+
+### `/oracles?address=<address>`
+
+To provide the status of each oracle given the address based on the last published time for any given token
```
fuxingloh, line 24: Based on the written statement below, "address" isn't an optional parameter; shouldn't address in `/oracles/:address` be part of the path instead?

Unless the scenario here is: given address is optional, we check the health of all oracles?

joeldavidw (approved):
🙏

fuxingloh (approved):

cwkang1998 (approved):

PR #1316, opened 2022-04-04
refactor(jellyfish-api-core): use precise typing for `AccountHistory`

### LINKED ISSUE #1303: Precise typing of `AccountHistory`
Split and differentiate `AccountHistory` into with and without txid child type
eg LM reward has no txid but found in `listaccounthistory`

#### Why is this needed:
Manage developer expectation. Reduce unnecessary rpc response manipulation. 

Production sample (LM reward) without txid
```
{
    "owner": "dFFPENo7FPMJpDV6fUcfo4QfkZrfrV1Uf8",
    "blockHeight": 1756148,
    "blockHash": "464fa7d2abb442e1597c27a35613378888b602b5d487c75105c498bc494c70e8",
    "blockTime": 1648749175,
    "type": "Commission",
    "poolID": "6",
    "amounts": [
      "0.00000007@DFI"
    ]
  }
```

cwkang1998:

This PR aims to provide more precise typing for `AccountHistory`, by utilizing function overriding and a more specific `AccountHistoryWithTxn` type for those with `no_rewards: true` options.

Fixes #1303

### DIFF
--- a/packages/jellyfish-api-core/src/category/account.ts
+++ b/packages/jellyfish-api-core/src/category/account.ts
@@ -278,6 +278,23 @@ export class Account {
     return await this.client.call('accounttoutxos', [from, payload, options.utxos], 'number')
   }
 
+  async listAccountHistory (): Promise<AccountHistory[]>;
+
+  async listAccountHistory (
+    owner: OwnerType | string,
+    options: Omit<AccountHistoryOptions, 'no_rewards'> & {no_rewards: false}
+  ): Promise<AccountHistoryWithTxn[]>;
+
+  async listAccountHistory (
+    owner: OwnerType | string,
+    options: Omit<AccountHistoryOptions, 'no_rewards'> & {no_rewards: true}
+  ): Promise<AccountHistoryWithTxn[]>;
+
+  async listAccountHistory (
+    owner: OwnerType | string,
+    options?: AccountHistoryOptions
+  ): Promise<AccountHistory[]>;
+
   /**
    * Returns information about account history
    *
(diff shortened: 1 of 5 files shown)


packages/jellyfish-api-core/src/category/account.ts
```diff
@@ -278,6 +278,23 @@ export class Account {
...
+  async listAccountHistory (): Promise<AccountHistory[]>;
+
+  async listAccountHistory (
+    owner: OwnerType | string,
+    options: Omit<AccountHistoryOptions, 'no_rewards'> & {no_rewards: false}
+  ): Promise<AccountHistory[]>;
+
+  async listAccountHistory (
+    owner: OwnerType | string,
+    options: Omit<AccountHistoryOptions, 'no_rewards'> & {no_rewards: true}
```
ivan-zynesis, line 290: @surangap can confirm everything else should come with txid when `no_rewards` true?
The flag also handle dToken pair rewards and permission right?

surangap: `no_rewards` controls the pool rewards. so yeah `no_rewards = true` should give all with txids.
when calculating pool rewards, it iterates through all pools. so dtoken pools are also included.

surangap: maybe can add a testcase for👆 if not there already

ivan-zynesis: 👍 I don't think there is any test with rewards distributing in action. See https://github.com/DeFiCh/jellyfish/blob/main/apps/playground-api/src/setups/setup.dex.ts#L212-L215 for the setup required @cwkang1998

PR #1324, opened 2022-04-05
feat(jellyfish-api-core): implement refundHtlcAll

cwkang1998:
Implements spv `spv_refundhtlcall` rpc call.

Fixes part of #202 

#### Additional comments?:
Need some confirmation with how the `ain` implementation of this rpc is. Right now `ain`'s implementation returns an array of strings, containing only one string, shown [here](https://github.com/DeFiCh/ain/blob/86f554f2454edc8c16684cac6f7332b437a991e0/src/spv/spv_wrapper.cpp#L1354-L1379).

The [rpc documentation from `ain`](https://github.com/DeFiCh/ain/blob/86f554f2454edc8c16684cac6f7332b437a991e0/src/spv/spv_rpc.cpp#L1126) however shows that it returns a string txid, which is not the case.

This current implementation for jellyfish deals with this by simply taking the first element of the returned array.

Edit: This rpc now returns a string array, faithful to its `ain` implementation.

### DIFF
--- a/packages/jellyfish-api-core/__tests__/category/spv/refundHtlcAll.test.ts
+++ b/packages/jellyfish-api-core/__tests__/category/spv/refundHtlcAll.test.ts
@@ -1,0 +1,210 @@
+import { MasterNodeRegTestContainer } from '@defichain/testcontainers'
+import { BigNumber, RpcApiError } from '@defichain/jellyfish-api-core'
+import { Testing } from '@defichain/jellyfish-testing'
+
+describe('Spv', () => {
+  let testing: Testing
+  let container: MasterNodeRegTestContainer
+
+  beforeEach(async () => {
+    testing = Testing.create(new MasterNodeRegTestContainer())
+    container = testing.container
+    await container.start()
+    await container.spv.fundAddress(await container.call('spv_getnewaddress')) // Funds 1 BTC
+  })
+
+  afterEach(async () => {
+    await container.stop()
+  })
+
+  it('should refundHtlcAll', async () => {
+    const pubKeyA = await container.call('spv_getaddresspubkey', [await container.call('spv_getnewaddress')])
+    const pubKeyB = await container.call('spv_getaddresspubkey', [await container.call('spv_getnewaddress')])
+    const htlc1 = await container.call('spv_createhtlc', [pubKeyA, pubKeyB, '10'])
+    const htlc2 = await container.call('spv_createhtlc', [pubKeyA, pubKeyB, '10'])
+
+    const fund1 = await container.call('spv_sendtoaddress', [htlc1.address, 0.1]) // Funds HTLC address
+    const fund2 = await container.call('spv_sendtoaddress', [htlc2.address, 0.2])
+    await container.spv.increaseSpvHeight(10)
+
+    const destinationAddress = await container.call('spv_getnewaddress')
+    const result = await testing.rpc.spv.refundHtlcAll(destinationAddress) // This refund should only happen after timeout threshold set in createHtlc. See https://en.bitcoin.it/wiki/BIP_0199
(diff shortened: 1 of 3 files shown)


canonbrother (commented):
LGTM!!
(Regarding the **return arr** from ain.. if that does really matter.. we'd open issue on ain)

ivan-zynesis:
> LGTM!! (Regarding the **return arr** from ain.. if that does really matter.. we'd open issue on ain)

Hmm prefer to return array instead for consistencies. I am thinking if ain side is a bug, what if the wallet has multiple htlc, then it should return txid of each refunded.

canonbrother (dismissed):

cwkang1998:
> > LGTM!! (Regarding the **return arr** from ain.. if that does really matter.. we'd open issue on ain)
> 
> Hmm prefer to return array instead for consistencies. I am thinking if ain side is a bug, what if the wallet has multiple htlc, then it should return txid of each refunded.

If so then I think I'll make it return a string array, but I'll just note it somewhere that there is this inconsistency then.

ivan-zynesis (approved):

canonbrother (approved):

packages/jellyfish-api-core/__tests__/category/spv/refundHtlcAll.test.ts
```diff
@@ -0,0 +1,219 @@
...
+    const htlc = await container.call('spv_createhtlc', [pubKeyA, pubKeyB, '10'])
+
+    await testing.container.call('spv_sendtoaddress', [htlc.address, 0.00000546]) // Funds HTLC address with dust
+
+    const promise = testing.rpc.spv.refundHtlcAll(await container.call('spv_getnewaddress'))
+    await expect(promise).rejects.toThrow(RpcApiError)
+    await expect(promise).rejects.toThrow('No unspent HTLC outputs found')
+  })
+
+  it('should not refundHtlcAll when not htlc is created for wallet', async () => {
```
surangap, line 214:
```suggestion
  it('should not refundHtlcAll when no htlc is created for wallet', async () => {
```

packages/jellyfish-api-core/__tests__/category/spv/refundHtlcAll.test.ts
```diff
@@ -0,0 +1,219 @@
...
+    /**
+     * Assert that the destination address received the refund
+     */
+    const listReceivingAddresses: any[] = await container.call('spv_listreceivedbyaddress')
+    const receivingAddress = listReceivingAddresses.find(({ address }: { address: string }) => address === destinationAddress)
+    expect(receivingAddress.address).toStrictEqual(destinationAddress)
+    expect(receivingAddress.txids.some((txid: string) => txid === results[0])).toStrictEqual(true)
+  })
+
+  it('should refundHtlcAll for expired only', async () => {
```
surangap, line 79: do you expect `htlc2` to be no refunded in this test? wondering where is the `for expired only`
Refer https://github.com/DeFiCh/jellyfish/pull/529#discussion_r688293825

cwkang1998: Yes I do. In real life scenario, `refundHtlc` would probably not be called before timeout. However in this case, `refundHtlcAll` would be called in a real life scenario, as it initiate refunds only for those that already timeout, and not htlc that has not timed out yet.

cwkang1998: Should I change the title to indicate for timeout?

ivan-zynesis (approved):

surangap (approved):

canonbrother (approved):

Issue #1268, opened 2022-03-28
`packages/jellyfish-api-core`: Add DFI2203 Futures work


surangap:
#### What would you like to be added:
map endpoints related to https://github.com/DeFiCh/ain/pull/1155

#### Why is this needed:
https://github.com/DeFiCh/dfips/issues/127

jingyi2811:
futureswap: @surangap 
withdrawfutureswap: 
listpendingfutureswaps: @jingyi2811
getpendingfutureswaps: 
getfutureswapblock:

jingyi2811:
After reading the python code, I believe futureswap and withdrawfutureswap RPCs are harder and you may spend more time on them. Therefore, I pick the rest of the RPCs

PR #1329, opened 2022-04-06
feat(jellyfish-transaction, jellyfish-transaction-builder): Add futureswap dftxs

surangap:
Add `futureswap` and `withdrawfutureswap` dftxs

Fixes part of #1268


surangap:
> > This PR is good to be merged to main.
> > Before merging, it is good to add this test item `should create multiple dtoken to dusd futureswaps within the same block` (See futureswap.test.ts)
> > Maybe @surangap you have good reason to remove this in your dftx test file, kindly advise
> 
> yeah two reasons
> 
> 1. `sendTransaction()` has a `generate(1)` inside it. But we can modify it.
> 2. Also `sendrawtransaction` RPC only supports sending only one tx at a time. so no multiple tx send at the same time.
> 3. Currently, `fundAddress()` also generates blocks inside. This is for the fees.
> 
> We need to update the above👆 functions to have it done within one block and check.
@jingyi2811 
btw i have the half-baked test case for this. Currently facing some issues with funding for txs fees. Coz with our current mock provider, one utxo can be used only once for some reason. I remember try digging it in the past. I guess have to continue doing it.
anw let's merge this since downstream require a release. 
will add this test later.

cwkang1998 (approved):

chanakasameera (approved):

PR #1383, opened 2022-04-25
fix(status-api): Update to check status of first active tickers of the oracle

nichellekoh:
Previously, the logic for checking status of the oracle takes the first ticker returned from whale. However, some of the tickers are dormant, which means no data or outdated data will be returned, resulting in inaccurate status.

Fixes #

### DIFF
--- a/apps/status-api/src/controllers/OracleStatusController.ts
+++ b/apps/status-api/src/controllers/OracleStatusController.ts
@@ -1,6 +1,6 @@
 import { Controller, Get, Param } from '@nestjs/common'
 import { WhaleApiClient } from '@defichain/whale-api-client'
-import { OraclePriceFeed } from '@defichain/whale-api-client/dist/api/Oracles'
+import { Oracle } from '@defichain/whale-api-client/dist/api/Oracles'
 import { SemaphoreCache } from '../../../whale/src/module.api/cache/semaphore.cache'
 
 type OracleStatus = 'outage' | 'operational'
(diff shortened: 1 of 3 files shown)


cwkang1998 (approved):

weiyuan95 (dismissed):

canonbrother (approved):

apps/status-api/src/controllers/OracleStatusController.ts
```diff
@@ -21,18 +21,31 @@ export class OracleStatusController {
...
+    const id = oracle.id
 
-    const nowEpoch = Date.now()
-    const latestPublishedTime = oraclePriceFeed.block.medianTime * 1000
-    const timeDiff = nowEpoch - latestPublishedTime
+    for (const priceFeed of oracle.priceFeeds) {
+      const token = priceFeed.token
+      const currency = priceFeed.currency
+      const oraclePriceFeed = await this.client.oracles.getPriceFeed(id, token, currency, 1)
+      if (oraclePriceFeed.length !== 0) { // check if ticker is not returning any data, else move on to the next ticker
```
fuxingloh, line 37:
```suggestion
      if (oraclePriceFeed.length -== 0) { // check if ticker is not returning any data, else move on to the next ticker
      continue
```

inverse logic for reliability?

fuxingloh (changes requested):

apps/status-api/src/controllers/OracleStatusController.ts
jingyi2811: Whenever you want to use !== 0, consider using >=0 instead for those values that couldn't be a negative value to increase code readability

fuxingloh (approved):

weiyuan95 (approved):
:shipit:

jingyi2811 (approved):

cwkang1998 (approved):

canonbrother (approved):

PR #1391, opened 2022-04-26
fix(status-chore): Add prefer-template rule to eslint

### LINKED ISSUE #1330: check and update lint for below
I am thinking about why does eslint allow this. Hmm. https://eslint.org/docs/rules/prefer-template

_Originally posted by @jingyi2811 in https://github.com/DeFiCh/jellyfish/pull/1273#discussion_r843581276_

jingyi2811:

Add prefer-template rule to eslint

Fixes #1330

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -35,7 +35,8 @@
       }
     ],
     "import/no-default-export": "error",
-    "no-else-return": "error"
+    "no-else-return": "error",
+    "prefer-template": "error"
   },
   "overrides": [
     {


jingyi2811:
This is what I did

1) Add `"prefer-template": "error"` to rules of .eslintrc
2) Commit and push
3) Run  eslint . --fix (This is to ensure that all the old code follows this new rule)
4) Commit and push

jingyi2811:
In the future, whenever you commit any code,
npx --no-install lint-staged (See husky file) will execute
lint-staged (See package.json) 's `eslint --fix`

The fix only apply to the code you are going to commit (We call these staged files)

canonbrother (approved):

cwkang1998 (approved):
🚀

cwkang1998:
> In the future, whenever you commit any code, npx --no-install lint-staged (See husky file) will execute lint-staged (See package.json) 's `eslint --fix`
> 
> The fix only applies to the code you are going to commit (We call these stage files)

If I am not wrong we have husky which automatically calls that for us, is there a difference between that and executing manually?

jingyi2811:
> > In the future, whenever you commit any code, npx --no-install lint-staged (See husky file) will execute lint-staged (See package.json) 's `eslint --fix`
> > The fix only applies to the code you are going to commit (We call these stage files)
> 
> If I am not wrong we have husky which automatically calls that for us, is there a difference between that and executing manually?

Good question

It is impossible to fix a file using `eslint --fix`
Seems like the command can only be executed / understandable by husky

I am unsure of the command to fix a file if doing it manually
What I know is `eslint . --fix ` can fix every file

The difference is 

husky fixes staged files
The manual way fixes everything (I think there is a command to fix a single file but the command is definitely not `eslint --fix`)

Husky seems connected well with GIT
The manual way is independent of GIT as far as I am concerned

fuxingloh (approved):

PR #1392, opened 2022-04-26
feat(workflows): Implemented workflow to push images to ECR from GitHub Actions

nichellekoh:
Replicate Docker images from ghcr.io over to AWS ECR on each region to prevent a single point of failure.

When ghcr.io goes down, our deployment pipeline is killed. We cannot use a single region provider either, it will still be a single point of failure. Replicating the Docker image to each AWS region with ocean using AWS ECR will prevent that single point of failure since ECR is region-independent.

This will create a release version image on AWS ECR Public repository: https://gallery.ecr.aws/x0i4b0k2/status-api

Fixes #


.github/workflows/release-ecr.yml
```diff
@@ -0,0 +1,53 @@
...
+
+env:
+  APPS: 'status-api'
+
+jobs:
+  build:
+    if: github.actor != 'dependabot[bot]'
+    name: Publish
+    runs-on: ubuntu-latest
+    environment: Release Apps
```
joeldavidw, line 19: environment: `ECR Publish Releasing`

.github/workflows/release-ecr.yml
```diff
@@ -0,0 +1,53 @@
...
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@94ab11c41e45d028884a99163086648e898eed25
+
+      - name: Login to Public ECR
+        uses: docker/login-action@dd4fa0671be5250ee6f50aedf4cb05514abda2c7
+        with:
+          registry: public.ecr.aws
+          username: ${{ secrets.OPEN_IMAGE_REGISTRY_AWS_ACCESS_KEY_ID }}
+          password: ${{ secrets.OPEN_IMAGE_REGISTRY_AWS_SECRET_ACCESS_KEY }}
```
joeldavidw, line 37:
```suggestion
          username: ${{ secrets.AWS_ACCESS_KEY_ID }}
          password: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
```

joeldavidw (changes requested):

joeldavidw (dismissed):

cwkang1998:
> Docker build preview for jellyfish/apps is ready!
> 
> Built with commit [182e1e3](https://github.com/JellyfishSDK/jellyfish/commit/182e1e37568558666ae00237a2bb2ac7af20b6f8)
> 
>     * ghcr.io/jellyfishsdk/legacy-api:pr-1392
> 
>     * ghcr.io/jellyfishsdk/ocean-api:pr-1392
> 
>     * ghcr.io/jellyfishsdk/playground-api:pr-1392
> 
>     * ghcr.io/jellyfishsdk/status-api:pr-1392

Possible to include the url into github-action bot as well for build preview?

nichellekoh:
> > Docker build preview for jellyfish/apps is ready!
> > Built with commit [182e1e3](https://github.com/JellyfishSDK/jellyfish/commit/182e1e37568558666ae00237a2bb2ac7af20b6f8)
> > ```
> > * ghcr.io/jellyfishsdk/legacy-api:pr-1392
> > 
> > * ghcr.io/jellyfishsdk/ocean-api:pr-1392
> > 
> > * ghcr.io/jellyfishsdk/playground-api:pr-1392
> > 
> > * ghcr.io/jellyfishsdk/status-api:pr-1392
> > ```
> 
> Possible to include the url into github-action bot as well for build preview?

Hmm, this only applies to release build, so there will not be any preview for PR.

cwkang1998 (dismissed):
:shipit:

.github/scripts/release-ecr-tags.js
```diff
@@ -0,0 +1,21 @@
...
+ */
+
+module.exports = ({ context }) => {
+  if (context.eventName === 'release') {
+    return getReleaseTag(context)
+  }
+  throw new Error('Release Violation: Could not determine the required release tags.')
+}
+
+function getReleaseTag(domain, app, context) {
```
jingyi2811, line 15: Actually, your code won't work.

`return getReleaseTag(context)` at line 10

pass in the first parameter which is used for `domain` but `domain` is never used in this function.

Hence your context will always be undefined.

You should get this runtime error
=> `TypeError: Cannot read property 'payload' of undefined`

.github/scripts/release-ecr-tags.js
```diff
@@ -0,0 +1,21 @@
...
+  const semver = context.payload.release.tag_name.replace('v', '')
+  if (semver.match(/^[0-9]+\.[0-9]+\.[0-9]+$/) === false) {
```
jingyi2811, line 17: 1) match will always return something if able to match (For eg: `[ '1.1.1', index: 0, input: '1.1.1', groups: undefined ]` for `1.1.1`)
2) match will always return null if unable to match (For eg: null for `22`)

As it is impossible to get false for ALL THE SCENARIO, this line `Release Violation: Could not determine the required release tags` will never be executed

See match docs below
Screenshot 2022-05-01 at 5 39 20 PM

You maybe can write something like this
```suggestion
  if (semver.match(/^[0-9]+\.[0-9]+\.[0-9]+$/) === null) {
```

jingyi2811: Maybe can remove this line `context.payload.release.tag_name.replace('v', '')`
If not your code may allow other strings like `vv1.1.1vv` or `v1.v1.v1`

If you agree with me kindly change the same code you referred to in Jellyfish

nichellekoh: Thanks @jingyi2811 for bringing this up. The existing jellyfish code replaces only the first 'v' which means strings like `vv1.1.1vv` or `v1.v1.v1` will not pass the regex condition, which was why it will work for such scenarios.

But I agree that your approach will be more fool proof, where currently strings without the `v` such as `1.1.1` will pass the regex. Let me update it on both existing and new file.

.github/scripts/release-ecr-tags.js
nichellekoh: Great catch @jingyi2811, thanks!

.github/workflows/release-ecr.yml
```diff
@@ -0,0 +1,53 @@
...
+
+jobs:
+  build:
+    if: github.actor != 'dependabot[bot]'
+    name: Publish
+    runs-on: ubuntu-latest
+    environment: ECR Release Publishing
+    strategy:
+      matrix:
+        app: [ legacy-api, ocean-api, playground-api, status-api ]
```
jingyi2811, line 22: I am unsure should this includes `rich-list-api`. What do you think?

jingyi2811 (dismissed):

.github/workflows/release-ecr.yml
cwkang1998: Regarding `rich-list`, since its currently on hold right now and not completed (see https://github.com/JellyfishSDK/jellyfish/pull/1173#issuecomment-1110768560), I think we can exclude it for now and add it later.

cwkang1998 (approved):

.github/scripts/release-tags.js
```diff
@@ -55,8 +55,8 @@ function getDomain(context) {
 }
 
 function getReleaseTag(domain, app, context) {
-  const semver = context.payload.release.tag_name.replace('v', '')
-  if (semver.match(/^[0-9]+\.[0-9]+\.[0-9]+$/) === false) {
+  const semver = context.payload.release.tag_name
+  if (semver.match(/^v[0-9]+\.[0-9]+\.[0-9]+$/) === null) {
```
fuxingloh, line 59: The `v` is replaced previously, but change keeps the v? This might not be intentional?

fuxingloh: Also ideally `v` should not be kept for docker release tag

fuxingloh (changes requested):

.github/scripts/release-tags.js
nichellekoh: If that is the case, will do the regex check as suggested by @jingyi2811 first, then return the version with the `v` removed

jingyi2811: Okay, I was misled by the `replace v with empty string` code.

jingyi2811: Why don't remove v from the regular expression?

jingyi2811: I fell like I probably misunderstanding something

nichellekoh: Either way it should work as the initial replace function implementation only replaces the first v (as mentioned in the previous thread). https://github.com/JellyfishSDK/jellyfish/commit/69cb57d7b487972f230d058449422fb854e083c3 will check the expression before replacing the `v` which works fine too. Actually, from my understanding it will prevent 1 additional string scenario which is e.g. `1.1.1` (initial semver without the v) from passing the regex condition

fuxingloh (approved):

cwkang1998 (approved):

PR #1413, opened 2022-05-06
fix(legacy_api): Updated v1 and v2 /listswaps regex to match tickers with a period

joeldavidw:
Updated v1 and v2 /listswaps regex to match tickers with a period


eli-lim (approved):
:shipit:

fuxingloh (approved):

PR #1433, opened 2022-05-17
chore(whale/sanity-testing): port over tests from whale ci newman

eli-lim:
Ports over the sanity tests from https://github.com/JellyfishSDK/whale/blob/main/.github/workflows/ci/postman_collection.json as typescript + jest tests

Part of #977 

#### Additional comments?:
- One thing to note is that the CI pipeline now builds whale twice. We can investigate in a future PR the possibility of sharing the build (output of the `build` workflow) with the sanity test env.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,6 +49,20 @@ jobs:
         with:
           fail_ci_if_error: true
 
+  docker-sanity:
+    needs: [ build, lint ]
+    name: Docker Sanity Test
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@2541b1294d2704b0964813337f33b291d3f8596b
+      - uses: actions/setup-node@17f8bd926464a1afa4c6a11669539e9c1ba77048
+        with:
+          node-version: '16'
+          cache: 'npm'
+
+      - run: npm ci
+      - run: npm run sanity
+
   lint:
     name: Lint
     runs-on: ubuntu-latest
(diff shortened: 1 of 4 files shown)


jingyi2811:
Every time I rerun the test, the system will build the image again. I wonder if the following code can change from `docker.buildimage` to `docker.getimage`
See https://github.com/JellyfishSDK/jellyfish/blob/main/jest.sanity.setup.js#L26

Refer to https://github.com/JellyfishSDK/jellyfish/blob/main/packages/testcontainers/src/containers/DockerContainer.ts#L141

eli-lim:
> Every time I rerun the test, the system will build the image again. I wonder if the following code can change from `docker.buildimage` to `docker.getimage`
> 
> if the image was built BEFORE

If the underlying (whale) code changes, will docker be aware and rebuild the image?

canonbrother (approved):

jingyi2811:
> > Every time I rerun the test, the system will build the image again. I wonder if the following code can change from `docker.buildimage` to `docker.getimage`
> > if the image was built BEFORE
> 
> @jingyi2811 thanks for the suggestion. But if the underlying (whale) code changes, will docker be aware and rebuild the image?

If there are code changes, a new docker name needs to be given and rebuild is needed.

jingyi2811:
If it is too hard then just ignore it, I am using an old Mac book with an Intell processor. Running tests on every test unit takes me 5 to 10 minutes.

kodemon:
There is definitely room for optimization improvements to the sanity layer dedicating more time to the process. The current main use case for our whale sanity tests is mostly in our devops/releases processes. It's separated from the standard unit tests, as its a pretty heavy handed process so we decided to make it opt in. _(No one wants to wait 5 - 10 minutes when running local unit tests)_ 😅

kodemon (approved):

jingyi2811 (approved):

ivan-zynesis (approved):

PR #1448, opened 2022-05-19
fix(whale-migration): fix issues caused by reverting initial whale-api-client refactoring

eli-lim:
- Fixes lint and build failing due to imported 'Method' that wasn't updated during the whale-api-client refactoring reversion
- Fixes WhaleApiClient not enforcing default values, causing status-api test failure due

Part of #977


packages/whale-api-client/src/WhaleApiClient.ts
```diff
@@ -80,7 +80,13 @@ export class WhaleApiClient {
   constructor (
     protected readonly options: WhaleApiClientOptions
   ) {
-    this.options = { ...DEFAULT_OPTIONS, ...options }
```
eli-lim, line 83: passing in `{ network: undefined }` will override the default value 'mainnet'

kodemon (approved):

PR #1461, opened 2022-05-23
fix(jellyfish-api-core, jellyfish-transaction-builder): Fix failed tests for fortcanningcrunch

surangap:

Fixes #https://github.com/DeFiCh/ain/pull/1243


packages/testcontainers/src/containers/RegTestContainer/index.ts
```diff
@@ -35,7 +35,8 @@ export class RegTestContainer extends DeFiDContainer {
       '-fortcanningheight=8',
       '-fortcanningmuseumheight=9',
       '-fortcanninghillheight=10',
-      '-fortcanningroadheight=11'
+      '-fortcanningroadheight=11',
+      '-fortcanningspicegardenheight=12'
```
surangap, line 39:
```suggestion
      '-fortcanningcrunchheight=12'
```

packages/jellyfish-api-core/__tests__/category/account/getAccountHistory.test.ts
```diff
@@ -44,8 +44,6 @@ describe('Account', () => {
     expect(history.txn).toStrictEqual(referenceHistory.txn)
     expect(history.txid).toStrictEqual(referenceHistory.txid)
     expect(history.type).toStrictEqual(referenceHistory.type)
-    expect(history.blockHash).toStrictEqual(referenceHistory.blockHash)
```
ivan-zynesis, line 47: Might be better to keep another history which in a minted block and non undefined block hash

packages/jellyfish-api-core/__tests__/category/loan/paybackLoan.test.ts
```diff
@@ -888,327 +888,328 @@ describe('paybackLoan failed', () => {
...
-      vaultId: vaultId,
-      amounts: '2@TSLA',
-      from: colAddr
-    })
-
-    await expect(promise).resolves.not.toThrow()
-  })
-})
+// Jimmy: Temporarily comment this code, as it should throw error
+// describe('paybackLoan before FortCanningHeight', () => {
```
ivan-zynesis, line 892: Shall we just `skip` it? Thus remained visible in test report.

apps/playground-api/src/setups/setup.gov.ts
```diff
@@ -64,12 +64,7 @@ export class SetupGov extends PlaygroundSetup<Record<string, any>> {
   }
 
   async has (each: any): Promise<boolean> {
-    const key = Object.keys(each)[0]
-    const gov = await this.client.masternode.getGov(key)
-    if (Object.keys(gov[key]).length > 0) {
-      return true
-    }
-    return false
+    return false // for just overwrite
```
jingyi2811, line 67: If return `false` is something expected, we should remove this function

jingyi2811: And the function name is `has`,  it is a positive word so return true seems more sensible

apps/playground-api/src/setups/setup.ts
```diff
@@ -33,10 +33,6 @@ export abstract class PlaygroundSetup<Each> {
     await this.before(list)
 
     for (const each of list) {
-      if (await this.has(each)) {
```
jingyi2811, line 36: @canonbrother Here, I changed your code.  As this has function always returns false, this code can be removed.

packages/jellyfish-api-core/__tests__/category/vault/placeAuctionBid.test.ts
```diff
@@ -464,13 +478,13 @@ describe('placeAuctionBid success', () => {
           batches: [
             {
               index: 0,
-              collaterals: ['5004.44437154@DFI', '0.49955556@BTC'], // 5004.44437154 + 4995.5556 = 9999.99997154
-              loan: '499.56012216@TSLA' // 499.56012216 * 15 = 7493.4018324
+              collaterals: ['5004.44449284@DFI', '0.49955555@BTC'],
```
jingyi2811, line 481: @surangap  All these values should remain the same.

apps/playground-api/src/setups/setup.ts
canonbrother: don't change that code.. its by designed

canonbrother (changes requested):
https://github.com/JellyfishSDK/jellyfish/pull/1461/files#r882297326

apps/playground-api/src/setups/setup.ts
jingyi2811: Okay

packages/jellyfish-api-core/__tests__/category/vault/placeAuctionBid.test.ts
surangap: Thought this was reverted back. should be fine now.

jingyi2811 (approved):

canonbrother (dismissed):

fuxingloh (approved):

fuxingloh:
I will update `activateAfterBlock` on WhaleApi controller such that it returns `activateAfterBlock: 0` to prevent downstream regression.

PR #1466, opened 2022-05-23
chore(whale-api-client): Default whale api client to v0 and let url be optional

nichellekoh:
Currently WhaleApiClient options version default to the current npm release version. I think to ensure forever backward compatibility for users, we should default to v0.

Duplicate of https://github.com/JellyfishSDK/whale/pull/980, changes raised here as part of migration

Fixes https://github.com/JellyfishSDK/whale/issues/979

### DIFF
--- a/packages/whale-api-client/src/WhaleApiClient.ts
+++ b/packages/whale-api-client/src/WhaleApiClient.ts
@@ -80,6 +80,7 @@ export class WhaleApiClient {
     protected readonly options: WhaleApiClientOptions
   ) {
     this.options = { ...DEFAULT_OPTIONS, ...options }
+    this.options.url = this.options.url?.replace(/\/$/, '')
   }
 
   /**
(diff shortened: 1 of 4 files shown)


packages/whale-api-client/src/WhaleApiClient.ts
```diff
@@ -81,7 +80,6 @@ export class WhaleApiClient {
     protected readonly options: WhaleApiClientOptions
   ) {
     this.options = { ...DEFAULT_OPTIONS, ...options }
-    this.options.url = this.options.url.replace(/\/$/, '')
```
fuxingloh, line 84: Should keep this so that the last `/` is trimmed

fuxingloh (approved):

fuxingloh:
@eli-lim note this change is applied on both Whale & Jellyfish side.

PR #1504, opened 2022-06-01
chore(apps/whale-api): Future swap indexer

### LINKED ISSUE #1484: Indexing FutureSwap on Whale Indexer

Instead of relying on defid for FutureSwap, move the indexing of FutureSwap to whale via and append-only indexing Add/Remove DfTx on whale indexer. 

**keys:**
- partition: address
- sort: height + txid

**notable fields:**
- type: `add` or `remove`
- everything else

**Controller Querying?**

For `AddressController.listFutureSwap`, given that FutureSwap goes into effect at a target block height, the controller should query from target height to minus interval height

/triage accepted
/area module-indexer module-api whale-api-client
/priority important-soon
/assign @canonbrother

canonbrother:
Code migration from https://github.com/JellyfishSDK/whale/pull/940

Fixes #1484


packages/whale-api-client/src/api/Address.ts
```diff
@@ -94,13 +94,25 @@ export class Address {
...
    */
   async listTransactionUnspent (address: string, size: number = 30, next?: string): Promise<ApiPagedResponse<AddressUnspent>> {
     return await this.client.requestList('GET', `address/${address}/transactions/unspent`, size, next)
   }
+
+  /**
+   * List future swap
+   *
+   * @param {string} address to list future swap
+   * @param {string} [size=30] to query
```
jingyi2811, line 109:
```suggestion
   * @param {number} [size=30] to query
```

apps/whale-api/src/module.api/address.controller.e2e.ts
```diff
@@ -960,3 +963,247 @@ describe('listTokens', () => {
...
+    const list = await controller.listFutureSwap(fromAddr, {
+      size: 30
+    })
+
+    {
+      const first = await controller.listFutureSwap(fromAddr, {
+        size: 3
+      })
+      expect(first.data.length).toStrictEqual(3)
+      expect(first.page?.next).not.toBeUndefined()
```
jingyi2811, line 1169:
```suggestion
      expect(first.page?.next).toBeDefined()
```

apps/whale-api/src/module.api/address.controller.e2e.ts
```diff
@@ -9,6 +9,9 @@ import { Testing } from '@defichain/jellyfish-testing'
 import { ForbiddenException } from '@nestjs/common'
 import BigNumber from 'bignumber.js'
 import { RegTestFoundationKeys } from '@defichain/jellyfish-network'
+import { FutureSwap } from '@defichain/jellyfish-api-core/src/category/account'
+
+/* eslint-disable no-lone-blocks */
```
jingyi2811, line 14: I think this can be removed.

jingyi2811:
It seems like listFutureSwap requires address as input. I wonder can we list all the items without an address?

apps/whale-api/src/module.api/address.controller.e2e.ts
canonbrother: we have restriction
https://github.com/JellyfishSDK/jellyfish/blob/0d8264f97b6297c667681414364e52b0b26b0056/.eslintrc#L62

canonbrother: hmm.. that part is not under coverage

canonbrother:
> It seems like listFutureSwap requires address as input. I wonder can we list all the items without an address? Because every time we call listFutureSwap, it actually meant for 1 week only from my understanding.

I think `without an address` that's too heavy and not user-friendly

apps/whale-api/src/module.api/address.controller.e2e.ts
jingyi2811: Okay, my bad. You are right.

jingyi2811 (dismissed):

kodemon (approved):

PR #1512, opened 2022-06-02
chore(packages/whale-api-*): revert whale packages file name casing

### LINKED ISSUE #1511: `packages/whale-api-*` filename casing change from lower to upper breaks client code outside of jellyfish monorepo
#### What happened:
`whale-api-client` and `whale-api-wallet` file name change (lower dot notation casing to pascal casing) will break client code outside of jellyfish monorepo that switch over from whale

```typescript
// client code relying on whale repo would have their imports written like this, 
// which will break when they switch to jellyfish
import { TokenData } from '@defichain/whale-api-client/src/api/tokens'

// they'll all need to be updated -
import { TokenData } from '@defichain/whale-api-client/src/api/Tokens'
```

No breaking due to unresolved imports

For backwards compatibility, we should revert the filename casing change.

eli-lim:
- Revert file name changes done early on in the migration

Fixes #1511


fuxingloh:
Maybe only exported `packages/*` should be reverted?

eli-lim:
> Maybe only exported `packages/*` should be reverted?

some changes are necessary in `apps/*` as well since they import interfaces from whale-api-client's files

fuxingloh (approved):

PR #1514, opened 2022-06-02
feat(migrate-getsubgraphswaps): Migrated getsubgraphswaps logic

### LINKED ISSUE #1483: `GET /dex/swaps/history`

For short; the implementation of `https://api.defichain.com/v1/getsubgraphswaps` on Whale for it to enjoy the low latency lookup of `AccountHistory` and LevelDB indexes. Using the exact same cursor pagination methodology that was done on legacy-api but without caching. Caching to be moved to `legacy-api`, Whale will provide up the date real-time resolution of data taking advantage of low latency LevelDB lookup.

https://github.com/JellyfishSDK/jellyfish/blob/8ab775c29b627e0f5cd8bf9c2e8b7d854ff5137d/apps/legacy-api/src/controllers/PoolPairController.ts#L90-L112

#### Why is this needed:

For the time being, to quickly improve over the wire performance of `v1/getsubgraphswaps` API without investing in a full-fledge indexer. 

/triage accepted
/proprity urgent-now
/packages module-api whale-api-client

chanyusiang:
- Migrated the getSubgraphSwaps logic from legacy-api to whale-api to quickly improve over the wire performance of `v1/getsubgraphswaps` API without investing in a full-fledge indexer

- Fixes #1483 

#### Additional comments?:
- A direct HTTP call from legacy-api to whale-api is made (instead of invoking whale-api-client) as the new endpoint in the latter is meant for exclusive use by the former
- **[Important]**: The `getsubgraphswaps` endpoint in `legacy-api` _(https://api.defichain.com/v1/getsubgraphswaps)_ is now dependent on a `whale-api` endpoint _(https://ocean.defichain.com/v0/mainnet/legacy/getsubgraphswaps)_ that does not exist at this moment. Therefore, the new endpoint in whale-api has to be deployed before the changes in legacy-api can be tested and deployed as well


eli-lim:
legacy-api poolpaircontroller test is failing - can we mock the response from ocean?

apps/legacy-api/src/pipes/NetworkValidationPipe.ts
```diff
@@ -5,14 +5,15 @@ import {
   PipeTransform
 } from '@nestjs/common'
 
-export type SupportedNetwork = 'mainnet' | 'testnet'
+export type SupportedNetwork = 'mainnet' | 'testnet' | 'regtest'
```
eli-lim, line 8: I don't think there's a need for this 
```suggestion
export type SupportedNetwork = 'mainnet' | 'testnet'
```

apps/whale-api/src/module.api/_module.ts
```diff
@@ -31,6 +31,8 @@ import { LoanController } from './loan.controller'
 import { LoanVaultService } from './loan.vault.service'
 import { PoolSwapPathFindingService } from './poolswap.pathfinding.service'
 import { PoolPairPricesService } from './poolpair.prices.service'
+import { LegacyController } from './legacy.controller'
+import { SimpleCache } from '../../../legacy-api/src/cache/SimpleCache'
```
eli-lim, line 35: We should move `SimpleCache` into libs/caches

apps/whale-api/src/module.api/legacy.controller.spec.ts
```diff
@@ -0,0 +1,98 @@
...
+  expect(subgraphSwaps.data.swaps.length).toStrictEqual(30)
+})
+
+it('should list 5 subgraph swaps', async () => {
+  controller = app.get(LegacyController)
+
+  const subgraphSwaps = await controller.getSubgraphSwaps('regtest', 5)
+
+  expect(subgraphSwaps.data.swaps.length).toStrictEqual(5)
+})
```
eli-lim, line 98: Can we move the test cases from `legacy-api/__tests__/controllers/PoolPairController.test.ts` here?

eli-lim (changes requested):

eli-lim (dismissed):
Need some reviews since I took over this PR

apps/whale-api/src/module.api/legacy.controller.ts
```diff
@@ -0,0 +1,334 @@
...
+
+@Controller('/legacy')
+export class LegacyController {
+  private readonly logger = new Logger(LegacyController.name)
+  constructor (
+    protected readonly rpcClient: JsonRpcClient,
+    protected readonly blockMapper: BlockMapper,
+    protected readonly transactionMapper: TransactionMapper,
+    protected readonly transactionVoutMapper: TransactionVoutMapper,
+    private readonly cache: SimpleCache,
```
fuxingloh, line 50: I think we should remove the cache in `whale-api` and let it hit defid for the account history data? Otherwise, it would compete with whale-api cache which is critically more important.

fuxingloh: Unless we can separate the cache layer since they compete with the same cache provider size.

apps/whale-api/src/module.api/legacy.controller.ts
```diff
@@ -0,0 +1,334 @@
...
+    private readonly network: NetworkName
+  ) {
+  }
+
+  @Get('getsubgraphswaps')
+  async getSubgraphSwaps (
+    @Query('limit') limit: number = 30,
+    @Query('next') nextString?: string
+  ): Promise<LegacySubgraphSwapsResponse> {
+    limit = Math.min(100, limit ?? 30)
```
fuxingloh, line 61:
```suggestion
    @Query('limit') limit: number = 20,
    @Query('next') nextString?: string
  ): Promise<LegacySubgraphSwapsResponse> {
    limit = Math.min(20, limit ?? 20)
```

Further limit to 20?

apps/whale-api/src/module.api/legacy.controller.ts
```diff
@@ -0,0 +1,334 @@
...
+      data: {
+        swaps: swaps
+      },
+      page: {
+        next: encodeBase64(next)
+      }
+    })
+  }
+
+  private async getSwapsHistory (
```
fuxingloh, line 81: Possible to separate this part and everything below into `legacy.subgraph.service.ts`?

apps/whale-api/src/module.api/legacy.controller.ts
fuxingloh: Currently, on `poolpair.controller.ts`, we have a simple implementation (also account history) without caching and it performs fine since everything is enclaved into a single `ocean: whale + defid` group within the local network hence the fast lookup.

fuxingloh (changes requested):
Some thoughts about it

apps/whale-api/src/module.api/legacy.controller.ts
eli-lim: will remove and verify the performance. I think it's fine since we'll also be limiting to 20 swaps each query

apps/libs/caches/src/LegacyCache.ts
```diff
@@ -1,16 +1,24 @@
...
-  ) {
-  }
+export class LegacyCache {
+  // An encapsulated cache that is not dependent on global CacheModule
+  // which the other services depend on
+  private readonly cacheManager: Cache = caching({
+    store: 'memory',
+    ttl: 600, // 10m
+    max: 10_000
+  })
```
eli-lim, line 21: manually verified this by comparing the cache keys with defid cache's keys - they're different

apps/whale-api/src/module.api/legacy.controller.ts
eli-lim: added cache back as performance is unacceptable without it

fuxingloh (approved):

canonbrother (approved):

PR #1526, opened 2022-06-04
fix(jellyfish-api-core): Fix failed tests after connecting to the latest ain code

jingyi2811:
Fix failed tests after connecting to the latest ain code.
See https://github.com/DeFiCh/ain/pull/1320

Fixes #


packages/jellyfish-api-core/__tests__/category/vault/listAuctionHistory.test.ts
```diff
@@ -1,7 +1,7 @@
 import BigNumber from 'bignumber.js'
 import { TestingGroup } from '@defichain/jellyfish-testing'
 import { GenesisKeys, MasterNodeRegTestContainer } from '@defichain/testcontainers'
-import { ListAuctionHistoryDetail, VaultLiquidation } from '../../../src/category/vault'
+import { ListAuctionHistoryDetail, VaultLiquidation } from '../../../src/category/loan'
```
ivan-zynesis, line 4: `vault` is refactored version, revert this line, test should remained working.

ivan-zynesis, line 4:
```suggestion
import { ListAuctionHistoryDetail, VaultLiquidation } from '../../../src/category/vault'
```

ivan-zynesis (changes requested):

packages/jellyfish-api-core/__tests__/category/loan/listAuctionHistory.test.ts
```diff
@@ -615,25 +615,25 @@ describe('Loan listAuctionHistory', () => {
           { index: 1 }
         )
 
-        expect(page.length).toStrictEqual(4)
+        expect(page.length).toStrictEqual(2)
```
surangap, line 618:
```
// This test shows that we can not use index only for the pagination start point as the DB composite index starts with maxBlockHeight.
```
so earlier, it did not work with `index` only and now it works? if then the comment should be removed.
Wondering what changed it. likewise does "should not listAuctionHistory with vaultId only" also changed?

surangap:
> Since it's related to product APIs, I need to understand the context of this change. What caused the size of the auction history to change in this release?

yes, something is not right. let me check

surangap:
> > Since it's related to product APIs, I need to understand the context of this change. What caused the size of the auction history to change in this release?
> 
> yes, something is not right. let me check cpp side changes



> > Since it's related to product APIs, I need to understand the context of this change. What caused the size of the auction history to change in this release?
> 
> yes, something is not right. let me check cpp side changes

https://github.com/DeFiCh/ain/pull/1215/files
Looks like this implements a filter for `BatchIndex` which is not what we want. what we need is pagination, not filtration.

Issue #1532, opened 2022-06-06
Major Regression for `apps/whale-api` due to untracked objects on TokenMapper & PoolMapper


fuxingloh:
#### What happened:

> See https://github.com/JellyfishSDK/jellyfish/pull/1528 for temporary mitigation. We will work on a fix, a re-index will be required. Please use the preview builds for production with the hot-fixed static code. If you're using a snapshot please use a snapshot earlier than height `1948150`.

The `apps/whale-api` crashed at block height `1948512`.

There is a Major Regression for `apps/whale-api` due to untracked objects on TokenMapper & PoolMapper that affected PoolSwap and CompositeSwap indexing. The bug was due to the stock split mechanism that generated 2 coinbase transactions without its accompanying DfTx to indicate what's the event.

/assign @fuxingloh 
/triage accepted
/priority urgent-now
/area apps
/apps whale-api

fuxingloh:
This also caused 10-15mins downtime on ocean, see https://status.defichain.com/incidents/b15s181nh2j1

fuxingloh:
Please use https://github.com/JellyfishSDK/jellyfish/pkgs/container/whale-api/24365785?tag=88972089d0e439aee173fde0a7fd7cc918c1da79 meanwhile while we work on a proper fix.

fuxingloh:
Currently these indexer contributes to the `TokenMapper`:
- `CreateTokenIndexer`
- `CreatePoolPairIndexer`
- `SetLoanTokenIndexer`

Now we have confirmed, `SetGov` for StockSplit affects it too.

We also need to check if there is any regression with how `LoanToken` get created.

sancag:
Hi, will a complete re-index be required or is it enough to use a snapshot prior to  the failing block?

fuxingloh:
Snapshot prior to `1948150` is enough.

fuxingloh:
> Hi, will a complete re-index be required or is it enough to use a snapshot prior to the failing block?

Apologize for the inconvenience, this was caught on the eleventh hour by @ivan-zynesis (thankfully otherwise it would have been worst).

lukasosterheider:
> Nothing to apologize. This is a complex topic that has not been done before and you guys are super quick in finding a solution.
> 
[...]

PR #1528, opened 2022-06-06
fix(indexer): regression missing object for create.pool.pair and create.token

fuxingloh:
#### What this PR does / why we need it:

https://status.defichain.com/incidents/b15s181nh2j1

Major Regression for PoolSwap & CompositeSwap indexing due to missing objects that aren't monitored and picked up by the Whale Indexer.

Please use the preview builds for production with the hot-fixed static code. If you're using a snapshot please use a snapshot earlier than height `1948150`.

#### Additional comments?:

This was spotted hours before the unlock of stock split thus the rushed hotfix release.

PR #1533, opened 2022-06-07
fix(whale-api): change to rely on defidcache instead of tokenMapper

eli-lim:
Fixes a regression by changing dexPrices calculator to rely on defid instead of tokenMapper

#### Background
- Problem: token 49 (amzn/v1) and 89 (amzn) share the same `symbol` in `TokenMapper` due to #1532
  - dexPrices logic depends on a map indexed by `symbol` - there was an assumption that `symbols` would be unique
  - as there's a name collision (both have `symbol` = 'AMZN'), token 89 is overwritten by 49 in the dexPrices map, hence it never gets returned in the endpoint response
- Solution: change the dexPrices logic to rely solely on defid cache, remove any dependency on `TokenMapper`
  - This is also more of a __future-proof__ fix in case similar token splits occur

Fixes a regression due to #1528

### DIFF
--- a/apps/whale-api/src/module.api/poolpair.prices.service.ts
+++ b/apps/whale-api/src/module.api/poolpair.prices.service.ts
@@ -38,8 +38,8 @@ export class PoolPairPricesService {
 
     // Get denomination token info
     const denominationToken = await this.getTokenBySymbol(denominationSymbol)
-    if (denominationToken === undefined || this.shouldIgnoreToken(denominationToken)) {
-      throw new Error(`Unexpected error: could not find token with symbol '${denominationSymbol}'`)
+    if (this.isUntradeableToken(denominationToken)) {
+      throw new UntradeableTokenError(denominationSymbol)
     }
 
     // Get all tokens and their info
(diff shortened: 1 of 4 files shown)


apps/whale-api/src/module.api/poolpair.prices.service.ts
```diff
@@ -73,32 +64,34 @@ export class PoolPairPricesService {
...
+      throw new TokenSymbolNotFoundError(symbol)
     }
-    return allTokenInfo
+
+    const [id, tokenInfo] = Object.entries(tokenResult)[0]
+    return { ...tokenInfo, id }
+  }
+
+  private isUntradeableToken (token: TokenInfoWithId): boolean {
+    return token.isLPS || !token.isDAT || token.symbol === 'BURN'
```
canonbrother, line 94: hmm... not sure want to hide old token(eg: AMZN-DUSD/v1) as well?? :thinking: 
https://ocean.defichain.com/v0/mainnet/poolpairs/54

fuxingloh (dismissed):

canonbrother (dismissed):

fuxingloh:
Could we take out:

- `apps/whale-api/src/module.indexer/model/dftx/create.pool.pair.ts`
- `apps/whale-api/src/module.indexer/model/dftx/create.token.ts`

Merge this, and then rebase onto https://github.com/JellyfishSDK/jellyfish/pull/1528 instead, since this is a fix?

Let me do it!

canonbrother (approved):

fuxingloh (approved):

PR #1539, opened 2022-06-09
fix(whale-api): hardcode amznv1 symbol

eli-lim:

Fixes #


fuxingloh:
Merged #1533 instead.

PR #1547, opened 2022-06-12
fix(apps/whale-api): remove Token & PoolPair indexers and resolve from defid directly

### LINKED ISSUE #1544: Remove TokenMapper & PoolPairMappers and its indexer, using Cached RPC to resolve Token & PoolPair
> Following the https://github.com/JellyfishSDK/jellyfish/issues/1532 regression, looking at simplification of Token & PoolPair Mapper design.

Given that `TokenMapper` and `PoolPairMapper` were created solely for `PoolSwapIndexer` and its continuous regression battle. I propose we remove the indexing of Token and PoolPair and just resolve them via "Cached RPC Service".

We can remove 4 indexers:
1. CreateTokenIndexer
1. CreatePoolPairIndexer
1. UpdatePoolPairIndexer
1. SetLoanTokenIndexer

And 3 mappers:

1. TokenMapper
1. PoolPairHistoryMapper
1. PoolPairTokenMapper

#### Why is this needed:

Battling change and regression on defid is a uphill battle, requiring our `whale-api` indexer to always be up to date is resource-intensive. Similar to #1533, this change further makes `whale-api` more little weight and resistant to upstream regression.

/triage accepted
/assign @canonbrother 
/priority urgent-now
/area apps
/apps whale-api

canonbrother:

Fixes #1544


apps/whale-api/src/module.api/cache/defid.cache.ts
```diff
@@ -104,13 +109,19 @@ export class DeFiDCache extends GlobalCache {
...
+   * @return {PoolPairInfoWithId | undefined}
+   */
+  private async fetchPoolPairInfo (idOrSymbol: string): Promise<PoolPairInfoWithId | undefined> {
     try {
-      const result = await this.rpcClient.poolpair.getPoolPair(id)
-      if (result[id] === undefined) {
-        return undefined
+      const result = await this.rpcClient.poolpair.getPoolPair(idOrSymbol)
+      const [id, poolPairInfo] = Object.entries(result)[0]
+      return {
```
kodemon, line 121: This no longer seem to be checking for potential `undefined` results. Considering this seems deliberate this is most likely as per design, but wanted to point this out in case this may need to be addressed.

apps/whale-api/src/module.indexer/model/dftx/composite.swap.spec.ts
```diff
@@ -32,6 +32,15 @@ beforeEach(async () => {
   }
 
   await container.generate(1)
+
+  app = await createTestingApp(container)
+  const defiCache = app.get(DeFiDCache)
+
+  const tokenResult = await container.call('listtokens')
+  // precache
```
ivan-zynesis, line 40: Curious if this is needed?

apps/whale-api/src/module.api/poolpair.service.ts
```diff
@@ -151,7 +149,7 @@ export class PoolPairService {
 
   private async getTokenUSDValue (id: number): Promise<BigNumber | undefined> {
     return await this.cache.get<BigNumber>(`PRICE_FOR_TOKEN_${id}`, async () => {
-      const tokenInfo = await this.tokenMapper.getByTokenId(id)
+      const tokenInfo = await this.deFiDCache.getTokenInfo(`${id}`)
```
kodemon, line 152: `nit:` Looking at this method now it does not seem like the `id` is used as a number anymore. So wondering if perhaps instead of converting the incoming id from `number` to `string`. To just set the expected id value to `string` in the method instead? 🤔

apps/whale-api/src/module.indexer/model/dftx/composite.swap.ts
```diff
@@ -44,33 +40,34 @@ export class CompositeSwapIndexer extends DfTxIndexer<CompositeSwap> {
...
+    }
+
+    const toToken = await this.deFiDCache.getTokenInfo(`${poolSwap.toTokenId}`)
+    if (toToken === undefined) {
+      throw new NotFoundApiException('Unable to find toToken')
+    }
+
+    let poolPair = await this.deFiDCache.getPoolPairInfo(`${fromToken.symbol}-${toToken.symbol}`)
+    if (poolPair === undefined) {
+      poolPair = await this.deFiDCache.getPoolPairInfo(`${toToken.symbol}-${fromToken.symbol}`)
```
ivan-zynesis, line 65: I think this will return pool pair after split instead of before, but probably no better solution now.

ivan-zynesis (changes requested):
Not this PR, but can quickly fix L157

```
next += Number(poolPairIds[poolPairIds.length - 1])
```

apps/whale-api/src/module.api/poolswap.pathfinding.service.ts
```diff
@@ -23,15 +22,21 @@ export class PoolSwapPathFindingService {
...
-    protected readonly poolPairTokenMapper: PoolPairTokenMapper,
     protected readonly deFiDCache: DeFiDCache,
     @Inject('NETWORK') protected readonly network: NetworkName
   ) {
   }
 
   @Interval(120_000) // 120s
   async syncTokenGraph (): Promise<void> {
-    const poolPairTokens = await this.poolPairTokenMapper.list(200)
+    const poolPairsResult = await this.deFiDCache.listPoolPairs(60) as PoolPairsResult
```
kodemon, line 32: `question:` if the method being called can potentially return `undefined`, should this be accounted for in this method by doing a defensive check?

Providing a `undefined` value to `Object.entries` would produce an error in this method.

apps/whale-api/src/module.indexer/model/dftx/composite.swap.spec.ts
canonbrother: nope.. i think i messed up something previously.. removed

apps/whale-api/src/module.api/cache/defid.cache.ts
canonbrother: hmm.. i think its okay as its wrapped inside a `try..catch` for rpc call, if getting `not found` from rpc will return undefined, else throw err

apps/whale-api/src/module.indexer/model/dftx/composite.swap.ts
fuxingloh: Hmmm

canonbrother:
> +=

@ivan-zynesis 
a bit weird.. `+=`...
```
// assume next is 7 for now

next += lastPoolPairId // 20 + 7

// the "next" will be started from poolPairdId 27
```

kodemon (approved):

kodemon (approved):

apps/whale-api/src/module.indexer/model/dftx/composite.swap.ts
fuxingloh: Might be problematic, but I think we can live with this

fuxingloh:
Let me see if I can further simplify this PR

fuxingloh:
Oh @canonbrother there is a merge conflict too

fuxingloh:
I'm to shift-left merge changes that are unrelated but required for #1547 to reduce the scope of this PR to get it moving faster. Reviewing this is getting complex due to the edge cases that we need to look at.

fuxingloh:
With #1561 merged can you de-conflict this?

apps/whale-api/src/e2e.module.ts
```diff
@@ -59,7 +61,13 @@ export async function stopTestingApp (container: MasterNodeRegTestContainer | Te
...
 async function createTestingModule (url: string): Promise<TestingModule> {
   return await Test.createTestingModule({
-    imports: [AppModule.forRoot('memory')]
+    imports: [
+      CacheModule.register(),
+      AppModule.forRoot('memory')
+    ],
+    providers: [
+      DeFiDCache
+    ]
```
fuxingloh, line 70: since we globally export in `apps/whale-api/src/module.api/_module.ts`, we don't need to export this here again?

fuxingloh: I will test this out on https://github.com/JellyfishSDK/jellyfish/pull/1566

apps/whale-api/src/e2e.module.ts
```diff
@@ -10,6 +10,8 @@ import waitForExpect from 'wait-for-expect'
 import { addressToHid } from './module.api/address.controller'
 import { ScriptAggregationMapper } from './module.model/script.aggregation'
 import { TestingGroup } from '@defichain/jellyfish-testing'
+import { DeFiDCache } from './module.api/cache/defid.cache'
+import { CacheModule } from '@nestjs/common'
```
fuxingloh, line 14:
```suggestion
```

apps/whale-api/src/e2e.module.ts
fuxingloh:
```suggestion
    imports: [AppModule.forRoot('memory')]
```

apps/whale-api/src/module.indexer/model/dftx/pool.swap.aggregated.ts
```diff
@@ -21,25 +21,25 @@ export class PoolSwapAggregatedIndexer extends DfTxIndexer<PoolSwap> {
...
   }
 
   async indexBlockStart (block: RawBlock): Promise<void> {
-    const poolPairs = await this.poolPairTokenMapper.listAllDesc()
+    const poolPairs = await this.deFiDCache.getPoolPairs()
 
     for (const interval of AggregatedIntervals) {
       for (const poolPair of poolPairs) {
-        const previous = await this.aggregatedMapper.query(`${poolPair.poolPairId}-${interval as number}`, 1)
+        const previous = await this.aggregatedMapper.query(`${poolPair.id}-${interval as number}`, 1)
```
fuxingloh, line 34:
```suggestion
        if (poolPair.creationHeight.gte(block.height)) {
          // Don't create bucket for PoolPair if block.height is greater
          continue
        }
        
        const previous = await this.aggregatedMapper.query(`${poolPair.id}-${interval as number}`, 1)
```

packages/whale-api-client/__tests__/api/poolpairs.test.ts
```diff
@@ -770,6 +770,16 @@ describe('poolswap aggregated', () => {
...
+      {
+        aggregated: {
+          amounts: {},
+          usd: 0
+        },
+        block: expect.any(Object),
+        bucket: expect.any(Number),
+        id: expect.any(String),
+        key: '10-86400'
+      },
```
fuxingloh, line 782:
```suggestion
```

fuxingloh (commented):
Might fix the test regression below, where we only index the bucket if the `PoolPair is created before the block.height`.

fuxingloh:
> Might fix the test regression below, where we only index the bucket if the `PoolPair is created before the block.height`.

Let me test on `#1547`, before I commit into this branch.

apps/whale-api/src/module.indexer/model/dftx/composite.swap.ts
fuxingloh: https://github.com/JellyfishSDK/jellyfish/pull/1567 fixes it

fuxingloh:
> > Might fix the test regression below, where we only index the bucket if the `PoolPair is created before the block.height`.
> 
> Let me test on #1566, before I commit into this branch.

Hmm, the test is very flaky when we remove this. Let me merge this in first and let's see how we can deal with this change later.

fuxingloh (approved):

PR #1560, opened 2022-06-15
fix(whale-api): network value being overwritten by joi

eli-lim:
- Fixes the `network` config being overwritten by Joi schema validation
- NestJS's ConfigModule expects the Joi schema to validate `process.env`, not the app config object

#### In Detail
- ConfigModule will use `Joi.object({ network: Joi.string().default('regtest')})` to validate process.env
- It checks `network` and sees that it's missing from `process.env` (as the env variable is `WHALE_NETWORK`)
- So it sets the default value provided - `network: 'regtest'` - in ConfigService, overwriting the value provided by `load: [AppConfiguration]`

#### 1. Setting the default to `helloworld`
image

#### 2. Notice that all the urls are prefixed with `helloworld`, even though `WHALE_NETWORK` env variable is set to 'regtest' 
image
image

### DIFF
--- a/apps/whale-api/src/app.configuration.ts
+++ b/apps/whale-api/src/app.configuration.ts
@@ -29,11 +29,11 @@ export function AppConfiguration (): any {
 
 export function ENV_VALIDATION_SCHEMA (): any {
   return Joi.object({
-    NODE_ENV: Joi.string().valid('production'),
+    NODE_ENV: Joi.string().optional(),
     // Allows you to override whale endpoint version.
     WHALE_VERSION: Joi.string().default('v0.0'),
     WHALE_NETWORK: Joi.string().valid('mainnet', 'testnet', 'regtest', 'devnet').default('regtest'),
-    WHALE_DEFID_URL: Joi.string().required(),
+    WHALE_DEFID_URL: Joi.string().optional(),
     WHALE_DATABASE_PROVIDER: Joi.string().optional(),
     WHALE_DATABASE_LEVEL_LOCATION: Joi.string().optional()
   })
--- a/apps/whale-api/src/app.configuration.ts
+++ b/apps/whale-api/src/app.configuration.ts
@@ -30,8 +30,7 @@ export function AppConfiguration (): any {
 export function ENV_VALIDATION_SCHEMA (): any {
   return Joi.object({
     NODE_ENV: Joi.string().optional(),
-    // Allows you to override whale endpoint version.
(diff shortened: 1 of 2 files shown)


apps/whale-api/src/app.configuration.ts
```diff
@@ -24,3 +26,14 @@ export function AppConfiguration (): any {
...
+export function ENV_VALIDATION_SCHEMA (): any {
+  return Joi.object({
+    NODE_ENV: Joi.string().optional(),
+    WHALE_VERSION: Joi.string().optional(),
+    WHALE_NETWORK: Joi.string().valid('mainnet', 'testnet', 'regtest', 'devnet').default('regtest'),
+    WHALE_DEFID_URL: Joi.string().optional(),
+    WHALE_DATABASE_PROVIDER: Joi.string().optional(),
+    WHALE_DATABASE_LEVEL_LOCATION: Joi.string().optional()
+  })
+}
```
kodemon, line 39: I think we can ticket a refactor declaring this like the following:
```ts
export const ENV_VALIDATION_SCHEMA = {
  return Joi.object({
    NODE_ENV: Joi.string().optional(),
    WHALE_VERSION: Joi.string().optional(),
    WHALE_NETWORK: Joi.string().valid('mainnet', 'testnet', 'regtest', 'devnet').default('regtest'),
    WHALE_DEFID_URL: Joi.string().optional(),
    WHALE_DATABASE_PROVIDER: Joi.string().optional(),
    WHALE_DATABASE_LEVEL_LOCATION: Joi.string().optional()
  })
}
```
Since this is a constant that doesn't take any parameters I don't see the need to use this as a method.

Then your implementation in https://github.com/JellyfishSDK/jellyfish/blob/c12ba51d48f297a14900b2e26bcf641a97613776/apps/whale-api/src/app.module.ts#L38 becomes:

```ts
validationSchema: ENV_VALIDATION_SCHEMA
```

fuxingloh (approved):

kodemon (approved):

canonbrother:
ohh... Joi is to validate `process.env` but configuration :sweat_smile: 
thanks for the *FIX* :pray:

canonbrother (approved):

apps/whale-api/src/app.configuration.ts
fuxingloh: Hmm I wonder if there are any implications, maybe without the ability to do dynamic injection.

PR #1567, opened 2022-06-16
Fix PoolPair not found or wrong PoolPair (post split) issue on indexer by using id only

fuxingloh:


canonbrother (approved):
its pretty good for me for optimizing the getPoolPair by "tokenAId-TokenBId" in indexer :clap:

ivan-zynesis (approved):
Store one pair as 2 paths is much cleaner approach!

Issue #1583, opened 2022-06-27
Wrong encoding standard on `DCT_ID`, `AccountToUtxo.mintingOutputsStart` and `Vout.nTokenId`


fuxingloh:
### What happened:

On `DeFiCh/ain` cpp node client; these 2 lines are different.

```cpp
READWRITE(COMPACTSIZE(v));
READWRITE(VARINT(v));
```

On jellyfish, we expected `VARINT` and `COMPACTSIZE` to be the commonly used [varuint-bitcoin](https://github.com/bitcoinjs/varuint-bitcoin). However, that's not true it's a different standard, `COMPACTSIZE != VARINT`.

When the value is `>128`, the variable is encoded differently.

VARINT: 129 = `0x8001`, MSB Base-128, https://en.wikipedia.org/wiki/LEB128
COMPACTSIZE: 129: `0x81`, Variable length integer, https://en.bitcoin.it/wiki/Protocol_documentation#Variable_length_integer

### Need to be fixed:

Going through the `DeFiCh/ain` cpp node client, these 3 implementations need to be swapped to the new standard. **Currently on MainNet, no transactions will cause it to fail since this issue is isolated to when the integer is `>128`, none of the integers are `>128` yet or will be anytime soon.**

- [x] Add VarInt support in `@defichain/jellyfish-buffer`: https://github.com/JellyfishSDK/jellyfish/pull/1586
- [ ] `DCT_ID`: https://github.com/JellyfishSDK/jellyfish/pull/1588
- [ ] `AccountToUtxo.mintingOutputsStart`: https://github.com/JellyfishSDK/jellyfish/pull/1588
- [ ] `Vout.nTokenId`: https://github.com/JellyfishSDK/jellyfish/pull/1588

/triage accepted
/priority urgent-now
/apps whale-api
/area apps packages
/assign @fuxingloh @ivan-zynesis @canonbrother

PR #1586, opened 2022-06-27
feat(VarInt): add VarInt support using bn.js for bitwise operations

fuxingloh:

Add `VarInt` support in jellyfish using `bn.js` for bitwise operations. `bn.js` was used as default javascript doesn't support bitwise operations beyond 31 bits of Unsigned Integer due to Bitwise limitations of JS and performance issues. See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Bitwise_AND#description 

> The operands are converted to 32-bit integers and expressed by a series of bits (zeroes and ones). Numbers with more than 32 bits get their most significant bits discarded. For example, the following integer with more than 32 bits will be converted to a 32-bit integer:

Test vectors were directly taken from https://github.com/DeFiCh/ain/blob/a73f92ddef351750650a5eae46718d2f0ad2c7c5/src/test/serialize_tests.cpp#L178-L229, not all test where applicable hence not all were implemented.

Fixes part of #1583


ivan-zynesis (approved):

packages/jellyfish-buffer/__tests__/Composer.test.ts
```diff
@@ -1182,6 +1182,82 @@ describe('ComposableBuffer.varUInt', () => {
...
+      composer.toBuffer(buffer)
+
+      const expected = Buffer.from('8100', 'hex')
+      expect(buffer.toString()).toStrictEqual(expected.toString())
+    })
+  })
+
+  describe('2 byte = 16511', () => {
+    it('should fromBuffer', () => {
+      shouldFromBuffer(composer, 'ff7f', 16511, () => value)
```
canonbrother, line 1242: @fuxingloh  sorry i don't know how it works
eg:
```sh
echo $((0x7f)) # output: 127
echo $((0xff7f) # output: 65407
```
how does `ff7f` == 16511??

fuxingloh: So VarInt is a 128-bit operation, think of it as there is a bit operation every 128 bits, 128 in unsigned decimal or up to 0x7f in hex.

> Variable-length integers: bytes are a MSB base-128 encoding of the number. **The high bit in each byte signifies whether another digit follows.** To make sure the encoding is one-to-one, one is subtracted from all but the last digit. Thus, the byte sequence a[] with length len, where all but the last byte has bit 128 set, encodes the number:

So for simplification think:

1. `0` is `0x00`
2. `1` is `0x01`
3. ... and `127` is `0x7f`
4. and now `128` there are 2 "128 bits", hence presented in HEX it's `0x8000`, 
5. `129` is `0x8001`

When reading the VarInt buffer array [`0x01`], `0x01` is a low "byte"; <128. Hence there isn't another digit that follows. Thus just `1`.

When reading the VarInt buffer array [`0x80`, `0x01`] , `0x80` is a high "byte"; >128. Hence there is another digit that follows, read `0x01`, okay that's a low byte. `0x80 %  0x80 * 0x80 + 0x01` = 129...

Or in bitwise operations:

```ts
export function readVarInt (buffer: SmartBuffer): number {
  let n = 0

  while (true) {
    const buff: number = buffer.readUInt8()
    n = n << 7
    n += buff & 0x7f
    if ((buff & 0x80) !== 0) {
      n = n += 1
    } else {
      return n
    }
  }
}
```

// Breaking it down: 🥲
```ts
// init
let n = 0
const buffer = [0x8001]

// buff=0x80
let buff: number = buffer.readUInt8()

// shift left n by 7 bits
// n = 0 // still 0
n = n << 7

// n = 0 + (0x80 & 0x7f)
// n = 0 // still 0
n = n + (buff & 0x7f)

// (0x80 & 0x80) !== 0, true
// n = 0 + 1, n = 1
if ((buff & 0x80) !== 0) {
  n = n + 1
} else {
  return n
}

// buff=0x01
let buff: number = buffer.readUInt8()

// shift left n by 7 bits,  1 << 7 = 128
// n = 128
n = n << 7

// n = 128 + (0x01 & 0x7f)
// n = 129
n = n + (buff & 0x7f)

// (0x01 & 0x80) !== 0, false
// return 129
if ((buff & 0x80) !== 0) {
  n = n + 1
} else {
  return n
}
```

canonbrother: i won't resolve it.. this thread is useful.. just let it open

canonbrother (approved):

fuxingloh:
> ok, let's check if I understood the problem correct @fuxingloh So the Jellyfish had `CompactSize` implementation as `VarInt` which is incorrect. now we have renamed all `VarInt` usage as `CompactSize` in Jellyfish which still should change to the actual VarInt, But this does not matter for now since < 127 it's the same encoding.
> 
> But since cpp side use real `VarInt` for fields like `DCT_ID`, we should move Jellyfish side usage to `VarInt` for 100% accuracy. I guess this will happen in the subsequent issues logged.

Yup, you're spot on with one minor detail. Jellyfish has `COMPACTSIZE` for implementation for `VARINT` as it was thought to be the only variable integer size implementation. Since `VARINT` wasn't actually exposed outside of leveldb operations at the original bitcoin client. Jellyfish just assumed the same which is clearly wrong since we need both variable integer encoding implementations for the different data types.

Yeah, we will move to use `VARINT` for 100% accuracy, it's actually done with #1588 (together with tests for when ID is >128), just waiting for this to be merged since it stacks it on top of this current PR.

surangap (approved):

PR #1623, opened 2022-07-07
feat(apps/whale-api): surface "paybackburn" type as string[]

### LINKED ISSUE #1600: Surface other "paybackburn" information in `getburninfo` e.g. DUSD

For backward compatibility with existing releases, the "paybackburn" field in `/stats/burn` is now showing DFI value only. While the 2.9.x release included new information and fields that include DUSD burn information. Exposed using the same "paybackburn" field but as an array. We need to surface that information in `/stats/burn` too.

#### Why is this needed:

`DeFiCh/ain` `2.9.x` release causes issues due to zero-day breaking change on the `paybackburn` field. From being a `number` to a `string[]`. For more information on issues causes: https://github.com/JellyfishSDK/jellyfish/issues/1589

https://github.com/JellyfishSDK/jellyfish/pull/1580 was created to fix the issue by acting as an "anti-corruption layer". However, we will still be required to resurface that information with another field.

/triage accepted
/area apps
/apps whale-api
/priority important-soon

canonbrother:

Fixes #1600

### DIFF
--- a/apps/package.json
+++ b/apps/package.json
@@ -38,7 +38,7 @@
     "@nestjs/platform-express": "^8.4.6",
     "@nestjs/platform-fastify": "^8.4.6",
     "@nestjs/schedule": "^2.0.1",
-    "@nestjs/terminus": "^8.1.0",
+    "@nestjs/terminus": "^8.1.1",
     "@nestjs/typeorm": "^8.0.4",
     "async-mutex": "^0.3.2",
     "bignumber.js": "^9.0.2",
--- a/package-lock.json
+++ b/package-lock.json
@@ -60,7 +60,7 @@
         "@nestjs/platform-express": "^8.4.6",
         "@nestjs/platform-fastify": "^8.4.6",
         "@nestjs/schedule": "^2.0.1",
-        "@nestjs/terminus": "^8.1.0",
+        "@nestjs/terminus": "^8.1.1",
         "@nestjs/typeorm": "^8.0.4",
         "async-mutex": "^0.3.2",
         "bignumber.js": "^9.0.2",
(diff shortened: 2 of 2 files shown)


apps/legacy-api/__tests__/controllers/StatsController.test.ts
```diff
@@ -35,7 +35,7 @@ describe('StatsController', () => {
         ),
         feeburn: expect.any(Number),
         auctionburn: expect.any(Number),
-        paybackburn: expect.any(String),
+        paybackburn: expect.any(Number) ?? expect.any(Array),
```
canonbrother: lastly thinking to not impl a strict type on `paybackburn` to cater diff type of paybackburn from dev and ocean

apps/legacy-api/src/controllers/stats/LegacyStatsProvider.ts
```diff
@@ -229,7 +229,7 @@ interface LegacyBurnInfo {
   tokens: string[]
   feeburn: number
   auctionburn: number
-  paybackburn: string
+  paybackburn: number | string[]
```
canonbrother, line 232: https://github.com/JellyfishSDK/jellyfish/pull/1623/files#r916460149

packages/whale-api-client/src/api/stats.ts
```diff
@@ -53,7 +53,7 @@ export interface StatsData {
     address: number
     fee: number
     auction: number
-    payback: number
+    payback: string[]
```
fuxingloh:
```suggestion
    payback: number
    paybacks: string[]
```

Surfacing an additional field so that we don't break backward compatibility since downstream already relies on it being a `number`, if it suddenly becomes a `string[]` will be a breaking change and cause critical systems that rely on it being a `number` to break.

> Since it's a deployed endpoint (`/v0/`), we need to be able to apply rolling updates and not cause breaking changes for consumer.

canonbrother:
ohh.. damn!! 
it get closed after dropped the first commit

PR #1650, opened 2022-07-19
feat(apps/whale-api): add `/stats/rewards/distribution`

### LINKED ISSUE #1649: Expose a new endpoint `/stats/rewards/distribution` for block reward distribution

> Related to https://github.com/JellyfishSDK/jellyfish/issues/1485

https://github.com/JellyfishSDK/jellyfish/blob/6202c4da1c0880da83638968b757dc6af7833e32/apps/whale-api/src/module.api/stats.controller.ts#L279-L295

The above has been deprecated, without touching the existing endpoint; we should expose a new endpoint to get block reward distribution. Using the new functions from `package/jellyfish-network`. 

Let's call the new endpoint `/stats/rewards/distribution`?

https://github.com/JellyfishSDK/jellyfish/blob/0c33f3311dc37c078994dc720e7169ce09ceea2c/packages/jellyfish-network/src/BlockRewardDistribution.ts#L29

#### Why is this needed:

Although we have an SDK for it, there isn't an API for it yet.

/triage accepted
/area apps docs
/apps whale-api
/priority important-soon
/assign @canonbrother 🙏

canonbrother:

Fixes #1649

### DIFF
--- a/apps/whale-api/src/module.api/stats.controller.e2e.ts
+++ b/apps/whale-api/src/module.api/stats.controller.e2e.ts
@@ -1,6 +1,6 @@
 import { MasterNodeRegTestContainer } from '@defichain/testcontainers'
 import { StatsController } from './stats.controller'
-import { createTestingApp, stopTestingApp } from '../e2e.module'
+import { createTestingApp, stopTestingApp, waitForIndexedHeight } from '../e2e.module'
 import { NestFastifyApplication } from '@nestjs/platform-fastify'
 
 const container = new MasterNodeRegTestContainer()
@@ -12,6 +12,8 @@ beforeAll(async () => {
   await container.waitForWalletCoinbaseMaturity()
 
   app = await createTestingApp(container)
+  await waitForIndexedHeight(app, 100)
+
   controller = app.get<StatsController>(StatsController)
 })
 
@@ -20,14 +22,17 @@ afterAll(async () => {
 })
 
 it('should getRewardDistribution', async () => {
+  await container.generate(10)
+  await waitForIndexedHeight(app, 110)
+
   const data = await controller.getRewardDistribution()
   expect(data).toStrictEqual({
-    masternode: expect.any(Number),
-    community: expect.any(Number),
-    anchor: expect.any(Number),
-    liquidity: expect.any(Number),
-    loan: expect.any(Number),
-    options: expect.any(Number),
-    unallocated: expect.any(Number)
+    masternode: 66.66,
+    community: 9.82,
+    anchor: 0.04,
+    liquidity: 50.9,
+    loan: 49.36,
+    options: 19.76,
+    unallocated: 3.46
   })
 })
--- a/apps/whale-api/src/module.api/stats.controller.e2e.ts
+++ b/apps/whale-api/src/module.api/stats.controller.e2e.ts
@@ -1,0 +1,33 @@
+import { MasterNodeRegTestContainer } from '@defichain/testcontainers'
+import { StatsController } from './stats.controller'
+import { createTestingApp, stopTestingApp } from '../e2e.module'
+import { NestFastifyApplication } from '@nestjs/platform-fastify'
+
+const container = new MasterNodeRegTestContainer()
+let app: NestFastifyApplication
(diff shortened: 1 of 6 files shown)


apps/whale-api/src/module.api/stats.controller.ts
```diff
@@ -96,6 +104,20 @@ export class StatsController {
     }
   }
 
+  @Get('/rewards/distribution')
+  async getRewardDistribution (
+    @Query('height', ParseIntPipe) reductionHeight = 0
+  ): Promise<RewardDistributionData> {
+    const options: CoinbaseSubsidyOptions = this.network === 'mainnet'
+      ? MainNetCoinbaseSubsidyOptions
+      : TestNetCoinbaseSubsidyOptions
```
eli-lim: I think we can set `options` once in the constructor so that we don't have to do repeat it for every request

apps/whale-api/src/module.api/stats.controller.ts
```diff
@@ -96,6 +104,20 @@ export class StatsController {
     }
   }
 
+  @Get('/rewards/distribution')
```
eli-lim, line 104: should add a test for this new endpoint in `stats.controller.spec.ts`?

canonbrother: hmm.. the `spec` is for unit testing.. so yea i created another `e2e` for it  :grimacing:

eli-lim: hmm okay actually that's fine then - e2e should cover it

apps/whale-api/src/module.api/stats.controller.ts
```diff
@@ -9,21 +9,34 @@ import { PriceTickerMapper } from '../module.model/price.ticker'
...
+  MainNetCoinbaseSubsidyOptions,
+  TestNetCoinbaseSubsidyOptions,
+  getBlockRewardDistribution
+} from '@defichain/jellyfish-network'
 import { BurnInfo } from '@defichain/jellyfish-api-core/dist/category/account'
 import { GetLoanInfoResult } from '@defichain/jellyfish-api-core/dist/category/loan'
 
 @Controller('/stats')
 export class StatsController {
+  options: CoinbaseSubsidyOptions
```
eli-lim:
```suggestion
  private readonly options: CoinbaseSubsidyOptions
```

one last `nit` 😛

apps/whale-api/src/module.api/stats.controller.ts
```diff
@@ -96,6 +109,16 @@ export class StatsController {
     }
   }
 
+  @Get('/rewards/distribution')
+  async getRewardDistribution (
+    @Query('height', ParseIntPipe) reductionHeight = 0
+  ): Promise<RewardDistributionData> {
+    const height = this.options.eunosHeight + reductionHeight * this.options.emissionReductionInterval
+    const subsidy = this.blockSubsidy.getBlockSubsidy(height)
```
fuxingloh: Since /stats/ is live data, we pass in the latest height from BlockMapper instead of computing the reduction intervals.

```suggestion
  async getRewardDistribution (): Promise<RewardDistributionData> {
    const block = requireValue(await this.blockMapper.getHighest(), 'block')
    const subsidy = this.blockSubsidy.getBlockSubsidy(block.height)
```

apps/whale-api/src/module.api/stats.controller.e2e.ts
```diff
@@ -0,0 +1,33 @@
...
+it('should getRewardDistribution', async () => {
+  const data = await controller.getRewardDistribution()
+  expect(data).toStrictEqual({
+    masternode: expect.any(Number),
+    community: expect.any(Number),
+    anchor: expect.any(Number),
+    liquidity: expect.any(Number),
+    loan: expect.any(Number),
+    options: expect.any(Number),
+    unallocated: expect.any(Number)
```
fuxingloh: We should test the static numbers for these, just incase we screwed up the number precision

fuxingloh (approved):

eli-lim (approved):

PR #1710, opened 2022-08-17
fix(apps/whale-api): Fix `listAccountHistory` json out of range

### LINKED ISSUE #1701: RpcApiError: JSON integer out of range
#### What happened:

While playing around with the WhaleApiClient, I've tried to list all the account history with `client.address.listAccountHistory` for one of my addresses. Since the history has a lot of entries, I used pagination via 'hasNext' and and the 'nextToken`. The first iterations went through well, but at some point, the WhaleApi returns a 400 Bad Request, indicating that I messed up.

```json
{"error": {
  "code": 400,
  "type": "BadRequest",
  "at": 1660373481810,
  "message": "RpcApiError: 'JSON integer out of range', code: -1, method: listaccounthistory","url":"/v0/mainnet/address/dKKSxM4iANcRupaQ6z2FsUs1kR8CTwErSF/history?size=50&next=7b0a3bb2a39aba333741f4319648213f4e3e155a497f0467523bab5f7e19952d-PoolSwap-546412"
}}
```

I expected to iterate through all pages without an error, finishing when there are no more results.

```ts
async function getFullAccountHistory(
  address: string
): Promise<AddressHistory[]> {
  const history: AddressHistory[] = [];
  let historyResponse = await client.address.listAccountHistory(address, 50);
  history.push(...historyResponse);
  while (historyResponse.hasNext) {
    historyResponse = await client.address.listAccountHistory(
      address,
      50,
      historyResponse.nextToken
    );
    history.push(...historyResponse);
  }
  return history;
}

getFullAccountHistory("dKKSxM4iANcRupaQ6z2FsUs1kR8CTwErSF")
```

canonbrother:

As per title.

Fixes #1701

#### Additional comments?:
The bug is triggerred while in `full.length % limit !== 0` scenario as it will always `not equal to` and `smaller than` limit.
> if `sliced.length !== limit => loop` here, it will be infinitely increasing `limit *= 2` which causes **overflow** error

https://github.com/DeFiCh/ain/blob/277b848c9fa46b9459f51cf97f15f5d891c1b7e0/src/masternodes/rpc_accounts.cpp#L312
https://github.com/DeFiCh/ain/blob/277b848c9fa46b9459f51cf97f15f5d891c1b7e0/src/univalue/lib/univalue_get.cpp#L119

To ensure **next1** is the last batch by verifying through `list.length` vs `limit`
`list.length !== limit` - isLastBatch

| action | limit | max | output (final output if sliced is n/a)              | sliced (final output) |
|--------|-------|-----|----------------------------------------------------|----------------------|
|        | 10    |     | [A50, B50, C50, D50, E50, F50, G50, H50, I15, J10] | n/a                  |
| next1  | 10    | 10  | [J10, K05, L01]                                    | [K05, L01]           |

### DIFF
--- a/apps/whale-api/src/module.api/address.controller.ts
+++ b/apps/whale-api/src/module.api/address.controller.ts
@@ -69,7 +69,24 @@ export class AddressController {
 
     if (next !== undefined) {
       const [txid, txType, maxBlockHeight] = next.split('-')
-
+      /**
+       *  Demostrate workaround the `loop`
+       *  full: [A50, B50, C50, D50, E50, F50, G50, H50, I15, J10, K05, L01]
+       *
+       *  | action | limit | max | output                         | sliced          |
+       *  |--------|-------|-----|--------------------------------|-----------------|
+       *  |        | 3     |     | [A50, B50, C50]                |                 |
+       *  | next1  | 3     | 50  | [A50, B50, C50]                | [ ]             |
+       *  | loop1  | 6     | 50  | [A50, B50, C50, D50, E50, F50] | [D50, E50, F50] |
+       *  | next2  | 3     | 50  | [A50, B50, C50]                | [ ]             |
+       *  | loop2  | 6     | 50  | [A50, B50, C50, D50, E50, F50] | [ ]             |
+       *  | loop3  | 12    | 50  | [A50, B50, C50, D50, E50, F50, | [G50, H50, I15] |
+       *  |        |       |     |  G50, H50, I15, J10, K05, L01] |                 |
+       *  | next3  | 3     | 15  | [I15, J10, K05]                | [J10, K05]      |
+       *  | loop4  | 6     | 15  | [I15, J10, K05, L01]           | [J10, K05, L01] |
+       *
+       *  Found **last batch** on loop4 as its `list.length !== limit`.
+       */
       const loop = async (maxBlockHeight: number, limit: number): Promise<AccountHistory[]> => {
         const list = await this.rpcClient.account.listAccountHistory(address, {
           limit: limit,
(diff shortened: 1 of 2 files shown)


apps/whale-api/src/module.api/address.controller.ts
```diff
@@ -87,7 +87,7 @@ export class AddressController {
         const start = foundIndex + 1 // plus 1 to exclude the prev txid
         const size = start + query.size
         const sliced = list.slice(start, size)
-        if (sliced.length !== query.size) {
+        if (sliced.length === 0) {
```
fuxingloh: I'm a little confused, you mind explaining how it fixes it?

canonbrother: yea sure.. should document it

canonbrother: phew.. updated!!

canonbrother:
revert back to draft as found current handling has conflict with
https://github.com/JellyfishSDK/jellyfish/blob/660df936ac08dc6a9c4099648aba47c7b8e9bf6f/apps/whale-api/src/module.api/_core/api.paged.response.ts#L89

apps/whale-api/src/module.api/address.controller.ts
```diff
@@ -87,7 +105,7 @@ export class AddressController {
         const start = foundIndex + 1 // plus 1 to exclude the prev txid
         const size = start + query.size
         const sliced = list.slice(start, size)
-        if (sliced.length !== query.size) {
+        if (sliced.length !== query.size && !isLastBatch) {
```
canonbrother, line 108: `sliced.length !== query.size` is required as https://github.com/JellyfishSDK/jellyfish/pull/1710#issuecomment-1220453064

added `!isLastBatch` checker to stop looping

(took some time on this.. lack of **competitive programming** practice :sweat_smile: )

apps/whale-api/src/module.api/address.controller.ts
canonbrother: resolved and refer to https://github.com/JellyfishSDK/jellyfish/pull/1710#discussion_r950346813

fuxingloh (approved):
LGTM!

PR #1723, opened 2022-08-26
fix(jellyfish-api-core, whale-api-client) Fix test failures against ain v2.10.0

Jouzo:

Fixes test failures against ain v2.10
Still WIP, `getvault` verbose needs to be fixed. A type regression was introduced in 2.9.3 in PR https://github.com/defiCh/ain/pull/1405


fuxingloh:
Superseded with https://github.com/JellyfishSDK/jellyfish/pull/1722, the fix for BigNumber here won't be compatible with the client downstream and will cause a regression regardless. As for error code drift, it should be updated and not commented out.

PR #1740, opened 2022-09-06
feat(whale-api): add name on poolpairs api response

lykalabrada:
On the redesigned screens for Token Selection (swap), we need to display the token name. In this PR, token name is added to the response of poolpairs apis.

Fixes #

#### Additional comments?:

- Affected endpoints: `/poolpairs`, `/poolpairs/paths/swappable/13`, `/poolpairs/dexprices`. There can be more affected routes, so definitely need the reviewers' help to check. 
- For `/poolpairs` api, `name` is added inside `tokenA` and `tokenB` data.

- Redesigned screen where we need to display the name:
image

### DIFF
--- a/apps/whale-api/src/module.api/poolpair.controller.ts
+++ b/apps/whale-api/src/module.api/poolpair.controller.ts
@@ -15,6 +15,7 @@ import { PoolPairService } from './poolpair.service'
 import { PoolSwapPathFindingService } from './poolswap.pathfinding.service'
 import BigNumber from 'bignumber.js'
 import { PoolPairInfo } from '@defichain/jellyfish-api-core/dist/category/poolpair'
+import { TokenInfo } from '@defichain/jellyfish-api-core/dist/category/token'
 import { parseDATSymbol } from './token.controller'
 import { PoolSwapMapper } from '../module.model/pool.swap'
 import { PoolSwapAggregatedMapper } from '../module.model/pool.swap.aggregated'
@@ -55,6 +56,8 @@ export class PoolPairController {
       const totalLiquidityUsd = await this.poolPairService.getTotalLiquidityUsd(info)
       const apr = await this.poolPairService.getAPR(id, info)
       const volume = await this.poolPairService.getUSDVolume(id)
+      info.nameTokenA = (await this.deFiDCache.getTokenInfo(info.idTokenA) as TokenInfo).name
+      info.nameTokenB = (await this.deFiDCache.getTokenInfo(info.idTokenB) as TokenInfo).name
       items.push(mapPoolPair(id, info, totalLiquidityUsd, apr, volume))
     }
 
@@ -81,6 +84,8 @@ export class PoolPairController {
     const totalLiquidityUsd = await this.poolPairService.getTotalLiquidityUsd(info)
     const apr = await this.poolPairService.getAPR(id, info)
     const volume = await this.poolPairService.getUSDVolume(id)
+    info.nameTokenA = (await this.deFiDCache.getTokenInfo(info.idTokenA) as TokenInfo).name
+    info.nameTokenB = (await this.deFiDCache.getTokenInfo(info.idTokenB) as TokenInfo).name
     return mapPoolPair(String(id), info, totalLiquidityUsd, apr, volume)
   }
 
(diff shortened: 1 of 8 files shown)


apps/whale-api/src/module.api/poolpair.controller.ts
```diff
@@ -203,6 +203,7 @@ function mapPoolPair (
   volume?: PoolPairData['volume']
 ): PoolPairData {
   const [symbolA, symbolB] = info.symbol.split('-')
+  const [nameA, nameB] = info.name.split('-')
```
fuxingloh: For PoolPair, this isn't a safe way to get `name` since the name of the underlying token can be updated without updating the poolpair. The `poolpair.name` can also be created without `-`, making this code fatal. 

I would recommend not implementing `name` for this code section if you don't need it.

fuxingloh (changes requested):

apps/whale-api/src/module.api/poolpair.controller.ts
lykalabrada: Oh I see, let me check. But I might need your help with this 😅

lykalabrada: Can I ask @fuxingloh, when will there be no `-`? I thought `name` on this part of the code is always a combination of tokenA and tokenB's names

fuxingloh: > Can I ask @fuxingloh, when will there be no `-`? I thought `name` on this part of the code is always a combination of tokenA and tokenB's names

You can update a PoolPair token to any name after the creation of the token https://jellyfishsdk.com/node/jellyfish/api/token#updatetoken

fuxingloh: If you need the name, you can use `DeFiDCache` and use the token id to resolve the name instead. Since it's a cache the complexity is managed and won't be a performance issue.

https://github.com/JellyfishSDK/jellyfish/blob/0200dc74d77f6d331e531cad0db6e991f477f26a/apps/whale-api/src/module.api/poolpair.controller.ts#L28

lykalabrada: Ok I see. I will try to use DeFiDCache. Thank you!

lykalabrada: Updated. Let me know if I can improve anything. Thanks! 🙂

packages/jellyfish-api-core/src/category/poolpair.ts
```diff
@@ -201,6 +201,8 @@ export interface PoolPairInfo {
   status: boolean
   idTokenA: string
   idTokenB: string
+  nameTokenA: string
+  nameTokenB: string
```
fuxingloh:
```suggestion
```

this file shouldn't be edited since it's internal

apps/whale-api/src/module.api/poolpair.controller.ts
```diff
@@ -81,6 +84,8 @@ export class PoolPairController {
     const totalLiquidityUsd = await this.poolPairService.getTotalLiquidityUsd(info)
     const apr = await this.poolPairService.getAPR(id, info)
     const volume = await this.poolPairService.getUSDVolume(id)
+    info.nameTokenA = (await this.deFiDCache.getTokenInfo(info.idTokenA) as TokenInfo).name
+    info.nameTokenB = (await this.deFiDCache.getTokenInfo(info.idTokenB) as TokenInfo).name
```
fuxingloh: How about moving `mapPoolPair` function into the class function of `PoolPairController` so that you don't need to duplicate this code and instead have it embedded into the function to keep it DRY.

fuxingloh (changes requested):

packages/jellyfish-api-core/src/category/poolpair.ts
lykalabrada: You're right, I don't even have to change this. Removed now.

apps/whale-api/src/module.api/poolpair.controller.ts
lykalabrada: Thanks for the suggestion! Updated.

fuxingloh (approved):
LGTM

fuxingloh:
@canonbrother @eli-lim can y'all do a quick review for this

canonbrother (approved):

kodemon (approved):

PR #1758, opened 2022-09-15
feat(status-api): Implement oracle status for number of active oracles

### LINKED ISSUE #1748: `status-api` Implement another monitoring api to track oracle statuses
Implement another monitoring api within `status-api` for oracle status on top of existing `getOracleStatus` to track the number of available oracle provider per ticker e.g. 4 available out of 6 total 

#### Why is this needed:
This will provide more accurate alerting for us to action on, i.e. if >30% of the oracle providers of each ticker is down.

nichellekoh:
To add another oracle status api that returns the status of each ticker e.g. DFI/USD instead of the existing per oracle address. This increases the meaningfulness of each status return to create better monitoring for our oracle health.

Fixes https://github.com/JellyfishSDK/jellyfish/issues/1748

#### Additional comments?:

cc: @joeldavidw

### DIFF
--- a/apps/status-api/src/controllers/OracleStatusController.ts
+++ b/apps/status-api/src/controllers/OracleStatusController.ts
@@ -27,6 +27,29 @@ export class OracleStatusController {
     }, 5) // cache status result for 5 seconds
   }
 
+  /**
+   * To provide the status of each ticker determined by the number of oracles responded given the ticker id
+   *
+   * @return {Promise<OracleStatus>}
+   * @param token
+   * @param currency
+   */
+  @Get('/:token/:currency')
+  async getOracleRespondStatus (@Param('token') token: string, @Param('currency') currency: string): Promise<{ status: OracleStatus }> {
+    return await this.cachedGet(`oracle-${token}-${currency}`, async () => {
+      const oracles = await this.client.prices.getOracles(token, currency, 60)
+      const total = oracles.filter(oracle => oracle.feed !== undefined).length
+
+      const prices = await this.client.prices.get(token, currency)
+      const active = prices.price.aggregated.oracles.active
+
+      if (active > 3 || (active / total) >= 0.75) {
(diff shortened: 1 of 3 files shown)


apps/status-api/src/controllers/OracleStatusController.ts
```diff
@@ -27,6 +27,29 @@ export class OracleStatusController {
...
+  @Get('/:token/:currency')
+  async getOracleRespondStatus (@Param('token') token: string, @Param('currency') currency: string): Promise<{ status: OracleStatus }> {
+    return await this.cachedGet(`oracle-${token}-${currency}`, async () => {
+      const oracles = await this.client.prices.getOracles(token, currency, 60)
+      const total = oracles.filter(oracle => oracle.feed !== undefined).length
+
+      const prices = await this.client.prices.get(token, currency)
+      const active = prices.price.aggregated.oracles.active
+
+      if (active > 3 || (active / total) >= 0.75) {
```
nichellekoh: Do we want it as operational for which of the following condition?
1. there are 3 active OR at least 75% active
2. there are 3 active AND at least 75% active

joeldavidw: There are 3 active AND at least 75% active

nichellekoh: The 75% condition will not be a good gauge for tickers that recently went through stock-split as the total oracle number will include the pre-split ones. Thinking of an alternative condition now

nichellekoh: Conclusion: 
If there are > 3 total oracles, alert if the active oracles <= 3.
If there are = 3 total oracles, alert if the active oracles < 3.  (outlier case)

We will remove the 75% condition

joeldavidw (approved):

PR #1784, opened 2022-09-26
feat(jellyfish-api-core): add blockchain getMempoolAncestors RPC

### LINKED ISSUE #48: DeFiChain Jellyfish RPC Implementations

delphk:
/kind feature

Fixes part of #48 
- Implements `getmempoolancestors` type for RPC.

### DIFF
--- a/packages/jellyfish-api-core/__tests__/category/blockchain/getMempoolAncestors.test.ts
+++ b/packages/jellyfish-api-core/__tests__/category/blockchain/getMempoolAncestors.test.ts
@@ -1,0 +1,93 @@
+import { MasterNodeRegTestContainer } from '@defichain/testcontainers'
+import { Testing } from '@defichain/jellyfish-testing'
+import BigNumber from 'bignumber.js'
+
+describe('getMempoolAncestors', () => {
+  const container = new MasterNodeRegTestContainer()
+  const testing = Testing.create(container)
+  const address = 'mwsZw8nF7pKxWH8eoKL9tPxTpaFkz7QeLU'
+  const amount = 0.003
+
+  beforeAll(async () => {
+    await testing.container.start()
+    await testing.container.waitForWalletCoinbaseMaturity()
+  })
+
+  afterAll(async () => {
+    await testing.container.stop()
+  })
+
+  it('should return empty array for transaction id without ancestors', async () => {
+    const txId = await testing.rpc.wallet.sendToAddress(address, amount)
+    const mempoolEntry = await testing.rpc.blockchain.getMempoolAncestors(txId)
+
+    expect(mempoolEntry.length).toStrictEqual(0)
+  })
+
+  it('should return error if transaction is not in mempool', async () => {
+    const txId = '9fb9c46b1d12dae8a4a35558f7ef4b047df3b444b1ead61d334e4f187f5f58b7'
+    await expect(testing.rpc.blockchain.getMempoolAncestors(txId))
+      .rejects
+      .toThrowError('RpcApiError: \'Transaction not in mempool\', code: -5, method: getmempoolancestors')
+  })
+
+  it('should return error if transaction id has an invalid length', async () => {
+    await expect(testing.rpc.blockchain.getMempoolAncestors('invalidtxidstring'))
+      .rejects
+      .toThrow('RpcApiError: \'parameter 1 must be of length 64 (not 17, for \'invalidtxidstring\')\', code: -8, method: getmempoolancestors')
+  })
+
+  describe('transactions with ancestors', () => {
+    beforeAll(async () => {
+      for (let i = 0; i < 3; i++) {
+        await testing.rpc.wallet.sendToAddress(address, amount)
+      }
+    })
+
+    it('should return JSON object if verbose is true', async () => {
+      const txId = await testing.rpc.wallet.sendToAddress(address, amount)
+      const mempoolAncestors = await testing.rpc.blockchain.getMempoolAncestors(txId, true)
+      const keys = Object.keys(mempoolAncestors)
+      expect(keys.length).toBeGreaterThan(0)
+      for (const key of keys) {
+        expect(mempoolAncestors[key]).toStrictEqual({
+          fees: expect.any(Object),
+          vsize: expect.any(BigNumber),
+          weight: expect.any(BigNumber),
+          fee: expect.any(BigNumber),
+          modifiedfee: expect.any(BigNumber),
+          time: expect.any(BigNumber),
+          height: expect.any(BigNumber),
+          descendantcount: expect.any(BigNumber),
+          descendantsize: expect.any(BigNumber),
+          descendantfees: expect.any(BigNumber),
+          ancestorcount: expect.any(BigNumber),
+          ancestorsize: expect.any(BigNumber),
+          ancestorfees: expect.any(BigNumber),
+          wtxid: expect.any(String),
+          depends: expect.any(Array),
+          spentby: expect.any(Array),
+          'bip125-replaceable': expect.any(Boolean)
+        })
+      }
+    })
+
+    it('should return array of transaction ids if verbose is false', async () => {
+      const txId = await testing.rpc.wallet.sendToAddress(address, amount)
+      const mempoolAncestors = await testing.rpc.blockchain.getMempoolAncestors(txId, false)
+      expect(mempoolAncestors.length).toBeGreaterThan(0)
+      for (const ancestorId of mempoolAncestors) {
+        expect(ancestorId).toStrictEqual(expect.stringMatching(/^[0-f]{64}$/))
+      }
+    })
+
+    it('should return array of transaction ids if verbose is undefined', async () => {
+      const txId = await testing.rpc.wallet.sendToAddress(address, amount)
+      const mempoolAncestors = await testing.rpc.blockchain.getMempoolAncestors(txId)
+      expect(mempoolAncestors.length).toBeGreaterThan(0)
+      for (const ancestorId of mempoolAncestors) {
+        expect(ancestorId).toStrictEqual(expect.stringMatching(/^[0-f]{64}$/))
+      }
+    })
+  })
+})
--- a/packages/jellyfish-api-core/__tests__/category/blockchain/getMempoolAncestors.test.ts
+++ b/packages/jellyfish-api-core/__tests__/category/blockchain/getMempoolAncestors.test.ts
@@ -2,11 +2,9 @@ import { MasterNodeRegTestContainer } from '@defichain/testcontainers'
 import { Testing } from '@defichain/jellyfish-testing'
 import BigNumber from 'bignumber.js'
 
-describe('getMempoolAncestors', () => {
+describe('transactions without ancestors', () => {
   const container = new MasterNodeRegTestContainer()
   const testing = Testing.create(container)
-  const address = 'mwsZw8nF7pKxWH8eoKL9tPxTpaFkz7QeLU'
-  const amount = 0.003
 
   beforeAll(async () => {
     await testing.container.start()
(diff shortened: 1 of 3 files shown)


packages/jellyfish-api-core/__tests__/category/blockchain/getMempoolAncestors.test.ts
```diff
@@ -0,0 +1,93 @@
...
+        })
+      }
+    })
+
+    it('should return array of transaction ids if verbose is false', async () => {
+      const txId = await testing.rpc.wallet.sendToAddress(address, amount)
+      const mempoolAncestors = await testing.rpc.blockchain.getMempoolAncestors(txId, false)
+      expect(mempoolAncestors.length).toBeGreaterThan(0)
+      for (const ancestorId of mempoolAncestors) {
+        expect(ancestorId).toStrictEqual(expect.stringMatching(/^[0-f]{64}$/))
```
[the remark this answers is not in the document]
fuxingloh:
```suggestion
        expect(ancestorId).toStrictEqual(expect.stringMatching(/^[0-9a-f]{64}$/))
```

Given the CodeQL warning, we should stricten it.

packages/jellyfish-api-core/__tests__/category/blockchain/getMempoolAncestors.test.ts
```diff
@@ -0,0 +1,93 @@
...
+      }
+    })
+
+    it('should return array of transaction ids if verbose is undefined', async () => {
+      const txId = await testing.rpc.wallet.sendToAddress(address, amount)
+      const mempoolAncestors = await testing.rpc.blockchain.getMempoolAncestors(txId)
+      expect(mempoolAncestors.length).toBeGreaterThan(0)
+      for (const ancestorId of mempoolAncestors) {
+        expect(ancestorId).toStrictEqual(expect.stringMatching(/^[0-f]{64}$/))
```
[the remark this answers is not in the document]
fuxingloh:
```suggestion
        expect(ancestorId).toStrictEqual(expect.stringMatching(/^[0-9a-f]{64}$/))
```

fuxingloh (changes requested):
This is not a full review, I will do one later.

packages/jellyfish-api-core/src/category/blockchain.ts
```diff
@@ -164,6 +164,36 @@ export class Blockchain {
...
+  /**
+   * Get all in-mempool ancestors if txId is in mempool as string[] if verbose is false
+   * else as json object
+   *
+   * @param {string} txId the transaction id
+   * @param {boolean} verbose default = false, true for json object, false for array of transaction ids
+   * @return {Promise<string[] | MempoolTx>}
+   */
+  async getMempoolAncestors (txId: string, verbose?: boolean): Promise<string[] | MempoolTx> {
+    return await this.client.call('getmempoolancestors', [txId, verbose], 'bignumber')
```
fuxingloh, line 194: A little history about this design, back when we designed the JSON-RPC interfaces, we wanted to make the object field DX-friendly. Hence, the invention allowed a declaration to declare where the BigNumber or numbers are. The idea was to be "ACCOMMODATING", and make them a number when the value isn't sensitive to precision, lost such as "maxFees" or height.

Although the intention was sound, we have some regrets now as it is unnecessarily complicated to declare new fields. Almost impossible when dealing with unknown key-value objects (this RPC, `MempoolTx`). Maybe it is better just to make them all BigNumber and call it a day.

- https://github.com/JellyfishSDK/jellyfish/issues/18
- https://github.com/JellyfishSDK/jellyfish/pull/97

packages/jellyfish-api-core/__tests__/category/blockchain/getMempoolAncestors.test.ts
```diff
@@ -0,0 +1,93 @@
+import { MasterNodeRegTestContainer } from '@defichain/testcontainers'
+import { Testing } from '@defichain/jellyfish-testing'
+import BigNumber from 'bignumber.js'
+
+describe('getMempoolAncestors', () => {
```
fuxingloh: For tests (`it('')` and `describe('')`), they should not be inter-dependent but self-sufficient.

Given that we use the same Container for many of the tests in the same file, it becomes challenging to write clean tests that are non-inter-dependent. 

And why do we use the same container — well simplify for test velocity and speed. Since spinning up multiple containers compared to one is much slower. 

You should break the `describe` and `it` into independent units that can be run on its own without relying on a sequential order. The `describe` block must describe a container with the condition that is scoped to a scenario. The scenario must not be created by a `it` condition (those are called side-effects).

> I think there are plentiful literature of how tests are written in this repo, but it's very difficult to find them for your reference since they're buried in PR comments.

packages/jellyfish-api-core/__tests__/category/blockchain/getMempoolAncestors.test.ts
```diff
@@ -0,0 +1,93 @@
...
+    await testing.container.start()
+    await testing.container.waitForWalletCoinbaseMaturity()
+  })
+
+  afterAll(async () => {
+    await testing.container.stop()
+  })
+
+  it('should return empty array for transaction id without ancestors', async () => {
+    const txId = await testing.rpc.wallet.sendToAddress(address, amount)
```
fuxingloh:
```suggestion
    const txId = await testing.rpc.wallet.sendToAddress('mwsZw8nF7pKxWH8eoKL9tPxTpaFkz7QeLU', 0.003)
```

We should also take a different approach to writing theses. E.g. things like making them abstract or DRY might inherently make tests harder to read and digest.

It might be better to make them static so the reader of the test can follow the test without cross-referencing fields from top and bottom of the file

fuxingloh (changes requested):

packages/jellyfish-api-core/src/category/blockchain.ts
delphk: left the precision as 'bignumber' (same as `getMempoolEntry`) for now

fuxingloh: Yeah, we have to for KeyVault objects.

packages/jellyfish-api-core/__tests__/category/blockchain/getMempoolAncestors.test.ts
```diff
@@ -0,0 +1,100 @@
...
+    expect(mempoolAncestors.length).toBeGreaterThan(0)
+    for (const ancestorId of mempoolAncestors) {
+      expect(ancestorId).toStrictEqual(expect.stringMatching(/^[0-9a-f]{64}$/))
+    }
+  })
+
+  it('should return array of transaction ids if verbose is undefined', async () => {
+    const txId = await testing.rpc.wallet.sendToAddress('mwsZw8nF7pKxWH8eoKL9tPxTpaFkz7QeLU', 0.003)
+    const mempoolAncestors = await testing.rpc.blockchain.getMempoolAncestors(txId)
+    expect(mempoolAncestors.length).toBeGreaterThan(0)
```
weiyuan95, line 95: IMO testing greater than 0 here is pretty scary since we're not expecting any behaviour. Ie. Since we sent 3 transactions to chain, we should eventually see 3 ancestors here. As it is currently, even if there is just 1 ancestor, the test will pass.

We can clear the mempool by generating a block after each test so that the tests are always run with a clean slate.

weiyuan95 (changes requested):

packages/jellyfish-api-core/src/category/blockchain.ts
```diff
@@ -164,6 +164,36 @@ export class Blockchain {
...
+
+  /**
+   * Get all in-mempool ancestors if txId is in mempool as string[] if verbose is false
+   * else as json object
+   *
+   * @param {string} txId the transaction id
+   * @param {boolean} verbose default = false, true for json object, false for array of transaction ids
+   * @return {Promise<string[] | MempoolTx>}
+   */
+  async getMempoolAncestors (txId: string, verbose?: boolean): Promise<string[] | MempoolTx> {
```
eli-lim: Let's use the default value `false` instead of sending `undefined`

```suggestion
  async getMempoolAncestors (txId: string, verbose: boolean = false): Promise<string[] | MempoolTx> {
```

packages/jellyfish-api-core/__tests__/category/blockchain/getMempoolAncestors.test.ts
```diff
@@ -0,0 +1,100 @@
...
+  afterAll(async () => {
+    await testing.container.stop()
+  })
+
+  it('should return JSON object if verbose is true', async () => {
+    const txId = await testing.rpc.wallet.sendToAddress('mwsZw8nF7pKxWH8eoKL9tPxTpaFkz7QeLU', 0.003)
+    const mempoolAncestors = await testing.rpc.blockchain.getMempoolAncestors(txId, true)
+    const keys = Object.keys(mempoolAncestors)
+    expect(keys.length).toBeGreaterThan(0)
+    for (const key of keys) {
```
eli-lim, line 60: Would prefer to have some stricter check here (e.g. asserting the exact count of the mempool ancestors for the given tx)

eli-lim (commented):
Looks good generally, but will need to make the tests stricter as jellyfish can act as a second layer to prevent regressions

packages/jellyfish-api-core/__tests__/category/blockchain/getMempoolAncestors.test.ts
delphk: Correct me if my understanding is wrong, but as the method is getting the ancestors of a particular transaction ID, the number of ancestors of that particular transaction might not be the same as the number of transactions in the mempool
(referencing the diagram [here](https://bitcoin.stackexchange.com/questions/99168/clarifying-descendants-ancestors-in-error-cases-for-too-long-mempool-chain))
In that case, it would be difficult to assert exactly how many ancestors a transaction has (except that it is greater than 0). Also ran the tests a number of times with logs and I could not get an ancestor length of 3

packages/jellyfish-api-core/__tests__/category/blockchain/getMempoolAncestors.test.ts
delphk: see my comment above, it would be tricky to assert the exact count of mempool ancestors for a particular transaction (unless there is only one other transaction in the mempool)

packages/jellyfish-api-core/__tests__/category/blockchain/getMempoolAncestors.test.ts
eli-lim: MasternodeRegtestContainer gives us a local private defichain network to test against
hence the number of transactions, when a block is mined, etc. are all deterministic

eli-lim: This is my understanding:

Say Alice initiates 3 transactions to Bob and Bob initiates 1 transaction to Cat,
the tx hierarchy would look like:

```mermaid
graph LR
  A(Alice) --"0.5 (1)"--> B(Bob)
  A --"0.5 (2)"--> B
  A --"0.5 (3)"--> B
  B --"1.5 (4)"--> C(Cat)
```

Prior to the block being mined, `(4)` would have ancestors `(1) (2) (3)`

This will mean that the test needs to set up 2 wallets - 1 representing Alice and another representing Bob
We can probably use the one provided by `testing.rpc.wallet` to represent Alice, but probably need to create another to represent Bob (though I haven't looked through how)

delphk: based on my observation of the mempool, the hierachy looks something more like this

```
         Tb
        /
Ta  --- Td
        \        
          Tc 
                 \--- Te
                  \___ Tf
```
so even though there are 6 transactions in the mempool, Td will have only 1 ancestor (Ta) but Te will have 2 ancestors (Ta and Tc).
alternatively, i could assert something like: 
```
const rawMempool = await testing.rpc.blockchain.getRawMempool(true)
expect(keys.length).toStrictEqual(rawMempool[txId].ancestorcount.toNumber() - 1)
```
ref line 432@: [https://github.com/DeFiCh/ain/blob/master/src/rpc/blockchain.cpp](https://github.com/DeFiCh/ain/blob/master/src/rpc/blockchain.cpp)

but that would assume that the getRawMempool method returns the correct number of ancestors. let me know your thoughts?

eli-lim: Yup this tree looks ok too. So I would test precisely as you've described:
- T<sub>d</sub> has 1 ancestor T<sub>a</sub>
- T<sub>e</sub> has 2 ancestors T<sub>a</sub> and T<sub>c</sub> (though I'm not sure if T<sub>c</sub> will be returned - hence the test demonstrating so would be useful 🤓)

I think it might not be necessary for multiple wallets to transact in order to build such a tx tree in the mempool, but intuitively I thought that would be the simplest scenario (caveat: haven't personally tried this out myself)

delphk: to address your query, yes, if T<sub>e</sub> has 2 ancestors, both T<sub>a</sub> and T<sub>c</sub> will be returned. the issue is, the above diagram is just for illustration purpose. in reality, T<sub>d</sub> could have 2 ancestors and T<sub>e</sub> could have 1 ancestor, even though T<sub>e</sub> occurred after T<sub>d</sub> (the mempool has its own sorting mechanism as transactions get added). hence it would be difficult to deterministically assert exactly how many ancestors a particular transaction (eg.  T<sub>e</sub>) has in the test, without actually referencing the current state of the mempool (eg. via methods like `testing.rpc.blockchain.getRawMempool(true)`)

weiyuan95: Agree that there are many possible and non-trivial ways to test this 'accurately' (There are a lot of different scenarios as well).  Looking at the [test](https://github.com/DeFiCh/ain/blob/master/test/functional/mempool_packages.py#L43) in the `ain` repo, IMO it sufficiently covers the `getmempoolancestors` case, and we can just do a smoke test in jellyfish.

weiyuan95 (approved):

cwkang1998 (approved):
Would love to have deterministic tests, but I think this is good for now.

fuxingloh (approved):

PR #1786, opened 2022-09-26
fix(status-api): Fix path param to allow special characters in token pair

nichellekoh:
Previously, the params was not able to map certain param such as BRK.B-USD, where there is a special character dot(.) in the api. This will be fixed in the pr, where the string will be taken in as an input before splitting into token and currency param.

### DIFF
--- a/apps/status-api/src/controllers/OracleStatusController.ts
+++ b/apps/status-api/src/controllers/OracleStatusController.ts
@@ -2,6 +2,7 @@ import { Controller, Get, Param } from '@nestjs/common'
 import { WhaleApiClient } from '@defichain/whale-api-client'
 import { Oracle } from '@defichain/whale-api-client/dist/api/oracles'
 import { SemaphoreCache } from '@defichain-apps/libs/caches'
+import { BadRequestApiException } from '../../../whale-api/src/module.api/_core/api.error'
 
 type OracleStatus = 'outage' | 'operational'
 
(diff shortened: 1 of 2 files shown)


apps/status-api/src/controllers/OracleStatusController.ts
```diff
@@ -31,12 +31,12 @@ export class OracleStatusController {
...
-   * @param currency
+   * @param pair
    */
-  @Get('/ticker/:token-:currency')
-  async getOracleRespondStatus (@Param('token') token: string, @Param('currency') currency: string): Promise<{ status: OracleStatus }> {
-    return await this.cachedGet(`oracle-${token}-${currency}`, async () => {
+  @Get('/ticker/:pair')
+  async getOracleRespondStatus (@Param('pair') pair: string): Promise<{ status: OracleStatus }> {
+    const [token, currency] = pair.split('-')
+    return await this.cachedGet(`oracle-${token}`, async () => {
```
joeldavidw:
```suggestion
    return await this.cachedGet(`oracle-${token}-${currency}`, async () => {
```
Are we missing the `${currency}` here or was it removed on purpose?

joeldavidw (changes requested):

apps/status-api/src/controllers/OracleStatusController.ts
nichellekoh: Ah, I wanted to replace it with `pair` but it was missed

nichellekoh:
```suggestion
    return await this.cachedGet(`oracle-${pair}`, async () => {
```

apps/status-api/src/controllers/OracleStatusController.ts
```diff
@@ -31,12 +31,12 @@ export class OracleStatusController {
...
-   * @param token
-   * @param currency
+   * @param pair
    */
-  @Get('/ticker/:token-:currency')
-  async getOracleRespondStatus (@Param('token') token: string, @Param('currency') currency: string): Promise<{ status: OracleStatus }> {
-    return await this.cachedGet(`oracle-${token}-${currency}`, async () => {
+  @Get('/ticker/:pair')
+  async getOracleRespondStatus (@Param('pair') pair: string): Promise<{ status: OracleStatus }> {
+    const [token, currency] = pair.split('-')
```
fuxingloh, line 45: Should probably regex validate the pair to fit certain constraints. Since now you're allowed to `A-B-C` and when `.split('-')` you get 3 items in the array.

fuxingloh (changes requested):

apps/status-api/src/controllers/OracleStatusController.ts
nichellekoh: I have added the regex validation to only allow 2 items; 1 dash before it gets split to token & currency.

fuxingloh (dismissed):

joeldavidw (approved):

fuxingloh (approved):

PR #1790, opened 2022-10-03
feat(packages/testcontainers): new `NativeChainContainer` based on `node-testcontainers`

### LINKED ISSUE #887: Move `@defichain/testcontainers` to use npm testcontainers exclusively

infinia-yzl:

This PR aims to migrate the current `dockerode` implementation to [`node-testcontainers`](https://github.com/testcontainers/testcontainers-node) in `package/testcontainers`. 

The existing dockerode implementation would still be compatible, the new testcontainers use the defacto standards defined in https://github.com/testcontainers/testcontainers-node.

Fixes #887

#### Additional comments?:
Trying to set up the baseline here so tasks can be worked on incrementally. Not all functions are added / tested yet, and they should be addressed moving on.


packages/testcontainers/src/index.ts
```diff
@@ -14,15 +14,15 @@ export const GenesisKeys = RegTestFoundationKeys
...
+export * from './containers/Dockerode/RegTestContainer/Masternode'
+export * from './containers/Dockerode/RegTestContainer/Persistent'
+export * from './containers/Dockerode/RegTestContainer/ContainerGroup'
 export * from './utils'
 
-export * from './containers/RegTestContainer/LoanContainer'
+export * from './containers/Dockerode/RegTestContainer/LoanContainer'
 
-export * from './containers/AppContainer/WhaleSanityContainer'
+export * from './containers/Dockerode/AppContainer/WhaleSanityContainer'
```
fuxingloh: To reduce the scope of change and review fatigue, should not change files that don't need to be changed, especially for this PR, since it should act more as an addition over modification or replacement. It would also help to keep the commit history intact.

packages/testcontainers/src/containers/Hydra/NativeChainContainer.ts
```diff
@@ -0,0 +1,225 @@
...
+    user: 'testcontainers-user',
+    password: 'testcontainers-password'
+  }
+
+  protected startOptions: StartOptions = NativeChainContainer.DefaultStartOptions
+
+  public withStartOptions (startOptions: StartOptions): this {
+    this.startOptions = startOptions
+    return this
+  }
```
fuxingloh: Should change it to `withRpcUser` and `withRpcPassword` since we're moving to Fluent API instead of options object injection.

packages/testcontainers/src/containers/Hydra/NativeChainContainer.ts
```diff
@@ -0,0 +1,225 @@
...
+    if (this.cachedRpcUrl === undefined) {
+      const port = this.getMappedPort(this.networkConfig.ports.rpc)
+      const user = this.startOptions.user
+      const password = this.startOptions.password
+      this.cachedRpcUrl = `http://${user}:${password}@127.0.0.1:${port}/`
+    }
+    return this.cachedRpcUrl
+  }
+
+  async getJsonRpcProvider (): Promise<JsonRpcClient> {
```
fuxingloh:
```suggestion
  async getJsonRpcClient (): Promise<JsonRpcClient> {
```

I would change to `getJsonRpcClient` since "Provider" is more of an EVM standard due to the ability to use different networking protocols for interfacing communication.

packages/testcontainers/src/containers/Hydra/NativeChainContainer.ts
```diff
@@ -0,0 +1,225 @@
...
+      .withStartupTimeout(120_000)
+
+    const startedContainer = new StartedNativeChainContainer(await super.start(), this.startOptions, this.networkConfig)
+    return startedContainer
+  }
+}
+
+export class StartedNativeChainContainer extends AbstractStartedContainer {
+  // Isaac: is this still necessary?
+  protected cachedRpcUrl?: string
```
fuxingloh: Not required with the use of `getMappedPort`, since the previous design didn't separate started state from the non-started state hence the need to dynamically resolve a cached URL. We could use init this field at the constructor.

packages/testcontainers/src/containers/Hydra/NativeChainContainer.ts
```diff
@@ -0,0 +1,225 @@
...
+  }
+
+  /* eslint-disable @typescript-eslint/no-non-null-assertion, no-void */
+  public static readonly PREFIX = 'defichain-testcontainers-'
+
+  /**
+   * Create container and start it immediately waiting for NativeChain to be ready
+   */
+  public async start (): Promise<StartedNativeChainContainer> {
+    const network = await new Network().start()
```
fuxingloh: The network configuration should be injected in with `withNetworkMode`, we shouldn't internalize this; it would limit the usability of testcontainer Fluent API and make it impossible to connect with other docker containers using the same docker network.

packages/testcontainers/src/containers/Hydra/NativeChainContainer.ts
```diff
@@ -0,0 +1,225 @@
...
+
+    const rand = Math.floor(Math.random() * 10000000)
+
+    this.withExposedPorts(...Object.values(this.networkConfig.ports))
+      .withName(`${NativeChainContainer.PREFIX}-${this.networkConfig.name}-${rand}`)
+      .withNetworkMode(network.getName())
+      .withCmd(this.getCmd())
+      .withStartupTimeout(120_000)
```
fuxingloh: I would move this to the constructor, given this `start ()` design it not Fluent like. When consumer uses the `with...` commands, they expected whatever they injecting in to be replaced. This design will override the replacement when they start the container.

```ts
const started = new NativeChainContainer()
  .withStartupTimeout(150_000)
  .start()
```

The above code expects the timeout to be 120s, however, start will always override the timeout making the 150s injection meaningless.

packages/testcontainers/src/containers/Hydra/NativeChainContainer.ts
```diff
@@ -0,0 +1,225 @@
...
+  protected startOptions: StartOptions = NativeChainContainer.DefaultStartOptions
+
+  public withStartOptions (startOptions: StartOptions): this {
+    this.startOptions = startOptions
+    return this
+  }
+
+  protected networkConfig: JellyfishNetwork = getNetwork('testnet')
+
+  public withJellyfishNetwork (networkName: NetworkName): this {
```
fuxingloh: How about `withBlockchianNetwork()` instead of "Jellyfish Network" since jellyfish network is an actual allowed jellyfish chain params on the C++ side.

packages/testcontainers/src/containers/Hydra/NativeChainContainer.ts
```diff
@@ -0,0 +1,225 @@
...
+
+  /**
+   * Convenience Cmd builder with StartOptions
+   */
+  protected getCmd (): string[] {
+    const { user, password } = this.startOptions
+
+    return [
+      'defid',
+      '-printtoconsole',
```
fuxingloh: I think you're missing quite a lot of required flags. E.g. the network. All of the flags currently used on Dockerode design are required.

packages/testcontainers/src/containers/Hydra/NativeChainContainer.ts
```diff
@@ -0,0 +1,225 @@
...
+  }
+
+  protected networkConfig: JellyfishNetwork = getNetwork('testnet')
+
+  public withJellyfishNetwork (networkName: NetworkName): this {
+    this.networkConfig = getNetwork(networkName)
+    return this
+  }
+
+  protected cachedRpcUrl?: string
```
fuxingloh: Not required here anymore, it should only be accessible once it's started.

fuxingloh (changes requested):
Not a full review, just an initial review on `NativeChainContainer.ts`

packages/testcontainers/src/containers/Hydra/NativeChainContainer.ts
infinia-yzl: Yes, I was considering this as well, but I was following the `MetaChainContainer` implementation here: https://github.com/DeFiCh/metachain/blob/main/packages/testcontainers/src/MetaChainContainer.ts#L39

I'll change it.

fuxingloh: Yeah, the implementation there is flawed. Ought to change it there too. Let me open an issue there.

packages/testcontainers/src/containers/Hydra/NativeChainContainer.ts
infinia-yzl: Yup, I will change this.

Worth noting that `MetaChainContainer` also implements it in this manner  https://github.com/DeFiCh/metachain/blob/main/packages/testcontainers/src/MetaChainContainer.ts#L44.

fuxingloh: Yup, I saw it there too. I've opened an issue there too https://github.com/DeFiCh/metachain/issues/107

cwkang1998:
Since this supercede #1249, closing that in favour of this, however that should still be a good reference for this PR.

infinia-yzl:
Also related: https://github.com/testcontainers/testcontainers-node/pull/364.

@cwkang1998 I couldn't get the container restart to work correctly (`/packages/testcontainers/__tests__/containers/NativeChainContainer/NativeChainContainerRestart.test.ts`). Appreciate any pointers or help.

@delphk as mentioned separately, could you please assist with following up on this? Thanks.

infinia-yzl:
@delphk  I'm also on the fence about my current implementation that groups `masternode`, `spv` under the NativeChainContainer classes. Maybe splitting them out and extending NativeChainContainer classes would be cleaner (like how it was previously).

fuxingloh:
> @delphk I'm also on the fence about my current implementation that groups `masternode`, `spv` under the NativeChainContainer classes. Maybe splitting them out and extending NativeChainContainer classes would be cleaner (like how it was previously).

`NativeChainContainer` don't have to be 1-1 compatible with the current design. This implementation is meant for `MetaChainContainer` and follows the `testcontainer-node` fluent-api design for Artemys compatibility.

packages/testcontainers/__tests__/containers/NativeChainContainer/NativeChainContainerRestart.test.ts
```diff
@@ -0,0 +1,115 @@
...
+      .withStartupTimeout(180_000)
+    container = await genericContainer.start()
+    await container.waitForWalletCoinbaseMaturity()
+  })
+
+  afterEach(async () => {
+    await container.stop()
+  })
+
+  it('should create a masternode and mint some blocks with setDeFiConf masternode_operator=', async () => {
```
cwkang1998: In order to test if restart behaviour is working as expected, you'll have to use a testing group for this. Here since you are testing already with a masternode, setting it to a masternode doesn't really change anything. What you are essentially doing here is making a master node a master node again.

Ideally you would want a test case where you have 2 nodes in a network, with one of them being a masternode, and the other not a masternode. And you'll set the non masternode to a masternode and check if it works after restart. You can refer to https://github.com/JellyfishSDK/jellyfish/blob/main/packages/jellyfish-api-core/__tests__/category/masternode/getMasternodeBlocks.test.ts which shows how its done using TestingGroup.

There's no need to test the restart behaviour for the NativeChainContainer, as it already uses the implementation from `testcontainers`, which is tested by them.

cwkang1998: cc @infinia-yzl @delphk

infinia-yzl: Will resolve this in a separate issue.

packages/jellyfish-network/src/Network.ts
```diff
@@ -27,6 +27,12 @@ export interface Network {
   scriptHashPrefix: 0x5a | 0x80 | 0xc4
   /** For message signing. */
   messagePrefix: '\x15Defi Signed Message:\n'
+
+  /** [HYDRA] */
+  ports: {
+    rpc: number
+    p2p: number
+  }
```
fuxingloh: Hmm, let's move this to another PR? Since it can be standalone.

packages/testcontainers/src/containers/NativeChainContainer.ts
```diff
@@ -0,0 +1,671 @@
...
+  rpcPassword: string
+  blockchainNetwork: BlockchainNetwork
+  masterNodeKey?: MasterNodeKey
+}
+
+export class StartedNativeChainContainer extends AbstractStartedContainer {
+  private assumedSpvHeight: number = 0
+  static SPV_EXPIRATION = 10
+  protected rpcUrl: string
+
```
fuxingloh, line 195: I'm thinking if it's cleaner to keep `StartedNativeChainContainer` simple by moving container `Rpc` and `WaitFor` utilities other classes.

```suggestion

  rpc = new NativeChainRpc(this)
  waitFor = new NativeChainWaitFor(rpc)
```

packages/testcontainers/src/containers/NativeChainContainer.ts
```diff
@@ -0,0 +1,671 @@
...
+      masterNodeKey
+    } = this
+    const startedContainer = new StartedNativeChainContainer(
+      await super.start(),
+      { rpcUser, rpcPassword, blockchainNetwork, masterNodeKey })
+
+    if (masterNodeKey != null) {
+      await startedContainer.call('importprivkey', [masterNodeKey.operator.privKey, 'operator', true])
+      await startedContainer.call('importprivkey', [masterNodeKey.owner.privKey, 'owner', true])
+    }
```
fuxingloh: I would move this to `StartedNativeChainContainer` instead of having the logic here.

Since right now "non started code" is causing side-effect in "started code".

fuxingloh: The duties of `NativeChainContainer` is to get to the state of `StartedNativeChainContainer`. Any side-effect that might happen at Start should be at StartedContainer so that polymorphic abstraction can happen.

packages/testcontainers/src/containers/NativeChainContainer.ts
```diff
@@ -0,0 +1,671 @@
...
+      '-eunospayaheight=7',
+      '-fortcanningheight=8',
+      '-fortcanningmuseumheight=9',
+      '-fortcanninghillheight=10',
+      '-fortcanningroadheight=11',
+      '-fortcanningcrunchheight=12',
+      '-fortcanningspringheight=13',
+      '-fortcanninggreatworldheight=14',
+      '-fortcanningepilogueheight=15'
+    ],
```
marktanrj: I think there are missing grand central flags
[reference](https://github.com/JellyfishSDK/jellyfish/blob/43cd88ab5d78caf0c4627079f04a15769f48b530/packages/testcontainers/src/containers/RegTestContainer/index.ts#L43)

```
'-grandcentralheight=16',
'-regtest-skip-loan-collateral-validation',
'-regtest-minttoken-simulate-mainnet=0'
```

marktanrj (commented):
LGTM, 

edit: just a question
I was wondering why `StartedNativeChainContainer` is in the same file as `NativeChainContainer` and not in a separate file? 🤔

packages/testcontainers/src/containers/NativeChainContainer.ts
infinia-yzl: About your question regarding both classes in the same file, I was following the implementation here: https://github.com/testcontainers/testcontainers-node/blob/6f8971796f5a9c50e9cc38651a87ba522a093470/src/modules/postgresql/postgresql-container.ts.

Not a very good answer, but I don't see any problems with both classes being in the same file, hence, no reason to deviate from their standard implementation.

packages/testcontainers/src/containers/NativeChainContainer.ts
infinia-yzl: Yes, thanks for pointing this out.

packages/testcontainers/src/containers/NativeChainContainer.ts
fuxingloh: > About your question regarding both classes in the same file.

Yeah, I guess conventions. Contextually speaking, probably no point separating them into different files either since the context is the same.

packages/jellyfish-network/src/Network.ts
helloscoopa: PR for this is here: https://github.com/JellyfishSDK/jellyfish/pull/1907

packages/jellyfish-network/src/Network.ts
```diff
@@ -27,7 +27,6 @@ export interface Network {
   scriptHashPrefix: 0x5a | 0x80 | 0xc4
   /** For message signing. */
   messagePrefix: '\x15Defi Signed Message:\n'
-
```
fuxingloh: Revert changes in this file

packages/testcontainers/src/containers/NativeChainContainer.ts
```diff
@@ -0,0 +1,238 @@
...
+  get blockchainNetwork (): BlockchainNetwork {
+    return this.config.blockchainNetwork
+  }
+
+  get masterNodeKey (): MasterNodeKey | undefined {
+    return this.config.masterNodeKey
+  }
+
+  public async withPrivateKeys (): Promise<this> {
+    if (this.masterNodeKey != null) {
```
fuxingloh:
```suggestion
  public async withMasternode (keys: MasterNodeKey): Promise<this> {
```

How about something like this?

packages/testcontainers/src/containers/NativeChainRpc.ts
```diff
@@ -0,0 +1,231 @@
+import { ExecResult } from 'testcontainers/dist/docker/types'
+import { StartedNativeChainContainer } from './NativeChainContainer'
+import fetch from 'cross-fetch'
+
+export class NativeChainRpc {
```
fuxingloh, line 5: We should point the old testcontainers RPC calls into this Class so we don't duplicate codes. Having just one "Rpc" helper to maintain.

```ts

export class NativeChainRpc {
  constructor (private readonly rpcUrl: string) {
  }

// ...
```

```ts
export class MasterNodeRegTestContainer extends RegTestContainer {
  private rpc = new NativeChainRpc(this)

  async generate (...): Promise<void> {
    return this.rpc.generate(...);
  }
```

fuxingloh: > the reason behind this suggestion — since we don't plan to deprecate the old testcontainers and we plan to support both environments, this suggestion make sure we always have both maintained and pointing to the same abstraction

infinia-yzl: Yup, I'm working on it now.

packages/testcontainers/src/containers/RegTestContainer/SPV.ts
```diff
@@ -6,7 +6,7 @@ import { RegTestContainer } from './'
 export class SPV {
   static EXPIRATION = 10
   private readonly client: RegTestContainer
-  private assummedSpvHeight: number
+  private readonly assummedSpvHeight: number
```
fuxingloh:
```suggestion
```

Is this safe to remove?

packages/testcontainers/src/containers/RegTestContainer/Masternode.ts
```diff
@@ -17,7 +22,7 @@ export class MasterNodeRegTestContainer extends RegTestContainer {
    */
   constructor (masternodeKey: MasterNodeKey = RegTestFoundationKeys[0], image: string = DeFiDContainer.image, options?: DockerOptions) {
```
fuxingloh:
```suggestion
  constructor (private readonly masternodeKey: MasterNodeKey = RegTestFoundationKeys[0], image: string = DeFiDContainer.image, options?: DockerOptions) {
```

How about this so that we can remove `_masternodeKey` and `get masternodeKey (): MasterNodeKey {`?

packages/testcontainers/__tests__/containers/NativeChainContainer/Coinbase.test.ts
```diff
@@ -0,0 +1,118 @@
...
+import { NativeChainContainer, StartedNativeChainContainer } from '../../../src'
+import waitForExpect from 'wait-for-expect'
+
+describe('coinbase maturity faster by time travel', () => {
+  let container: StartedNativeChainContainer
+
+  beforeAll(async () => {
+    const startedNetwork = await new Network().start()
+    container = await new NativeChainContainer()
+      .withNetworkMode((startedNetwork).getName())
```
fuxingloh:
```suggestion
    const network = await new Network().start()
    container = await new NativeChainContainer()
      .withNetwork(network)
```

You can do this now with testcontainers v9.

packages/testcontainers/src/containers/NativeChainContainer.ts
```diff
@@ -0,0 +1,250 @@
...
+
+  /**
+   * Generate a name for a new docker container with network type and random number
+   */
+  generateName (): string {
+    const rand = Math.floor(Math.random() * 10000000)
+    return `${NativeChainContainer.PREFIX}-${this.blockchainNetwork.name}-${rand}`
+  }
+
+  protected blockchainNetwork: BlockchainNetwork = getNetwork('testnet')
```
fuxingloh:
```suggestion
  protected blockchainNetwork: BlockchainNetwork = getNetwork('regtest')
```

I would default to regtest, since that's 99% of the scenario.

packages/testcontainers/src/containers/NativeChainContainer.ts
```diff
@@ -0,0 +1,250 @@
...
+    this.addedCmds.push(newCmd)
+    return this
+  }
+
+  /**
+   * Create container and start it immediately waiting for NativeChain to be ready
+   */
+  public async start (): Promise<StartedNativeChainContainer> {
+    this.withExposedPorts(...(this.hasExposedPorts ? this.ports : Object.values(this.blockchainNetwork.ports)))
+      .withName(this.name ?? this.generateName())
```
fuxingloh: Move default `withName()` behaviour to constructor since that has batteries attached, you don't want to override name unnecessary breaking testcontainer fluency.

fuxingloh: The rest lke `withExposedPorts` and `withCommand`, it makes sense why we have to do it that way. But best not to touch `withName`.

packages/testcontainers/src/containers/NativeChainContainer.ts
```diff
@@ -0,0 +1,250 @@
...
+    } = this
+
+    const sncc = new StartedNativeChainContainer(
+      await super.start(),
+      {
+        rpcUser,
+        rpcPassword,
+        blockchainNetwork
+      }
+    )
```
fuxingloh:
```suggestion
    const config = {
      rpcUser: this.rpcUser,
      rpcPassword: this.rpcPassword,
      blockchainNetwork: this.blockchainNetwork,
    }

    const sncc = new StartedNativeChainContainer(await super.start(), config)
```

I think this is more concise and better for readability.

packages/testcontainers/src/containers/NativeChainContainer.ts
```diff
@@ -0,0 +1,250 @@
...
+    return (masterNodeKey != null) ? await sncc.withMasterNode(masterNodeKey) : sncc
```
fuxingloh:
```suggestion
    if (this.masterNodeKey != null) {
      await sncc.withMasterNode(this.masterNodeKey)
    }
    return sncc;
```

How about this?

fuxingloh (dismissed):
Generally LGTM, added a few suggestions.

packages/testcontainers/src/containers/NativeChainContainer.ts
```diff
@@ -0,0 +1,245 @@
...
+    this.masterNodeKey = key
+    return this
+  }
+
+  protected hasPreconfiguredMasternode: boolean = false
+
+  public withPreconfiguredRegtestMasternode (masterNodeKey = RegTestFoundationKeys[0]): this {
+    this.withMasterNodeKey(masterNodeKey)
+    this.withBlockchainNetwork('regtest')
+    this.hasPreconfiguredMasternode = true
```
fuxingloh: You can check if `masterNodeKey !== undefined` instead of having another field?

fuxingloh (approved):
Generally, this LGTM.

fuxingloh:
@cwkang1998 can you do a review?

packages/testcontainers/src/containers/RegTestContainer/Masternode.ts
```diff
@@ -1,23 +1,22 @@
 import { MasterNodeKey, RegTestFoundationKeys } from '@defichain/jellyfish-network'
-import { waitForCondition } from '../../utils'
 import { DockerOptions } from 'dockerode'
 import { DeFiDContainer, StartOptions } from '../DeFiDContainer'
 import { RegTestContainer } from './index'
+import { NativeChainWaitFor } from '../../index'
```
cwkang1998, line 5: Considering you're also exporting `Masternode` in `../../index.ts`, I feel like this should've caused a circular dependency issue.

A quick check using `madge` reveals it as well:

```
❯ madge . --circular
Processed 17 files (294ms) 

✖ Found 11 circular dependencies!

1) dist/containers/AppContainer/WhaleSanityContainer.js > dist/containers/AppContainer/index.js > dist/containers/RegTestContainer/Masternode.js > dist/containers/DeFiDContainer.js > dist/index.js
2) dist/containers/DeFiDContainer.js > dist/index.js
3) dist/containers/DeFiDContainer.js > dist/index.js > dist/containers/MainNetContainer.js
4) dist/containers/NativeChainContainer.js > dist/containers/NativeChainRpc.js
5) dist/index.js > dist/containers/NativeChainContainer.js > dist/containers/NativeChainRpc.js
6) dist/containers/RegTestContainer/Masternode.js > dist/containers/DeFiDContainer.js > dist/index.js > dist/containers/RegTestContainer/LoanContainer.js
7) dist/containers/RegTestContainer/Masternode.js > dist/containers/DeFiDContainer.js > dist/index.js
8) dist/containers/DeFiDContainer.js > dist/index.js > dist/containers/RegTestContainer/Persistent.js
9) dist/containers/RegTestContainer/Masternode.js > dist/containers/DeFiDContainer.js > dist/index.js > dist/containers/RegTestContainer/Persistent.js
10) dist/containers/DeFiDContainer.js > dist/index.js > dist/containers/RegTestContainer/index.js
11) dist/containers/DeFiDContainer.js > dist/index.js > dist/containers/TestNetContainer.js
```

A pretty simple fix are to remove this imports from `index.ts` where you export your file, and the circular dep problem will be gone.

packages/testcontainers/package.json
```diff
@@ -18,7 +18,8 @@
     "cross-fetch": "^3.1.5",
     "dockerode": "^3.3.4",
     "tar-fs": "^2.1.1",
-    "uuid": "^9.0.0"
+    "uuid": "^9.0.0",
+    "testcontainers": "^9.0.1"
```
cwkang1998, line 22: Will probably need a bump, latest at `9.1.0`.

cwkang1998 (approved):
Minor `nit` addressable in a future PR. Otherwise LGTM!

cwkang1998:
In future PRs, can probably add tests for restart as well for `NativeChainContainer`.
 https://github.com/JellyfishSDK/jellyfish/blob/9ba3b137d90613e5110d02eff117069c89d55d65/packages/testcontainers/__tests__/containers/ContainerRestart.test.ts.

PR #1841, opened 2022-10-25
feat(v3): bump to grandcentral and use `master-1d169a751`

shohamc1:

Bump testcontainers to `grandcentral` with `master-1d169a75`. Fixes all broken tests and known regression from this change.

### DIFF
--- a/apps/rich-list-api/docker-compose.yml
+++ b/apps/rich-list-api/docker-compose.yml
@@ -47,4 +47,5 @@ services:
       -fortcanningspringheight=13
       -fortcanninggreatworldheight=14
       -fortcanningepilogueheight=15
+      -grandcentralheight=16
       -regtest-skip-loan-collateral-validation
--- a/apps/rich-list-api/docker-compose.yml
+++ b/apps/rich-list-api/docker-compose.yml
@@ -13,7 +13,7 @@ services:
       POSTGRES_DB: defichainrichlist
 
   defi-blockchain:
-    image: defi/defichain:master-fff6b7794
+    image: defi/defichain:master-1d169a751
     ports:
       - "19554:19554"
     command: >
--- a/apps/rich-list-api/docker-compose.yml
+++ b/apps/rich-list-api/docker-compose.yml
@@ -13,7 +13,7 @@ services:
       POSTGRES_DB: defichainrichlist
 
   defi-blockchain:
-    image: defi/defichain:master-23c76b3b9
+    image: defi/defichain:HEAD-aae120d78
     ports:
       - "19554:19554"
     command: >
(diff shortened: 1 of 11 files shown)


apps/rich-list-api/docker-compose.yml
```diff
@@ -13,7 +13,7 @@ services:
       POSTGRES_DB: defichainrichlist
 
   defi-blockchain:
-    image: defi/defichain:HEAD-49fba65ce
+    image: defi/defichain:epic-grandcentral-bfb0eedb7
```
fuxingloh: You should update the `command: >` as well with the new flag.

packages/jellyfish-api-core/__tests__/category/mining/getMiningInfo.test.ts
```diff
@@ -39,6 +39,6 @@ describe('Mining', () => {
     expect(info.networkhashps).toBeGreaterThan(0)
     expect(info.pooledtx).toStrictEqual(0)
     expect(info.chain).toStrictEqual('regtest')
-    expect(info.warnings).toStrictEqual('')
+    expect(['', 'This is a pre-release test build - use at your own risk - do not use for mining or merchant applications']).toContain(info.warnings)
```
fuxingloh, line 42: We should not change the order of assertions just so that the logic flows that way since assertions are strictly `expect(something).toBe(expected)`.

For other projects we use extended, https://jest-extended.jestcommunity.dev/docs/matchers/toBeOneOf

Since we don't have it here, I suggest we use Regex Matcher?

packages/jellyfish-api-core/src/category/token.ts
```diff
@@ -98,6 +98,21 @@ export class Token {
...
+   *
+   * @param {string} amounts Amount as json string, or array. Example: '[ \"amount@token\" ]'
+   * @param {string} from Address containing tokens to be burned
+   * @param {string} context Additional data necessary for specific burn type
+   * @param {UTXO[]} [utxos = []] A json array of json objects. Provide it if you want to spent specific UTXOs
+   * @param {string} [utxos.txid] The transaction id
+   * @param {number} [utxos.vout] The output number
+   * @return {Promise<string>} The hex-encoded hash of broadcasted transaction
+   */
+  async burnTokens (amounts: string, from: string, context?: string, utxos: UTXO[] = []): Promise<string> {
```
fuxingloh, line 113: Not tested?

fuxingloh (approved):
Generally, LGTM and added a few comments.

fuxingloh:
> Generally, LGTM and added a few comments.

Also, write a descriptive PR; it's easier for reviewers with no context. 

PR description gets squashed into the commit as a commit message, so that's important to keep it descriptive and accurate.

packages/jellyfish-api-core/src/category/token.ts
shohamc1: Seems like this was part of the cherry-pick I did on a10afe2657ee00ae0 from #1860. The tests are on [that PR](https://github.com/JellyfishSDK/jellyfish/pull/1860/files#diff-b78b933dcd14e178a555768bb95090dba23b139884641dc16ecb1c784e000fd6).

PR #1843, opened 2022-10-26
feat(jellyfish-api-core): add net getNetTotals RPC

marktanrj:
/kind feature

Fixes part of #48 
- Implements `getnettotals` type for RPC.

### DIFF
--- a/packages/jellyfish-api-core/__tests__/category/net/getNetTotals.test.ts
+++ b/packages/jellyfish-api-core/__tests__/category/net/getNetTotals.test.ts
@@ -39,6 +39,7 @@ describe('Network on masternode', () => {
 
   beforeAll(async () => {
     await container.start()
+    await container.generate(1)
   })
 
   afterAll(async () => {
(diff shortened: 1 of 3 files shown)


packages/jellyfish-api-core/__tests__/category/net/getNetTotals.test.ts
```diff
@@ -0,0 +1,65 @@
...
+    })
+  })
+})
+
+describe('Network on masternode', () => {
+  const container = new MasterNodeRegTestContainer()
+  const client = new ContainerAdapterClient(container)
+
+  beforeAll(async () => {
+    await container.start()
```
fuxingloh, line 41:
```suggestion
    await container.start()
    await container.generate(1)
```

Should test it with side effects to see if there are any changes, otherwise, it's the same scenario without masternode.

fuxingloh (changes requested):
LGTM, with 1 request

fuxingloh (approved):

PR #1853, opened 2022-11-08
feat(whale-api): use best path on total liquidity in usd

### LINKED ISSUE #1852: `poolpair`: update `getTotalLiquidityUsd` to use `dexPrices` for accurate results
Due to introduction of `csETH-dETH`, we now have a pair that is not stablecoin or DFI based. The function to calculate `totalLiquidity.usd` only works with DUSD, USDT, USDC or DFI pairs. 
#### Why is this needed:
To be able to provide a more accurate value for USD. Another thing that is incorrect also for `getTotalLiquidityUsd`, is it assumes that stablecoins are always $1.

https://github.com/JellyfishSDK/jellyfish/blob/4048de6159246a437c8e3823b1e93cad9a576dbd/apps/whale-api/src/module.api/poolpair.service.ts#L77

pierregee:
- [x] Update the total liquidity to use the graph (best path) if pair has no stablecoin
- [x] Added `H-I` liquidity to be able to test the best path api
image

Fixes #1852 

#### Additional comments?:
- getUSDVolume and aggregated USD still uses fix derivation (need to update? CC: @thedoublejay)
- Note: The solution only uses path to USDT - it doesn't average USDC and USDT

### DIFF
--- a/apps/whale-api/src/module.api/poolpair.service.ts
+++ b/apps/whale-api/src/module.api/poolpair.service.ts
@@ -29,11 +29,13 @@ import { AccountHistory } from '@defichain/jellyfish-api-core/dist/category/acco
 import { DeFiDCache } from './cache/defid.cache'
 import { parseDisplaySymbol } from './token.controller'
 import { RpcApiError } from '@defichain/jellyfish-api-core'
+import { PoolSwapPathFindingService } from './poolswap.pathfinding.service'
 
 @Injectable()
 export class PoolPairService {
   constructor (
     @Inject('NETWORK') protected readonly network: NetworkName,
+    private readonly poolSwapPathfindingService: PoolSwapPathFindingService,
     protected readonly rpcClient: JsonRpcClient,
     protected readonly deFiDCache: DeFiDCache,
     protected readonly cache: SemaphoreCache,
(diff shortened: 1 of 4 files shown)


apps/whale-api/src/module.api/poolpair.service.ts
```diff
@@ -70,32 +72,31 @@ export class PoolPairService {
...
-    const [a, b] = info.symbol.split('-')
-    if (['DUSD', 'USDT', 'USDC'].includes(a)) {
-      return info.reserveA.multipliedBy(2)
-    }
+    const usdtToken = await this.deFiDCache.getTokenInfoBySymbol('USDT')
 
-    if (['DUSD', 'USDT', 'USDC'].includes(b)) {
-      return info.reserveB.multipliedBy(2)
+    if (usdtToken === undefined) {
+      throw new NotFoundException('Unable to find USDT token')
```
thedoublejay: Should we throw error here? Or just keep it as undefined like the current behavior

thedoublejay (changes requested):

apps/whale-api/src/module.api/poolpair.service.ts
```diff
@@ -70,32 +72,31 @@ export class PoolPairService {
...
-      return
-    }
+    const usdtTokenId = Object.keys(usdtToken)[0]
+    let tokenARate = new BigNumber(1)
+    let tokenBRate = new BigNumber(1)
 
-    if (a === 'DFI') {
-      return info.reserveA.multipliedBy(2).multipliedBy(USDT_PER_DFI)
+    if (info.idTokenA !== usdtTokenId) {
+      const { estimatedReturn: estimatedReturnTokenA } = await this.poolSwapPathfindingService.getBestPath(info.idTokenA, usdtTokenId)
```
canonbrother, line 90: hmm... is that possible estimatedReturn zero or negative value??

packages/whale-api-client/__tests__/api/poolpairs.test.ts
```diff
@@ -615,14 +623,14 @@ describe('poolswap', () => {
         displaySymbol: 'DUSD'
       },
       apr: {
-        reward: 0,
-        total: 0.12174783188792529,
-        commission: 0.12174783188792529
+        reward: null,
+        total: null,
+        commission: null
```
canonbrother: hmm.. can help verify what causes it `null` ?? :thinking:

pierregee: Hi, yup it's expected. `TEST-DUSD` is isolated in the graph, that's why it can't be computed and returns `null`. I kept the test as is and didn't add a path to test this scenario too

pierregee: Can also add DUSD-DFI pair so it will not be null, if preferred (?)

pierregee: If there's a path to USDT, then this can be computed 
image

pierregee: For visualization 😄 
image

canonbrother (dismissed):

thedoublejay (dismissed):

apps/whale-api/src/module.api/poolpair.service.ts
```diff
@@ -70,32 +72,31 @@ export class PoolPairService {
   }
 
   /**
-   * TODO(fuxingloh): graph based matrix resolution
-   * Currently implemented with fix pair derivation
-   * Ideally should use vertex directed graph where we can always find total liquidity if it can be resolved.
+   * Graph based matrix resolution
+   * Returns the total liquidity of the poolpair by finding its best path to USDT
```
fuxingloh, line 107: I'm worried about the performance.

But I assume this method is cached, shouldn't be much of a hit on performance.

apps/whale-api/src/module.api/poolpair.service.ts
```diff
@@ -72,8 +74,8 @@ export class PoolPairService {
   /**
    * TODO(fuxingloh): graph based matrix resolution
    * Currently implemented with fix pair derivation
-   * Ideally should use vertex directed graph where we can always find total liquidity if it can be resolved.
-   */
+  * Ideally should use vertex directed graph where we can always find total liquidity if it can be resolved.
+  */
```
fuxingloh:
```suggestion
   * Ideally should use vertex directed graph where we can always find total liquidity if it can be resolved.
   */
```

apps/whale-api/src/module.api/poolpair.service.ts
```diff
@@ -96,6 +98,36 @@ export class PoolPairService {
...
     }
+
+    return await this.getTotalLiquidityUsdByBestPath(info)
+  }
+
+  /**
+   * Graph based matrix resolution
+   * Returns the total liquidity of the poolpair by finding its best path to USDT
+   */
+  async getTotalLiquidityUsdByBestPath (info: PoolPairInfo): Promise<BigNumber | undefined> {
```
fuxingloh, line 109: Should add additional test scenarios to cover these scenarios (those that were previously included).

fuxingloh (changes requested):
Generally, LGTM with 1 request change.

apps/whale-api/src/module.api/poolpair.service.ts
pierregee: Sure! Added `H-I` to test this scenario - [link](https://github.com/JellyfishSDK/jellyfish/pull/1853/files#diff-229e1b7ca9e66f6d396e594f84ff4cd6fcfb35321aca2aca2e312864279b462cR703). The ids and values adjusted too 😬 
 image

fuxingloh (approved):

PR #1856, opened 2022-11-10
fix(jellyfish-api-core): update testPoolSwap arguments

shohamc1:

Exposes the path and verbose arguments on `testpoolswap`. Removes the defunct UTXO argument.

### DIFF
--- a/packages/jellyfish-api-core/__tests__/category/poolpair/testpoolswap.test.ts
+++ b/packages/jellyfish-api-core/__tests__/category/poolpair/testpoolswap.test.ts
@@ -196,6 +196,44 @@ describe('Poolpair', () => {
     })
   })
 
+  it('should throw error when no tokenFrom', async () => {
+    const promise = client.poolpair.testPoolSwap({
+      from: await getNewAddress(container),
+      tokenFrom: '',
+      amountFrom: 13,
+      to: await getNewAddress(container),
+      tokenTo: 'DFI'
+    })
+
+    await expect(promise).rejects.toThrow(RpcApiError)
+    await expect(promise).rejects.toMatchObject({
+      payload: {
+        code: -8,
+        message: 'tokenFrom is empty',
+        method: 'testpoolswap'
+      }
+    })
+  })
+
+  it('should throw error when no tokenTo', async () => {
+    const promise = client.poolpair.testPoolSwap({
+      from: await getNewAddress(container),
+      tokenFrom: 'DFI',
+      amountFrom: 13,
+      to: await getNewAddress(container),
+      tokenTo: ''
+    })
+
+    await expect(promise).rejects.toThrow(RpcApiError)
+    await expect(promise).rejects.toMatchObject({
+      payload: {
+        code: -8,
+        message: 'tokenTo is empty',
+        method: 'testpoolswap'
+      }
+    })
+  })
+
   it('should fail if wrong path option is used', async () => {
     const tokenBatAddress = await getNewAddress(container)
     const poolLiquidityAddress = await getNewAddress(container)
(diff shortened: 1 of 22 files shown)


packages/jellyfish-api-core/__tests__/category/poolpair/testpoolswap.test.ts
```diff
@@ -57,6 +63,9 @@ describe('Poolpair', () => {
       tokenTo: 'DFI'
     }) // 199.99999729@0
 
+    if (typeof result !== 'string') {
+      throw new Error(`result was ${typeof result}, expected string`)
+    }
```
cwkang1998: Since this is a test, I think a more appropriate way of doing this is:

```suggestion
    expect(typeof result).toStrictEqual('string')
```

packages/jellyfish-api-core/src/category/poolpair.ts
```diff
@@ -153,10 +153,12 @@ export class PoolPair {
...
    * @param {string} metadata.tokenTo swap to token {symbol/id}
    * @param {number} [metadata.maxPrice] acceptable max price
-   * @return {Promise<string>} formatted as 'amount@token' swapped
+   * @param {'auto' | 'direct' | 'composite' | Array<string>} [path] swap path to use, defaults to auto. Provide array of poolpair IDs to set path manually.
+   * @param {boolean} [verbose] return pool path used, defaults to false
+   * @return {Promise<string | TestPoolSwapVerboseResult>} formatted as 'amount@token' swapped
    */
-  async testPoolSwap (metadata: PoolSwapMetadata): Promise<string> {
-    return await this.client.call('testpoolswap', [metadata], 'bignumber')
+  async testPoolSwap (metadata: PoolSwapMetadata, path: 'auto' | 'direct' | 'composite' | string[] = 'auto', verbose: boolean = false): Promise<string | TestPoolSwapVerboseResult> {
```
cwkang1998: Remember to document this change in the md file as well.

cwkang1998: You can also use multiple function signature to indicate what will be returned given which parameters, so that dev will have a better experience. Something like `getAccount` in https://github.com/JellyfishSDK/jellyfish/blob/main/packages/jellyfish-api-core/src/category/account.ts#L136-L173.

cwkang1998 (changes requested):
Looks good, just some minor changes

cwkang1998 (dismissed):
LGTM.

packages/jellyfish-api-core/src/category/poolpair.ts
```diff
@@ -153,10 +153,34 @@ export class PoolPair {
...
+   * @param {number} [metadata.maxPrice] acceptable max price
+   * @param {'auto' | 'direct' | 'composite' | Array<string>} [path] swap path to use, defaults to auto. Provide array of poolpair IDs to set path manually.
+   * @param {boolean} [verbose=true] return pool path used, defaults to false
+   * @return {Promise<string>} formatted as 'amount@token' swapped
+   */
+  testPoolSwap (metadata: PoolSwapMetadata, path?: 'auto' | 'direct' | 'composite' | string[], verbose?: true): Promise<TestPoolSwapVerboseResult>
+
+  async testPoolSwap (
+    metadata: PoolSwapMetadata,
+    path: 'auto' | 'direct' | 'composite' | string[] = 'auto',
```
fuxingloh, line 180: Could we have all the conditions tested?

fuxingloh: And what is `string[]`?

shohamc1: `string[]` is an array of poolpair IDs, for when the user wants to manually specify the composite swap path.

fuxingloh (approved):

PR #1859, opened 2022-11-10
feat(jellyfish-api-core): unsetgov RPC

shohamc1:

Provide support for `unsetgov` RPC in v3.

#### Additional comments?:

Requires GrandCentral fork to be active, and requires a newer container.

### DIFF
--- a/packages/jellyfish-api-core/__tests__/category/masternode/unsetGov.test.ts
+++ b/packages/jellyfish-api-core/__tests__/category/masternode/unsetGov.test.ts
@@ -1,0 +1,73 @@
+import { RpcApiError } from '@defichain/jellyfish-api-core'
+import { MasterNodeRegTestContainer } from '@defichain/testcontainers'
+import { createToken } from '@defichain/testing'
+import { ContainerAdapterClient } from '../../container_adapter_client'
+
+describe('Masternode', () => {
+  const container = new MasterNodeRegTestContainer()
+  const client = new ContainerAdapterClient(container)
+
+  beforeAll(async () => {
+    await container.start()
+    await container.waitForReady()
+    await container.waitForWalletCoinbaseMaturity()
+    await container.generate(20)
+
+    await createToken(container, 'CAT')
+  })
+
+  afterAll(async () => {
+    await container.stop()
+  })
+
+  it('should throw error if not enabled', async () => {
+    const promise = client.masternode.unsetGov({ ICX_TAKERFEE_PER_BTC: '' })
+
+    await expect(promise).rejects.toThrow(RpcApiError)
+    await expect(promise).rejects.toMatchObject({
+      payload: {
+        code: -32600,
+        message: 'Test UnsetGovVariableTx execution failed:\nUnset Gov variables not currently enabled in attributes.',
+        method: 'unsetgov'
+      }
+    })
+  })
+
+  it('should unset governance variables', async () => {
+    await client.masternode.setGov({ ICX_TAKERFEE_PER_BTC: 0.001 })
+    await client.masternode.setGov({ ATTRIBUTES: { 'v0/params/feature/gov-unset': 'true' } })
+    await container.generate(1)
+
+    const govBefore = await client.masternode.getGov('ICX_TAKERFEE_PER_BTC')
+    expect(govBefore.ICX_TAKERFEE_PER_BTC.toNumber()).toStrictEqual(0.001)
+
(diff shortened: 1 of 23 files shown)


packages/testcontainers/src/containers/DeFiDContainer.ts
```diff
@@ -35,7 +35,7 @@ export abstract class DeFiDContainer extends DockerContainer {
     if (process?.env?.DEFICHAIN_DOCKER_IMAGE !== undefined) {
       return process.env.DEFICHAIN_DOCKER_IMAGE
     }
-    return 'defi/defichain:HEAD-49fba65ce'
+    return 'defi/defichain:master-1dff273bf'
```
fuxingloh: For this change, do we need to update to the latest image? 

Or will it also work with the existing image?

shohamc1: I believe the new image is required, the RPC was added in https://github.com/DeFiCh/ain/commit/e8ca8a65ab980fd2ed7ba978bbc7400bb5ac0655.

fuxingloh: Oh, I guess we should merge after v3 PR are merged since there is breaking changes with that image.

packages/jellyfish-api-core/__tests__/category/masternode/unsetGov.test.ts
```diff
@@ -0,0 +1,73 @@
...
+
+describe('Masternode', () => {
+  const container = new MasterNodeRegTestContainer()
+  const client = new ContainerAdapterClient(container)
+
+  beforeAll(async () => {
+    await container.start()
+    await container.waitForReady()
+    await container.waitForWalletCoinbaseMaturity()
+    await container.generate(20)
```
cwkang1998: Why is 20 blocks required to be generated? If there's a reason maybe you can add in comments to indicate as well.

packages/jellyfish-api-core/__tests__/category/masternode/unsetGov.test.ts
```diff
@@ -0,0 +1,73 @@
...
+
+    await createToken(container, 'CAT')
+  })
+
+  afterAll(async () => {
+    await container.stop()
+  })
+
+  it('should throw error if not enabled', async () => {
```
cwkang1998, line 22: Maybe can add another test case to cover the case for when a non-set governance variable / attributes is unset.

packages/jellyfish-api-core/src/category/masternode.ts
```diff
@@ -175,6 +175,21 @@ export class Masternode {
     return await this.client.call('setgovheight', [input, activationHeight, utxos], 'number')
   }
 
+  /**
+   * Unset special governance variables
+   *
+   * @param {Record<string, any>} input json object
```
cwkang1998: From looking at https://github.com/DeFiCh/ain/blob/254bd6292b1ca280b4bbd1273473037cd08c8825/src/masternodes/mn_rpc.cpp#L582

This typing can be:

```suggestion
   * @param {Record<string, number | string | Array<string>>} input json object
```

cwkang1998 (changes requested):

cwkang1998:
Also, do remember to add this into the docs, and also resolve the merge conflict

packages/jellyfish-api-core/src/category/masternode.ts
```diff
@@ -175,6 +175,20 @@ export class Masternode {
...
+   * Unset special governance variables
+   *
+   * @param {Record<string, number | string | string[]>} input json object
+   * @param {UTXO[]} [utxos = []] Specific utxos to spend
+   * @param {string} [utxos.txid] The transaction id
+   * @param {string} [utxos.vout] The output number
+   * @return {Promise<string>} hash
+   *
+   */
+  async unsetGov (input: Record<string, number | string | string[]>, utxos: UTXO[] = []): Promise<string> {
```
cwkang1998: `nit`: lets call 'input', 'variables' instead, following what's used in the defichain rpc, which is more descriptive.

```suggestion
  async unsetGov (variables: Record<string, number | string | string[]>, utxos: UTXO[] = []): Promise<string> {
```

cwkang1998 (changes requested):

packages/jellyfish-api-core/src/category/masternode.ts
cwkang1998: Other than this should be good!

fuxingloh (approved):

PR #1871, opened 2022-11-24
feat(playground-api): set consortium gov attrs in playground

pierregee:
- To enable consortium in playground and have default consortium members + limit


pierregee:
Creating PR directly instead of the forked branch :D

PR #1872, opened 2022-11-25
feat(jellyfish-api-core): add getNodeAddress RPC

shancht:
**What this PR does / why we need it:** /kind feature

**Which issue(s) does this PR fixes?:**  Adding getnodeaddresses

**Usage**: Fixes getnodeaddress @ [Issue#48](https://github.com/JellyfishSDK/jellyfish/issues/48)

### DIFF
--- a/packages/jellyfish-api-core/__tests__/category/net/getNodeAddresses.test.ts
+++ b/packages/jellyfish-api-core/__tests__/category/net/getNodeAddresses.test.ts
@@ -1,27 +1,38 @@
 import { MasterNodeRegTestContainer } from '@defichain/testcontainers'
-import { TestingGroup } from '@defichain/jellyfish-testing'
 import { ContainerAdapterClient } from '../../container_adapter_client'
 
 describe('Network on masternode', () => {
   const container = new MasterNodeRegTestContainer()
   const client = new ContainerAdapterClient(container)
-  const group = TestingGroup.create(2)
 
   beforeAll(async () => {
-    await container.start()
-    await group.start()
+    await container.start({
+      startFlags: [
+        {
+          name: '-regtest',
+          value: 0
+        },
+        {
+          name: '-testnet',
+          value: 1
+        },
+        {
+          name: '-rpcport',
+          value: 19554
+        }
+      ]
+    }, false)
   })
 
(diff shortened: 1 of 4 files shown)


packages/jellyfish-api-core/__tests__/category/net/getNodeAddresses.test.ts
```diff
@@ -0,0 +1,27 @@
...
+
+  beforeAll(async () => {
+    await container.start()
+    await group.start()
+  })
+
+  afterAll(async () => {
+    await container.stop()
+    await group.stop()
+  })
```
fuxingloh:
```suggestion
  const group = TestingGroup.create(2)

  beforeAll(async () => {
    await group.start()
  })

  afterAll(async () => {
    await group.stop()
  })
```

When you're using `TestingGroup.create(numberOfContainer)`, you essentially setup multiple containers. Hence you don't need to setup anymore containers.

Accessing container at index 0:
`group.get(0).rpc.getNodeAddresses(1)`

Accessing container at index 1:
`group.get(1).rpc.getNodeAddresses(1)`

packages/jellyfish-api-core/__tests__/category/net/getNodeAddresses.test.ts
```diff
@@ -0,0 +1,27 @@
...
+
+  it('should get something in the array', async () => {
+    const address = await container.getNewAddress()
+    await container.addNode(address)
+
+    const info = await client.net.getNodeAddresses(1)
+    console.log(info)
```
fuxingloh: Remember to assert

```ts
expect(info).toStrictEqual([
  {},
  {},
])
```

fuxingloh (changes requested):

fuxingloh:
Also, we're using "Semantic Pull Request", you should relabel your title.

shohamc1:
Note on the `should have available nodes` test - it is potentially flaky since the node discovery process might not complete in 10 seconds. ~Additionally the test would break if any of the mydeficha.in testnet nodes went down or changed their IP addresses.~

Edit: External node dependency solved with 34d19e081fcc1

PR #1863, opened 2022-11-18
feat(jellyfish-transaction): Add updatemasternode dftx

marktanrj:
#### What this PR does / why we need it:

- updatemasternode DfTx parsing

#### Which issue(s) does this PR fixes?:

Fixes part of #1842

PR #1892, opened 2022-12-01
feat(jellyfish-transaction-builder): Extend createDeFiTx to allow for custom vins and vouts

marktanrj:
- Allow use of custom vins and vouts for updatemasternode dftx in #1863 to provide collateral tx when updating owner address

### DIFF
--- a/packages/jellyfish-transaction-builder/src/txn/txn_builder.ts
+++ b/packages/jellyfish-transaction-builder/src/txn/txn_builder.ts
@@ -89,6 +89,8 @@ export abstract class P2WPKHTxnBuilder {
    * Craft a transaction with OP_DEFI_TX from the output of OP_CODES.OP_DEFI_TX_.
    * This is a helper method for creating custom defi transactions.
    *
+   * Optionally, allow for custom inputs with vin and vout.
+   *
    * As DeFi custom transaction will always require small amount of DFI,
    * collectPrevouts() is set to search for at least 0.001 DFI amount of prevout.
    * This will also evidently merge small prevout during the operation.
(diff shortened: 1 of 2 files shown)


imkven (approved):
LGTM.

infinia-yzl (approved):

packages/jellyfish-transaction-builder/src/txn/txn_builder.ts
```diff
@@ -98,15 +100,21 @@ export abstract class P2WPKHTxnBuilder {
...
    * @param {Script} changeScript to send unspent to after deducting the fees
    * @param {BigNumber} [outValue=0] for the opDeFiTx, usually always be 0.
+   * @param {Array<{ vin: Vin, vout: Vout, prevout: Prevout }>} [customVinVout = []] for custom vin and vout
    */
   async createDeFiTx (
     opDeFiTx: OP_DEFI_TX,
     changeScript: Script,
-    outValue: BigNumber = new BigNumber('0')
+    outValue: BigNumber = new BigNumber('0'),
+    customVinVout: Array<{ vin: Vin, vout: Vout, prevout: Prevout }> = []
```
fuxingloh, line 109: This overloader concept isn't really applicable and it doesn't gel well.

- Given that can inject any `Vin`, the `Vin` can already be collected by `await this.collectPrevouts(minFee)` and included in vins, giving you an invalid Transaction due to duplicated Vins.
- The assumption is with the `Vin` provided that you have the signing rights to it.
- That `{Vin, Vout, Prevout}` comes in a Set, that's not true since only `Vin` is tied to `Prevout`. The amount `Vin` to `Vout` ratio is not exclusive. So an array won't make sense.

fuxingloh: We also have to ensure collectPrevout doesn't collect the `Vin` you need to sign in a special way.

fuxingloh: Your setup requires:
1. A custom "Prevouts Collector" implementation since you need control of the `Vins`
2. Injection of `Vout` in a special order.

fuxingloh: Would it make sense for the upstream implementation to just prepare their Transaction from scratch?

We change `protected async sign` to `public sync sign` to allow that. And anything else that is required we make it public or create an abstraction for it.

marktanrj: Hmm, a little lost 😓 correct me if I'm wrong:
- I will implement a "Prevouts Collector" as a method in `P2WPKHTxnBuilder` class
- then combine the outputs of it along with outputs of `await this.collectPrevouts(minFee)` so Vin and Vout can be placed in the correct order that doesn't have duplicates?

marktanrj: > prepare their Transaction from scratch

do you mean build the transaction from scratch with vin and vout and then sign it, all in the updatemasternode function in `packages/jellyfish-transaction-builder/src/txn/txn_builder_masternode.ts`?

image

fuxingloh: Yup, that's right. Since your requirements require vin and vout to be arranged in a certain way. Nobody else will probably need it. It is better to keep the custom logic there instead of bleeding it into an abstraction.

marktanrj: Ok, that makes sense. I will do it this way since no other functions need it 👍

marktanrj:
I will build the updatemasternode transaction from scratch in #1894 in `txn_builder_masternode.ts`, no need for an abstraction with this PR

Issue #1842, opened 2022-10-25
Support `CustomTxType::UpdateMasternode` on Jellyfish


fuxingloh:
#### What would you like to be added:

> Support https://github.com/DeFiCh/ain/pull/1437

Add `CustomTxType::UpdateMasternode` DfTx support. This includes: 
- RPC typing @marktanrj #1847
- DfTx parsing @marktanrj #1863
- `whale-api` Masternode Index changes @imkven #1869
- DfTx transaction builder @marktanrj #1894

fuxingloh:
You might want to tackle with multiple PRs, instead of implementing them all at once.

PR #1894, opened 2022-12-01
feat(jellyfish-transaction-builder): Add updatemasternode transaction builder

marktanrj:
- updatemasternode transaction builder

Fixes part of #1842

#### Additional comments?:
- #1863 needed for typings
- <del> #1892 needed for createDeFiTx </del>


packages/jellyfish-transaction-builder/src/txn/txn_builder_masternode.ts
```diff
@@ -41,4 +48,87 @@ export class TxnBuilderMasternode extends P2WPKHTxnBuilder {
...
+      value: total,
+      script: changeScript,
+      tokenId: 0x00
+    }
+
+    const mergedVin = [...vin]
+    const mergedVout = [deFiOut, change]
+    const mergedPrevouts = [...prevouts]
+
+    if (collateralInfo != null) {
```
delphk: should prefer strict equality checking (` !==` )

marktanrj: Thanks, amended

infinia-yzl (dismissed):

delphk (commented):
Similar to the RPC tests, it would be good to check that the masternodes have been updated correctly through `listMasternodes()` in the tests

delphk (dismissed):
LGTM

infinia-yzl (dismissed):

packages/jellyfish-transaction-builder/src/txn/txn_builder_masternode.ts
```diff
@@ -1,11 +1,18 @@
+import { RawTransaction } from '@defichain/jellyfish-api-core/dist/category/rawtx'
```
fuxingloh: This type isn't an included dependency in `jellyfish-transaction-builder`; it won't be available unless explicitly installed.

fuxingloh: It also creates an expectation that consumer of this class uses `jellyfish-api-core` to build the transaction. They might be using whale which don't provide this information.

packages/jellyfish-transaction-builder/src/txn/txn_builder_masternode.ts
```diff
@@ -41,4 +48,90 @@ export class TxnBuilderMasternode extends P2WPKHTxnBuilder {
...
+   * @param {Script} changeScript to send unspent to after deducting the (converted + fees)
+   * @param {Object} [collateralInfo] needed when updating owner address
+   * @param {RawTransaction} collateralInfo.rawCollateralTx to get relevant data needed for collateral vin and vout
+   * @param {Script} collateralInfo.newOwnerScript for new owner address
+   * @return {Promise<TransactionSegWit>}
+   */
+  async update (
+    updateMasternode: UpdateMasternode,
+    changeScript: Script,
+    collateralInfo?: {
```
fuxingloh:
```suggestion
    collateral?: {
```

IMO, we could just call it `collateral` since all datatypes are analogous to `info` being passed.

packages/jellyfish-transaction-builder/src/txn/txn_builder_masternode.ts
```diff
@@ -41,4 +48,90 @@ export class TxnBuilderMasternode extends P2WPKHTxnBuilder {
...
+      rawCollateralTx: RawTransaction
```
fuxingloh: As described above, we might need to use another more prevalent data type that is available everywhere. Since `RawTransaction` might have more information that we might not need. We need to keep the scope small; we could also declare the type here.

fuxingloh (changes requested):
The test looks good; I've requested changes for the source.

packages/jellyfish-transaction-builder/src/txn/txn_builder_masternode.ts
marktanrj: ok, I replaced the raw transaction param with txid, value, tokenId

packages/jellyfish-transaction-builder/src/txn/txn_builder_masternode.ts
```diff
@@ -41,4 +47,90 @@ export class TxnBuilderMasternode extends P2WPKHTxnBuilder {
...
+   * @param {Script} collateral.newOwnerScript for new owner address
+   * @return {Promise<TransactionSegWit>}
+   */
+  async update (
+    updateMasternode: UpdateMasternode,
+    changeScript: Script,
+    collateral?: {
+      txid: string
+      value: BigNumber
+      tokenId: number
```
fuxingloh:
```suggestion
```

Let's remove this too, since (see below).

packages/jellyfish-transaction-builder/src/txn/txn_builder_masternode.ts
```diff
@@ -41,4 +47,90 @@ export class TxnBuilderMasternode extends P2WPKHTxnBuilder {
...
+          OP_CODES.OP_DEFI_TX_UPDATE_MASTER_NODE(updateMasternode)
+        ]
+      },
+      tokenId: 0x00
+    }
+
+    const change: Vout = {
+      value: total,
+      script: changeScript,
+      tokenId: 0x00
```
fuxingloh, line 88: You're already using a fixed `0x00` here, and since the beginning of `ain`, it has always been `0x00`. We don't intend to introduce any more `tokenId`. It's more or less a deprecated field.

packages/jellyfish-transaction-builder/src/txn/txn_builder_masternode.ts
```diff
@@ -41,4 +47,90 @@ export class TxnBuilderMasternode extends P2WPKHTxnBuilder {
...
+
+    if (collateral !== null && collateral !== undefined) {
+      const { txid, value, tokenId, newOwnerScript } = collateral
+
+      const collateralPrevout: Prevout = {
+        txid,
+        vout: 1,
+        script: changeScript,
+        value: new BigNumber(value),
+        tokenId
```
fuxingloh:
```suggestion
        tokenId: 0x00
```

So we can just use `0x00`.

packages/jellyfish-transaction-builder/src/txn/txn_builder_masternode.ts
```diff
@@ -41,4 +47,90 @@ export class TxnBuilderMasternode extends P2WPKHTxnBuilder {
...
+      }
+      const collateralVout: Vout = {
+        script: newOwnerScript,
+        value: new BigNumber(value),
+        tokenId
```
fuxingloh:
```suggestion
        tokenId: 0x00
```

packages/jellyfish-transaction-builder/src/txn/txn_builder_masternode.ts
```diff
@@ -41,4 +47,90 @@ export class TxnBuilderMasternode extends P2WPKHTxnBuilder {
...
+
+  /**
+   * Build update masternode transaction
+   *
+   * @param {UpdateMasternode} updateMasternode transaction to create
+   * @param {Script} changeScript to send unspent to after deducting the (converted + fees)
+   * @param {Object} [collateral] needed when updating owner address
+   * @param {string} collateral.txid collateral txid
+   * @param {BigNumber} collateral.value collateral amount
+   * @param {number} collateral.tokenId collateral tokenId
```
fuxingloh:
```suggestion
```

packages/jellyfish-transaction-builder/src/txn/txn_builder_masternode.ts
```diff
@@ -41,4 +47,90 @@ export class TxnBuilderMasternode extends P2WPKHTxnBuilder {
...
+      script: changeScript,
+      tokenId: 0x00
+    }
+
+    const mergedVin = [...vin]
+    const mergedVout = [deFiOut, change]
+    const mergedPrevouts = [...prevouts]
+
+    if (collateral !== null && collateral !== undefined) {
+      const { txid, value, tokenId, newOwnerScript } = collateral
```
fuxingloh:
```suggestion
      const { txid, value, newOwnerScript } = collateral
```

fuxingloh (changes requested):
Generally, LGTM, but how about we remove `tokenId`? Reasons below.

packages/jellyfish-transaction-builder/src/txn/txn_builder_masternode.ts
marktanrj: I see, amended.

fuxingloh (approved):
LGTM

delphk (approved):

PR #1907, opened 2022-12-06
chore(packages/jellyfish-network): Add network port configuration

helloscoopa:
We're migrating from `dockerode` to `testcontainers`
This PR adds port configurations to be used by `testcontainers`

### DIFF
--- a/packages/jellyfish-network/src/Network.ts
+++ b/packages/jellyfish-network/src/Network.ts
@@ -4,9 +4,9 @@
 export type NetworkName = Network['name']
 
 /**
-  * Network specific DeFi configuration.
-  * They can be found in DeFiCh/ain project in file chainparams.cpp, under base58Prefixes
-  */
+ * Network specific DeFi configuration.
+ * They can be found in DeFiCh/ain project in file chainparams.cpp, under base58Prefixes
+ */
 export interface Network {
   name: 'mainnet' | 'testnet' | 'regtest' | 'devnet'
   bech32: {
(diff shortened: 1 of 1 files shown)


packages/jellyfish-network/src/Network.ts
```diff
@@ -27,12 +27,18 @@ export interface Network {
   scriptHashPrefix: 0x5a | 0x80 | 0xc4
   /** For message signing. */
   messagePrefix: '\x15Defi Signed Message:\n'
+
+  /** [HYDRA] */
```
fuxingloh:
```suggestion
```

packages/jellyfish-network/src/Network.ts
```diff
@@ -4,9 +4,9 @@
 export type NetworkName = Network['name']
 
 /**
- * Network specific DeFi configuration.
- * They can be found in DeFiCh/ain project in file chainparams.cpp, under base58Prefixes
- */
+  * Network specific DeFi configuration.
+  * They can be found in DeFiCh/ain project in file chainparams.cpp, under base58Prefixes
+  */
```
fuxingloh:
```suggestion
 * Network specific DeFi configuration.
 * They can be found in DeFiCh/ain project in file chainparams.cpp, under base58Prefixes
 */
```

You need to revert these documentation indentation changes. It's not following the correct default expected indentations defined here https://jsdoc.app/about-getting-started.html

fuxingloh (changes requested):

infinia-yzl (approved):

fuxingloh (approved):

fuxingloh:
> Adds port configurations to be used by #1790

You should describe why ports need to be declared. Instead of just referencing another issue, this issue/pr should be standalone.

helloscoopa:
> You should describe why ports need to be declared. Instead of just referencing another issue, this issue/pr should be standalone.

Updated 👍🏼

PR #1908, opened 2022-12-07
fix(jellyfish-api-core): Update listGovProposalVotes arguments

shohamc1:

- Update listGovProposalVotes arguments as per https://github.com/DeFiCh/ain/pull/1635. This is required to keep compatibility with the next release.


packages/jellyfish-api-core/src/category/governance.ts
```diff
@@ -129,13 +129,25 @@ export class Governance {
...
+   * @param {number} [pagination.start] default is 0
+   * @param {boolean} [pagination.including_start] default = true
+   * @param {number} [pagination.limit] to limit number of records
    * @return {Promise<ListVotesResult[]>} Proposal vote information
    */
-  async listGovProposalVotes (proposalId: string, masternode: MasternodeType | string = MasternodeType.MINE, cycle: number = 0): Promise<ListVotesResult[]> {
-    return await this.client.call('listgovproposalvotes', [proposalId, masternode, cycle], 'number')
+  async listGovProposalVotes (
+    options: ListGovProposalVotesOptions,
+    pagination: ListGovProposalVotesPagination = {
```
kyleleow: Correct me if i am wrong, from the code in `ain` https://github.com/DeFiCh/ain/blob/90c05a00d40381bf6da1b5db05a5e6301cd95a33/src/masternodes/rpc_proposals.cpp#L604

should `pagination` object be nested inside `options`?

shohamc1: No, the [`pagination`](https://github.com/DeFiCh/ain/blob/90c05a00d40381bf6da1b5db05a5e6301cd95a33/src/masternodes/rpc_proposals.cpp#L524-L548) object is a top level argument, there is no `options` argument. 

In the RPC, all of the arguments can be passed as JSON object, which prevents having to define optional arguments. The [PR description](https://github.com/DeFiCh/ain/pull/1635) talks about this.

kyleleow: ah okay got it, thanks for explaining!

delphk (commented):
will be good to add test cases for pagination as well

packages/jellyfish-api-core/src/category/governance.ts
shohamc1: Misunderstood what you meant, we are using the JSON object here to pass all the arguments so the pagination object should be inside the options object. Fixed this in https://github.com/JellyfishSDK/jellyfish/pull/1908/commits/028f92c56498373a8a85f5b53fe607191a202711.

Thanks for bringing this up🫡

packages/jellyfish-api-core/src/category/governance.ts
```diff
@@ -129,13 +129,20 @@ export class Governance {
...
-   * @param {string} proposalId Proposal id
-   * @param {MasternodeType | string} [masternode=MasternodeType.MINE] masternode id or reserved words 'mine' to list votes for all owned accounts or 'all' to list all votes
-   * @param {number} [cycle=0] cycle: 0 (show current), cycle: N (show cycle N), cycle: -1 (show all)
+   * @param {ListGovProposalVotesOptions} options
+   * @param {string} options.proposalId Proposal id
+   * @param {MasternodeType | string} [options.masternode=MasternodeType.MINE] masternode id or reserved words 'mine' to list votes for all owned accounts or 'all' to list all votes
+   * @param {number} [options.cycle=-1] cycle: 0 (show current), cycle: N (show cycle N), cycle: -1 (show all)
+   * @param {ListGovProposalVotesPagination} [options.pagination]
+   * @param {number} [options.pagination.start] default is 0
+   * @param {boolean} [options.pagination.including_start] default = true
```
delphk: from [https://github.com/DeFiCh/ain/blob/master/src/masternodes/rpc_proposals.cpp#L529-L544](https://github.com/DeFiCh/ain/blob/master/src/masternodes/rpc_proposals.cpp#L529-L544) seems like the default for `cycle` is 0 and `including_start` is false

packages/jellyfish-api-core/__tests__/category/governance/listGovProposalVotes.test.ts
```diff
@@ -67,6 +67,52 @@ describe('Governance', () => {
...
+      }
+    })
+    const votes2 = await testing.rpc.governance.listGovProposalVotes({
+      proposalId: proposalId,
+      pagination: {
+        start: 2,
+        including_start: true,
+        limit: 2
+      }
+    })
```
delphk, line 110: apart from this test, would be also good to test the different combinations of pagination options (eg. `start` is omitted, `including_start` is omitted or false, `limit` is omitted or 0)

packages/jellyfish-api-core/src/category/governance.ts
fuxingloh:
```suggestion
   * @param {number} [options.pagination.start = 0]
   * @param {boolean} [options.pagination.including_start = false]
```

You can specify defaults in JSDoc.

fuxingloh (changes requested):

packages/jellyfish-api-core/src/category/governance.ts
shohamc1: `including_start` defaults to `false` when `start` is provided, otherwise it is `true`. Not too sure on what to document the default as in this case.

`cycle` does default to 0, will update this 🙏

shohamc1: Fixed: https://github.com/JellyfishSDK/jellyfish/pull/1908/commits/09c483e9cd769cbf47fbce486d20bca448c6bf91

packages/jellyfish-api-core/__tests__/category/governance/listGovProposalVotes.test.ts
shohamc1: Fixed: https://github.com/JellyfishSDK/jellyfish/pull/1908/commits/09c483e9cd769cbf47fbce486d20bca448c6bf91

packages/jellyfish-api-core/src/category/governance.ts
shohamc1: Fixed: https://github.com/JellyfishSDK/jellyfish/pull/1908/commits/09c483e9cd769cbf47fbce486d20bca448c6bf91

delphk (dismissed):

packages/jellyfish-api-core/__tests__/category/governance/listGovProposalVotes.test.ts
fuxingloh: The test should be in separated `it()`s describing the pagination condition.

fuxingloh (changes requested):

fuxingloh (approved):

PR #1933, opened 2022-12-16
chore(packages/jellyfish-transaction-builder): Add more tests to createcfp

### LINKED ISSUE #1930: Include additional tests for createCfp in `txn_builder_governance.ts`

Create a test scenario where 

https://github.com/JellyfishSDK/jellyfish/blob/7ac010bd9bc7c8f3a119ebfaab94d7bb93d5aed6/packages/jellyfish-transaction-builder/src/txn/txn_builder_governance.ts#L20

The condition above is triggered when the amount is greater than `10`.

Currently, only one scenario has been added/tested.

https://github.com/JellyfishSDK/jellyfish/blob/7ac010bd9bc7c8f3a119ebfaab94d7bb93d5aed6/packages/jellyfish-transaction-builder/__tests__/txn/txn_builder_governance_createcfp.test.ts#L43

We need to test when it's greater than 10, less than 10, and when it's 10. Make sure the balance is deducted properly.

/triage accepted
/priority important-soon
/assign @infinia-yzl for **Hydra**
/area packages

helloscoopa:

To check all the cases of proposal amounts and balances.

Fixes #1930

### DIFF
--- a/packages/jellyfish-transaction-builder/__tests__/txn/txn_builder_governance_createcfp.test.ts
+++ b/packages/jellyfish-transaction-builder/__tests__/txn/txn_builder_governance_createcfp.test.ts
@@ -14,6 +14,7 @@ describe('createCfp', () => {
   let providers: MockProviders
   let builder: P2WPKHTransactionBuilder
   const testing = Testing.create(new MasterNodeRegTestContainer())
+  const fundAmount = 12000100
 
   beforeAll(async () => {
     await testing.container.start()
@@ -25,7 +26,7 @@ describe('createCfp', () => {
     builder = new P2WPKHTransactionBuilder(providers.fee, providers.prevout, providers.elliptic, RegTest)
 
     await testing.container.waitForWalletBalanceGTE(11)
-    await fundEllipticPair(testing.container, providers.ellipticPair, 12000100)
+    await fundEllipticPair(testing.container, providers.ellipticPair, fundAmount)
     await providers.setupMocks()
   })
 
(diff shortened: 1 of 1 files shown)


packages/jellyfish-transaction-builder/__tests__/txn/txn_builder_governance_createcfp.test.ts
```diff
@@ -34,49 +35,76 @@ describe('createCfp', () => {
   })
 
   it('should createCfp', async () => {
```
fuxingloh: Let's create separates `it()`s or use `it.each([1,2,3])`?

infinia-yzl (changes requested):
Try to cover these as well:
- Test with a large value for creationFee
- Test with a negative value for creationFee

packages/jellyfish-transaction-builder/__tests__/txn/txn_builder_governance_createcfp.test.ts
```diff
@@ -25,22 +26,34 @@ describe('createCfp', () => {
...
     await providers.setupMocks()
   })
 
   afterAll(async () => {
     await testing.container.stop()
   })
 
-  it('should createCfp', async () => {
+  it.each([{
+    nAmount: 999,
```
delphk: suggestion: instead of using `nAmount` of `999`, `1000`, `1001`, i would test with larger variances (eg. `10`, `1000`, `100000`), especially since there is a change range, using test amounts of small variances might not be so obvious that it is working as intended (as the difference may fall within the change range)

helloscoopa: Fixed [`7e7f304`](https://github.com/JellyfishSDK/jellyfish/pull/1933/commits/7e7f304a501060667518639367c9b42e1407efc5)

packages/jellyfish-transaction-builder/__tests__/txn/txn_builder_governance_createcfp.test.ts
```diff
@@ -25,22 +26,34 @@ describe('createCfp', () => {
...
+    creationFee: 10,
+    changeRange: [fundAmount - 10.001, fundAmount - 10]
+  }, {
+    nAmount: 1000,
+    creationFee: 10,
+    changeRange: [fundAmount - 20.001, fundAmount - 20]
+  }, {
+    nAmount: 100000,
+    creationFee: 1000,
+    changeRange: [fundAmount - 1020.001, fundAmount - 30.01]
```
delphk:
```suggestion
    changeRange: [fundAmount - 1020.001, fundAmount - 1020]
```

delphk (approved):

infinia-yzl (approved):
:shipit:

marktanrj (approved):

fuxingloh (approved):

PR #1934, opened 2022-12-16
feat(jellyfish-api-core): add signMessage RPC

shancht:
**What this PR does / why we need it:** 
/kind feature

**Which issue(s) does this PR fixes?:** 
Adding `signMessage()` from [Issue#48](https://github.com/JellyfishSDK/jellyfish/issues/48)

This features allow users to sign an arbitrary message using the private key of an address. 
It requires wallet passphrase to be set with `walletpassphrase` call if the wallet is encrypted.

**Additional comments?:**
The primary purpose of `signMessage()` is to present a signature to showcase that the funds are in the control of the private key holder. The holder may also generate a signature with custom prefixes to showcase proof of authorisation.

Usage Example: [Here](https://bitcoin.stackexchange.com/a/3339/160)

### DIFF
--- a/packages/jellyfish-api-core/__tests__/category/wallet/signMessage.test.ts
+++ b/packages/jellyfish-api-core/__tests__/category/wallet/signMessage.test.ts
@@ -1,0 +1,41 @@
+import { MasterNodeRegTestContainer } from '@defichain/testcontainers'
+import { ContainerAdapterClient } from '../../container_adapter_client'
+import { RpcApiError } from '@defichain/jellyfish-api-core'
+
+describe('Sign Message', () => {
+  const container = new MasterNodeRegTestContainer()
+  const client = new ContainerAdapterClient(container)
+
+  beforeAll(async () => {
+    await container.start()
+  })
+
+  afterAll(async () => {
+    await container.stop()
+  })
+
+  it('should throw error if private key is not available', async () => {
+    const promise = client.wallet.signMessage('mpLQjfK79b7CCV4VMJWEWAj5Mpx8Up5zxB', 'This is a test message')
+
+    await expect(promise).rejects.toThrow(RpcApiError)
+    await expect(promise).rejects.toMatchObject({
+      payload: {
+        code: -4,
+        message: 'Private key not available',
+        method: 'signmessage'
+      }
+    })
+  })
+
+  it('should throw error if address is invalid', async () => {
+    const promise = client.wallet.signMessage('test', 'This is a test message')
+    await expect(promise).rejects.toThrow(RpcApiError)
+    await expect(promise).rejects.toMatchObject({
+      payload: {
+        code: -3,
+        message: 'Invalid address',
+        method: 'signmessage'
+      }
+    })
+  })
+})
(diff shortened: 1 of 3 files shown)


shancht:
Hi, the DefiChain codebase (/rpcwallet.cpp) included a section to ensure that the wallet is not encrypted/locked. I'd like to seek clarification on how I could best write a syntax for this section in Jellyfish please:

```
    auto locked_chain = pwallet->chain().lock();
    LOCK2(pwallet->cs_wallet, locked_chain->mutex());

    EnsureWalletIsUnlocked(pwallet);
```

Thank you!

shohamc1:
> Hi, the DefiChain codebase (/rpcwallet.cpp) included a section to ensure that the wallet is not encrypted/locked. I'd like to seek clarification on how I could best write a syntax for this section in Jellyfish please:
> 
> ```
>     auto locked_chain = pwallet->chain().lock();
>     LOCK2(pwallet->cs_wallet, locked_chain->mutex());
> 
>     EnsureWalletIsUnlocked(pwallet);
> ```
> 
> Thank you!

Not sure if this is a useful test, but you can set a wallet passphrase using `encryptwallet` and then use `walletlock` to lock the wallet. You can unlock the wallet later using `walletpassphrase`.

packages/jellyfish-api-core/__tests__/category/wallet/signMessage.test.ts
```diff
@@ -0,0 +1,41 @@
...
+  beforeAll(async () => {
+    await container.start()
+  })
+
+  afterAll(async () => {
+    await container.stop()
+  })
+
+  it('should throw error if private key is not available', async () => {
+    const promise = client.wallet.signMessage('mpLQjfK79b7CCV4VMJWEWAj5Mpx8Up5zxB', 'This is a test message')
```
fuxingloh:
```suggestion
    const address = await client.wallet.getNewAddress()
    const promise = client.wallet.signMessage(address, 'This is a test message')
```

This is how you create a new address within the Wallet and then use it for signing.

fuxingloh (dismissed):

fuxingloh:
LGTM,

@shohamc1 can you take a look at this and give it a review? Thanks!

packages/jellyfish-api-core/src/category/wallet.ts
```diff
@@ -324,6 +324,18 @@ export class Wallet {
   async listWallets (): Promise<string[]> {
     return await this.client.call('listwallets', [], 'number')
   }
+
+  /**
+   * Sign a message with the private key of an address
+   * Requires wallet passphrase to be set with walletpassphrase call if wallet is encrypted.
```
shohamc1:
```suggestion
   * Requires wallet to be unlocked for usage. Use `walletpassphrase` to unlock wallet.
```

shohamc1 (changes requested):
Looks good minus this (the password is set using `encryptwallet`).

Edit: I would also test signing using an address not from the wallet. It should fail with `Private key not available`.

https://github.com/DeFiCh/ain/blob/0beb92ed218f335377d8591d33f66dba85ff94a7/src/wallet/rpcwallet.cpp#L577

shohamc1 (dismissed):

fuxingloh (approved):

shohamc1 (approved):

PR #1959, opened 2023-01-09
chore(playground-api): generate blocks using random masternode addresses

fullstackninja864:
Generate blocks using random masternode addresses

Fixes #


apps/playground-api/src/PlaygroundBlock.ts
```diff
@@ -12,7 +12,9 @@ export class PlaygroundBlock {
 
   @Interval(3000)
   async generate (): Promise<void> {
-    await this.client.call('generatetoaddress', [1, PlaygroundSetup.address, 1], 'number')
+    const { masternodes }: MiningInfo = await this.client.mining.getMiningInfo()
+    const randomAddress = masternodes[Math.floor(Math.random() * masternodes.length)].operator
+    await this.client.call('generatetoaddress', [1, randomAddress, 1], 'number')
```
fuxingloh: You need to add tests to assert this is valid.

fullstackninja864: added testcase

apps/playground-api/src/PlaygroundBlock.ts
```diff
@@ -1,7 +1,7 @@
 import { Interval } from '@nestjs/schedule'
 import { Injectable, Logger } from '@nestjs/common'
 import { ApiClient } from '@defichain/jellyfish-api-core'
-import { PlaygroundSetup } from './setups/setup'
+import { MiningInfo } from '@defichain/jellyfish-api-core/dist/category/mining'
```
fuxingloh:
```suggestion
import { mining } from '@defichain/jellyfish-api-core'
```

apps/playground-api/src/PlaygroundBlock.ts
```diff
@@ -12,7 +12,9 @@ export class PlaygroundBlock {
 
   @Interval(3000)
   async generate (): Promise<void> {
-    await this.client.call('generatetoaddress', [1, PlaygroundSetup.address, 1], 'number')
+    const { masternodes }: MiningInfo = await this.client.mining.getMiningInfo()
```
fuxingloh:
```suggestion
    const { masternodes }: mining.MiningInfo = await this.client.mining.getMiningInfo()
```

Could we do this instead?

apps/playground-api/src/PlaygroundBlock.ts
fuxingloh:
```suggestion
    const randomOperator = masternodes[Math.floor(Math.random() * masternodes.length)].operator
    await this.client.call('generatetoaddress', [1, randomOperator, 1], 'number')
```

apps/playground-api/__tests__/PlaygroundBlock.test.ts
```diff
@@ -0,0 +1,55 @@
+import { PlaygroundApiTesting } from '../testing/PlaygroundApiTesting'
+import waitForExpect from 'wait-for-expect'
+import { PlaygroundProbeIndicator } from '../src/PlaygroundIndicator'
+import { PlaygroundBlock } from '../src/PlaygroundBlock'
+import { RegTestFoundationKeys } from '@defichain/jellyfish-network'
+import { StartOptions, MasterNodeRegTestContainer } from '@defichain/testcontainers'
+import { TestingGroup } from '@defichain/jellyfish-testing'
+import { MiningInfo } from '@defichain/jellyfish-api-core/dist/category/mining'
```
fuxingloh: Same for this, to prevent test/build race condition.

apps/playground-api/__tests__/PlaygroundBlock.test.ts
```diff
@@ -0,0 +1,55 @@
...
+})
+
+describe('playgroundBlock', () => {
+  it('should be able to mint blocks using multiple nodes', async () => {
+    const count = await testing.container.getBlockCount()
+    const { masternodes }: MiningInfo = await testing.container.call('getmininginfo', [])
+    const playgroundBlock = testing.app.get(PlaygroundBlock)
+    for (let i = 0; i < 100; i++) {
+      await playgroundBlock.generate()
+      await testing.container.generate(1)
```
fuxingloh:
```suggestion
```

This seems redundant? I assume playgroundBlock already generated it.

apps/playground-api/__tests__/PlaygroundBlock.test.ts
```diff
@@ -0,0 +1,55 @@
...
+    }
+    const updatedCount = await testing.container.getBlockCount()
+    const { masternodes: updatedMasterNodes }: MiningInfo = await testing.container.call('getmininginfo', [])
+    expect(updatedCount).toBeGreaterThan(count)
+    updatedMasterNodes.forEach((eachNode) => {
+      const currentNode = masternodes.find((each) => each.operator === eachNode.operator)
+      expect(eachNode.mintedblocks).toBeGreaterThan(currentNode?.mintedblocks ?? 0)
+    })
```
fuxingloh: The assert condition has a very high probability of failure cause we change the assert condition to as long as 2 unique masternode minted?

fuxingloh (changes requested):

apps/playground-api/__tests__/PlaygroundBlock.test.ts
```diff
@@ -0,0 +1,66 @@
...
+    const { masternodes: updatedMasterNodes }: mining.MiningInfo = await testing.container.call('getmininginfo', [])
+    expect(updatedBlockCount).toBeGreaterThan(blockCount)
+    const minerMNcount = updatedMasterNodes.reduce((minerMNcount, eachNode) => {
+      const currentMintedBlocks = masternodes.find((each) => each.operator === eachNode.operator)?.mintedblocks ?? 0
+      if (eachNode.mintedblocks > currentMintedBlocks) {
+        return ++minerMNcount
+      }
+      return minerMNcount
+    }, 0)
+    expect(minerMNcount).toBeGreaterThanOrEqual(2)
```
pierregee: Can add a doc/comment on why it's expecting `>=2`

apps/playground-api/__tests__/PlaygroundBlock.test.ts
```diff
@@ -0,0 +1,68 @@
...
+afterAll(async () => {
+  await testing.stop()
+})
+
+describe('playgroundBlock', () => {
+  it('should be able to mint blocks using multiple nodes', async () => {
+    const blockCount = await testing.container.getBlockCount()
+    const { masternodes }: mining.MiningInfo = await testing.container.call('getmininginfo', [])
+    const playgroundBlock = testing.app.get(PlaygroundBlock)
+    for (let i = 0; i < 20; i++) {
```
kyleleow, line 51: may be good to add a comment that why we chose 20 and not any other number

apps/playground-api/__tests__/PlaygroundBlock.test.ts
```diff
@@ -0,0 +1,68 @@
...
+import { RegTestFoundationKeys } from '@defichain/jellyfish-network'
+import { StartOptions, MasterNodeRegTestContainer } from '@defichain/testcontainers'
+import { TestingGroup } from '@defichain/jellyfish-testing'
+import { mining } from '@defichain/jellyfish-api-core'
+
+class PlaygroundApiMasterNodeRegTestContainer extends MasterNodeRegTestContainer {
+  protected getCmd (opts: StartOptions): string[] {
+    return [
+      ...super.getCmd(opts),
+      `-masternode_operator=${RegTestFoundationKeys[1].operator.address}`,
```
kyleleow, line 15: may be good to add a comment that why we add 7 MN operators here (helps newbie like me to understand)

apps/playground-api/src/PlaygroundBlock.ts
```diff
@@ -12,7 +10,9 @@ export class PlaygroundBlock {
 
   @Interval(3000)
   async generate (): Promise<void> {
-    await this.client.call('generatetoaddress', [1, PlaygroundSetup.address, 1], 'number')
+    const { masternodes }: mining.MiningInfo = await this.client.mining.getMiningInfo()
+    const randomOperator = masternodes[Math.floor(Math.random() * masternodes.length)].operator
```
kyleleow: i am not expert in this, but what if we mint the block sequentially, maybe just 4 in total, not all to save time? in that way we can remove the randomness and be sure which MN to use for testing, need your thought on this @fuxingloh .

Harsh's thought is to keep this randomness to simulate the actual blockchain behaviour

fuxingloh: Yeah, you're right. I would say sequential is better since you can anticipate them for tests.

I am averse to Harsh idea since if we want them to be random to simulate actual blockchain behavior, we should just use TestNet, or rather just use MainNet all together.

apps/playground-api/src/PlaygroundBlock.ts
```diff
@@ -10,10 +9,16 @@ export class PlaygroundBlock {
...
+    const operator = masternodes[blocks % masternodes.length].operator
+    await this.client.call('generatetoaddress', [1, operator, 1], 'number')
     const count = await this.client.blockchain.getBlockCount()
     this.logger.log(`generated new block - height: ${count}`)
   }
+
+  @Interval(3000)
+  async generate (): Promise<void> {
+    await this.generateToAddress()
+  }
```
fuxingloh:
```suggestion
  @Interval(3000)
  async generate (): Promise<void> {
    const { masternodes, blocks }: mining.MiningInfo = await this.client.mining.getMiningInfo()
    const operator = masternodes[blocks % masternodes.length].operator
    await this.client.call('generatetoaddress', [1, operator, 1], 'number')
    const count = await this.client.blockchain.getBlockCount()
    this.logger.log(`generated new block - height: ${count}`)
  }
```

Given that `generateToAddress` and `generate` are identical, we don't need another method? 

The decorator is a behavior applied to that method, so IMO should not warrant a different name either.

fuxingloh (changes requested):

apps/playground-api/__tests__/PlaygroundBlock.test.ts
```diff
@@ -0,0 +1,64 @@
...
+      await playgroundBlock.generate()
+      await wait(3000)
+    }
+    const updatedBlockCount = await testing.container.getBlockCount()
+    const { masternodes: updatedMasterNodes }: mining.MiningInfo = await testing.container.call('getmininginfo', [])
+    expect(updatedBlockCount).toBeGreaterThan(blockCount)
+    updatedMasterNodes.forEach((eachNode) => {
+      const currentMintedBlocks = masternodes.find((each) => each.operator === eachNode.operator)?.mintedblocks ?? 0
+      expect(eachNode.mintedblocks).toBeGreaterThanOrEqual(currentMintedBlocks)
+    })
```
fuxingloh: This test condition seems overly complex; would it be easier just to check each masternode has minted?

fuxingloh (changes requested):

apps/playground-api/__tests__/PlaygroundBlock.test.ts
fullstackninja864: Updated the logic just to verify that each masternode contributed to minting.

apps/playground-api/__tests__/PlaygroundBlock.test.ts
```diff
@@ -0,0 +1,63 @@
...
+    // generate next 20 block to confirm multiple masternodes are mining the blocks
+    for (let i = 0; i < 20; i++) {
+      await playgroundBlock.generate()
+      await wait(3000)
+    }
+    const updatedBlockCount = await testing.container.getBlockCount()
+    expect(updatedBlockCount).toBeGreaterThan(blockCount)
+    const { masternodes }: mining.MiningInfo = await testing.container.call('getmininginfo', [])
+    // check each masternode minted blocks
+    masternodes.forEach((eachNode) => {
```
fuxingloh, line 60: Assert `masternode >= 7`, other wise expect masternodes for loop is flawed

fuxingloh (changes requested):

apps/playground-api/__tests__/PlaygroundBlock.test.ts
fullstackninja864: Making count to strict equal to 8

fuxingloh (approved):

PR #1972, opened 2023-01-13
feat(apps/whale-api): add new endpoint to expose mint limits and minted amount per member

### LINKED ISSUE #1961: Add consortium member's mint limit and daily mint limit, minted token and daily minted token attributes
Need to add the following attributes: mint limit and daily mint limit, minted token and daily minted token. https://github.com/DeFiCh/ain/blob/90c05a00d40381bf6da1b5db05a5e6301cd95a33/test/functional/feature_consortium.py#L392

#### Why is this needed:
We need to show the following on the Bridge UI:
- To show the daily limit + daily minted
- To show an error message if the daily/member limit has been reached

lykalabrada:
As described on the linked issue below

Fixes https://github.com/JellyfishSDK/jellyfish/issues/1961


pierregee (changes requested):
We still need the global limit somewhere in the response. One scenario that could happen is that there's a global limit but no member limit

pierregee:
Need to cover [tests](https://github.com/JellyfishSDK/jellyfish/blob/main/packages/whale-api-client/__tests__/api/consortium.test.ts) on whale-api-client for this new endpoint too

lykalabrada:
> We still need the global limit somewhere in the response. One scenario that could happen is that there's a global limit but no member limit

Added global mint stats. Also added the test on the `whale-api-client` side

pierregee (dismissed):

apps/whale-api/src/module.api/consortium.controller.spec.ts
```diff
@@ -3,125 +3,126 @@ import { TestingGroup } from '@defichain/jellyfish-testing'
...
+
+async function setGovAttr (attributes: object): Promise<void> {
+  const hash = await alice.rpc.masternode.setGov({ ATTRIBUTES: attributes })
+  expect(hash).toBeTruthy()
+  await alice.generate(1)
+  await tGroup.waitForSync()
+}
+
+async function setMemberInfo (tokenId: string, memberInfo: Array<{ id: string, name: string, backingId: string, ownerAddress: string, mintLimit: string, dailyMintLimit: string }>): Promise<void> {
+  const members: any = {}
```
pierregee: all  good, just one nitpick , this is better? 
```suggestion
const members: { [key: string]: { [key: string]: string } } = {}
```

pierregee (dismissed):

apps/whale-api/src/module.api/consortium.controller.ts
```diff
@@ -23,4 +23,18 @@ export class ConsortiumController {
       ttl: 600 // 10 minutes
     }) as AssetBreakdownInfo[]
   }
+
+  /**
+   *  Gets the mint stats information of a specific consortium member
+   *
+   * @return {Promise<MemberMintStatsInfo>}
+    */
+  @Get('/mintstats/:memberid')
```
thedoublejay: Would this be better to just be stats? So it can contain mint and burn stats in the future?

thedoublejay (changes requested):

apps/whale-api/src/module.api/consortium.controller.ts
lykalabrada: Updated!

thedoublejay (approved):

fuxingloh (approved):

PR #1980, opened 2023-01-18
feat(jellyfish-api-core): add submitBlock RPC

DocteurPing:

This PR add the function submitBlock to Jellyfish

Fixes #

### DIFF
--- a/packages/jellyfish-api-core/__tests__/category/mining/submitblock.test.ts
+++ b/packages/jellyfish-api-core/__tests__/category/mining/submitblock.test.ts
@@ -1,11 +1,96 @@
 import { MasterNodeRegTestContainer } from '@defichain/testcontainers/dist/index'
 import { ContainerAdapterClient } from '../../container_adapter_client'
-import { Testing } from '@defichain/jellyfish-testing'
+import { BlockHeader } from '../../../../jellyfish-block/src/BlockHeader'
+import { CBlock, Block } from '../../../../jellyfish-block/src/Block'
+import BigNumber from 'bignumber.js'
+import { SmartBuffer } from 'smart-buffer'
+import { OP_CODES, OP_PUSHDATA, TransactionSegWit } from '@defichain/jellyfish-transaction'
 
 describe('submit block', () => {
   const container = new MasterNodeRegTestContainer()
   const client = new ContainerAdapterClient(container)
-  const testing = Testing.create(container)
+
+  const blockHeader: BlockHeader = {
+    version: 536870912,
+    hashPrevBlock: 'a1cbb25b88b0a45060c5c635782ac5a562bed53b0f1d68f8eab21145ebda7566',
+    hashMerkleRoot: 'b023099453883ccb6f27e9363e4c14a0237e49c6f8e285a45fdcdebb9d565c72',
+    time: 1630596477,
+    bits: 0x1839d282,
+    stakeModifier: 'd2ee35e495f98d2657c8a1ea7b692735edc993e944cb5c6baf025ed69adc5dda',
+    height: new BigNumber(1155173),
+    mintedBlocks: new BigNumber(57),
+    signature: '1f8e16e55310c7dc0f3203868134510552a34e77e140aff85ccef76f410ba2c1033cc3019f16c51c489d7b430d635e4deb1d4cf384c2276d8485a4ad7b3784ca39'
+  }
+
+  const firstTransaction: TransactionSegWit = {
+    version: 0x4,
+    marker: 0x00,
+    flag: 0x01,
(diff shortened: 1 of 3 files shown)


packages/jellyfish-api-core/__tests__/category/mining/submitblock.test.ts
```diff
@@ -0,0 +1,35 @@
...
+  })
+
+  it('should submit a block', async () => {
+    const blockHash = await testing.misc.waitForBlockHash(1)
+    const hash = await testing.rpc.blockchain.getBlock(blockHash, 0)
+    const promise = client.mining.submitBlock(hash)
+    await expect(promise).resolves.not.toThrow()
+  })
+
+  it('should throw an error if the block is not valid', async () => {
```
fuxingloh, line 128: IMO this test is way too simple and limited in scope. It should have scenarios like submitting a static block and various conditions and versions that should cover regression against changes in block design.

fuxingloh (changes requested):

packages/jellyfish-api-core/src/category/mining.ts
```diff
@@ -44,6 +44,18 @@ export class Mining {
   async estimateSmartFee (confirmationTarget: number, estimateMode: EstimateMode = EstimateMode.CONSERVATIVE): Promise<SmartFeeEstimation> {
     return await this.client.call('estimatesmartfee', [confirmationTarget, estimateMode], 'number')
   }
+
+  /**
+   * Attempts to submit new block to network.
+   * See https://en.bitcoin.it/wiki/BIP_0022 for full specification.
+   *
+   * @param {string} hexdata the hex-encoded block data to submit
+   * @param {string} dummy dummy value, for compatibility with BIP22. This value is ignored.
```
fuxingloh:
```suggestion
   * @param {string} dummy value, for compatibility with BIP22. This value is ignored.
```

packages/jellyfish-api-core/__tests__/category/mining/submitblock.test.ts
```diff
@@ -0,0 +1,132 @@
...
+    await expect(promise).resolves.not.toThrow()
+  })
+
+  it('should submit a block with different version', async () => {
+    block.blockHeader.version = 2
+    const composable = new CBlock(block)
+    const buffer = new SmartBuffer()
+    composable.toBuffer(buffer)
+    const data = buffer.toBuffer().toString('hex')
+    const promise = client.mining.submitBlock(data)
```
fuxingloh:
```suggestion
    const hexdata = buffer.toBuffer().toString('hex')
    const promise = client.mining.submitBlock(hexdata)
```

packages/jellyfish-api-core/__tests__/category/mining/submitblock.test.ts
```diff
@@ -0,0 +1,132 @@
...
+    await expect(promise).resolves.not.toThrow()
+  })
+
+  it('should submit a block with different version', async () => {
+    block.blockHeader.version = 2
```
fuxingloh: You're causing side effect on const, it's polluted. Do this instead:

```suggestion
    const v2Block = {
      ...block,
      blockHeader: {
        ...block.blockHeader,
        version: 2
      }
    }
```

Or you should reconstruct the block each time you want to use it.

packages/jellyfish-api-core/__tests__/category/mining/submitblock.test.ts
```diff
@@ -0,0 +1,132 @@
...
+  afterAll(async () => {
+    await container.stop()
+  })
+
+  it('should submit a block', async () => {
+    const composable = new CBlock(block)
+    const buffer = new SmartBuffer()
+    composable.toBuffer(buffer)
+    const data = buffer.toBuffer().toString('hex')
+    const promise = client.mining.submitBlock(data)
```
fuxingloh:
```suggestion
    const hexdata= buffer.toBuffer().toString('hex')
    const promise = client.mining.submitBlock(hexdata)
```

to match the parameter you used.

fuxingloh (changes requested):

packages/jellyfish-api-core/src/category/mining.ts
DocteurPing: the first dummy is the name of the variable should i still remove it ?

fuxingloh: preferably the variable should be part of the statement.

E.g:
- `{string} dummy value for something`
- `{string} hexdata is the hex-encoded`

Otherwise separate by -:

- `{string} dummy - A dummy value for something`

PR #2001, opened 2023-01-24
fix(whale-api-client): fixed listProposalVotes pagination api

fullstackninja864:
In this pr we have fixed pagination issue in listProposalVotes listing
Previously to calculate next cursor we were using 

```
return ApiPagedResponse.of(votes, size, () => {
  return (votes.length - 1).toString()
})
```

Considering above logic if we have 20 voting and we fetch 10 voting list per api request then for 1st api request we will get next as 9 and in 2nd api request also we will get next as 9. this ends up with having unlimited pagination as it will display 11-20th voting list for all upcoming api response.
To fix this unlimited pagination we have updated the logic to consider previous next value to give expected result

```
return ApiPagedResponse.of(votes, size, () => {
  if (next === undefined) {
    return (votes.length - 1).toString()
  }
  return (next + votes.length).toString()
})
```

Fixes #

### DIFF
--- a/packages/whale-api-client/__tests__/api/governance.test.ts
+++ b/packages/whale-api-client/__tests__/api/governance.test.ts
@@ -326,10 +326,12 @@ describe('governance - listProposalVotes', () => {
     const result1 = await client.governance.listGovProposalVotes({ id: cfpProposalId, size: 2, masternode: MasternodeType.ALL, cycle: -1 })
     expect(result1.length).toStrictEqual(2)
     expect(result1.hasNext).toStrictEqual(true)
+    expect(result1.nextToken).toStrictEqual('1')
 
     const result2 = await client.governance.listGovProposalVotes({ id: cfpProposalId, size: 2, next: result1.nextToken, masternode: MasternodeType.ALL, cycle: -1 })
     expect(result2.length).toStrictEqual(2)
     expect(result2.hasNext).toStrictEqual(true)
+    expect(result2.nextToken).toStrictEqual('3')
 
     const result3 = await client.governance.listGovProposalVotes({ id: cfpProposalId, size: 2, next: result2.nextToken, masternode: MasternodeType.ALL, cycle: -1 })
     expect(result3.length).toStrictEqual(0)
--- a/packages/whale-api-client/__tests__/api/governance.test.ts
+++ b/packages/whale-api-client/__tests__/api/governance.test.ts
@@ -326,11 +326,11 @@ describe('governance - listProposalVotes', () => {
     const result1 = await client.governance.listGovProposalVotes({ id: cfpProposalId, size: 2, masternode: MasternodeType.ALL, cycle: -1 })
     expect(result1.length).toStrictEqual(2)
     expect(result1.hasNext).toStrictEqual(true)
-    
+
     const result2 = await client.governance.listGovProposalVotes({ id: cfpProposalId, size: 2, next: result1.nextToken, masternode: MasternodeType.ALL, cycle: -1 })
     expect(result2.length).toStrictEqual(2)
     expect(result2.hasNext).toStrictEqual(true)
-    
+
     const result3 = await client.governance.listGovProposalVotes({ id: cfpProposalId, size: 2, next: result2.nextToken, masternode: MasternodeType.ALL, cycle: -1 })
     expect(result3.length).toStrictEqual(0)
     expect(result3.hasNext).toStrictEqual(false)
--- a/packages/whale-api-client/__tests__/api/governance.test.ts
+++ b/packages/whale-api-client/__tests__/api/governance.test.ts
@@ -322,6 +322,19 @@ describe('governance - listProposalVotes', () => {
     expect(result.length).toStrictEqual(3)
   })
 
+  it('should listProposalVotes with pagination', async () => {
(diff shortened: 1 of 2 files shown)


packages/whale-api-client/__tests__/api/governance.test.ts
```diff
@@ -322,6 +322,19 @@ describe('governance - listProposalVotes', () => {
...
+    const result = await client.governance.listGovProposalVotes({ id: cfpProposalId, size: 2, masternode: MasternodeType.ALL, cycle: -1 })
+    expect(result.length).toStrictEqual(2)
+    expect(result.hasNext).toStrictEqual(true)
+    const result1 = await client.governance.listGovProposalVotes({ id: cfpProposalId, size: 2, next: result.nextToken, masternode: MasternodeType.ALL, cycle: -1 })
+    expect(result1.length).toStrictEqual(2)
+    expect(result1.hasNext).toStrictEqual(true)
+    const result2 = await client.governance.listGovProposalVotes({ id: cfpProposalId, size: 2, next: result1.nextToken, masternode: MasternodeType.ALL, cycle: -1 })
+    expect(result2.length).toStrictEqual(0)
+    expect(result2.hasNext).toStrictEqual(false)
+    expect(result2.nextToken).toStrictEqual(undefined)
```
eli-lim:
```suggestion
    const result1 = await client.governance.listGovProposalVotes({ id: cfpProposalId, size: 2, masternode: MasternodeType.ALL, cycle: -1 })
    expect(result1.length).toStrictEqual(2)
    expect(result1.hasNext).toStrictEqual(true)
    
    const result2 = await client.governance.listGovProposalVotes({ id: cfpProposalId, size: 2, next: result1.nextToken, masternode: MasternodeType.ALL, cycle: -1 })
    expect(result2.length).toStrictEqual(2)
    expect(result2.hasNext).toStrictEqual(true)
    
    const result3 = await client.governance.listGovProposalVotes({ id: cfpProposalId, size: 2, next: result2.nextToken, masternode: MasternodeType.ALL, cycle: -1 })
    expect(result3.length).toStrictEqual(0)
    expect(result3.hasNext).toStrictEqual(false)
    expect(result3.nextToken).toStrictEqual(undefined)
```

packages/whale-api-client/__tests__/api/governance.test.ts
```diff
@@ -322,6 +322,19 @@ describe('governance - listProposalVotes', () => {
...
 
+  it('should listProposalVotes with pagination', async () => {
+    const result = await client.governance.listGovProposalVotes({ id: cfpProposalId, size: 2, masternode: MasternodeType.ALL, cycle: -1 })
+    expect(result.length).toStrictEqual(2)
+    expect(result.hasNext).toStrictEqual(true)
+    const result1 = await client.governance.listGovProposalVotes({ id: cfpProposalId, size: 2, next: result.nextToken, masternode: MasternodeType.ALL, cycle: -1 })
+    expect(result1.length).toStrictEqual(2)
+    expect(result1.hasNext).toStrictEqual(true)
+    const result2 = await client.governance.listGovProposalVotes({ id: cfpProposalId, size: 2, next: result1.nextToken, masternode: MasternodeType.ALL, cycle: -1 })
+    expect(result2.length).toStrictEqual(0)
```
eli-lim: Seems odd that `result1.hasNext` is `true`, but calling the API again with `result1.nextToken` gives a list of length 0 🤔 

Might be a natural limitation of how the cursor pagination is currently implemented?
Maybe we want to take a look at how `hasNext` is computed

kyleleow: This is indeed the current implementation, where the pagination response doesn’t perform a +1 look ahead.

This is the related issue I created last time https://github.com/JellyfishSDK/jellyfish/issues/1490 and to quote fx: 

> “..it’s expected a tradeoff when it comes with sort keys LSM querying.”

packages/whale-api-client/__tests__/api/governance.test.ts
```diff
@@ -322,6 +322,21 @@ describe('governance - listProposalVotes', () => {
     expect(result.length).toStrictEqual(3)
   })
 
+  it('should listProposalVotes with pagination', async () => {
+    const result1 = await client.governance.listGovProposalVotes({ id: cfpProposalId, size: 2, masternode: MasternodeType.ALL, cycle: -1 })
+    expect(result1.length).toStrictEqual(2)
+    expect(result1.hasNext).toStrictEqual(true)
```
kyleleow, line 328: Can we add assertion for `nextToken` for both result1 and result2?

thedoublejay (approved):

kyleleow (approved):

pierregee (approved):

packages/whale-api-client/__tests__/api/governance.test.ts
fuxingloh: Yeah, jellyfish LSM implementation doesn't do a +1 look ahead. Maybe we should, but it's not really a priority here. Since the clients are "taught" or rather forced to handle this.

> The purest lowest form of sorted sets implementation technically doesn't do it either since the last/next is always the last object; you can always create the "next" token yourself if you know how to do it. The "next" is merely a helper/suggestion.

fuxingloh (approved):

PR #2010, opened 2023-01-27
(chore): add flag to indicate that a txn is a coinbase txn

DieHard073055:
In order for us to ignore coinbase transactions < 100 confirmations in the `WhalePrevoutProvider`. We need a flag.

Fixes a dependency needed for https://github.com/JellyfishSDK/jellyfish/issues/1819


infinia-yzl (changes requested):
Looks like a simple flag. Looks good.

But CI fails, please fix.

PR #2016, opened 2023-01-31
fix(apps/whale-api): api error when getting all data using OCG listing endpoint but response is empty

kyleleow:
- As titled, ~to allow `nextProvider` of `ApiPagedResponse.of` to be able to return `undefined`~ to enforce data and limit to be `> 0` in `ApiPagedResponse.of`
- This is to fix the error on whale side currently when downstream is requesting all results using `listProposals()` and `listProposalVotes()` but response is empty. This is because on the upstream (`ain`) is using `size: 0` to return all results.
- To add the previously skipped test cases

Fixes #


apps/whale-api/src/module.api/governance.service.ts
```diff
@@ -99,8 +102,10 @@ export class GovernanceService {
       }))
 
       return ApiPagedResponse.of(votes, size, () => {
-        if (next === undefined) {
+        if (next === undefined && votes.length > 0) {
           return (votes.length - 1).toString()
+        } else if (next === undefined) {
+          return undefined
         }
```
fuxingloh:
```suggestion
        if (next === undefined) {
          if (votes.length > 0) {
            return (votes.length - 1).toString()
          }
          return undefined
        }
```

Prehaps?

apps/whale-api/src/module.api/governance.service.ts
```diff
@@ -51,6 +51,9 @@ export class GovernanceService {
     })
 
     return ApiPagedResponse.of(proposals, size, (proposal) => {
+      if (proposal === undefined) {
+        return undefined
+      }
```
fuxingloh: Curious, how come it returns undefined?

kyleleow: When we are trying to get all proposals, we are passing `limit: 0`, and if data happens to be empty, line 90 will attempts to assign the `next` token and throws error
https://github.com/JellyfishSDK/jellyfish/blob/e212ed1c5909dc40509fbfe8f970b1e612cd54fa/apps/whale-api/src/module.api/_core/api.paged.response.ts#L88-L95

apps/whale-api/src/module.api/governance.service.ts
kyleleow: wanted to not use nested loops lol, but sure will update

fuxingloh: Logically this reduces the `if + &`  from 3 to 2, and it's easier to follow. 

Initially, when I read it, I saw `next === undefined`; my brain will automatically think all conditions where `next is undefined` will be handled in this code branch only. But when I saw another `next === undefined` scenario in the next `if` statement, it can hard to follow since the conditions flow like "switch-cases" instead of "logical branches".

"switch-cases" you have the revisit all conditions and reevaluate each case, while for logical branches, based on the `{}` scopes, you can easily map out within this branch/scope all conditions here are when `next is undefined`.

> 🥹

apps/whale-api/src/module.api/governance.service.ts
fuxingloh: How come it can be undefined in a list?

fuxingloh: Because you edited `apps/whale-api/src/module.api/_core/api.paged.response.ts`, I'm asking cause when you edit this, you might change the fundamentals of how a list responses can be handled/return changing all implementation contractual expectation.

kyleleow: The list is not undefined, the value being accessed by the array index is undefined, i.e.: `data[-1]`.
This will only happen when `limit: 0` and when the list returned is empty. 

Okay, understand where you are coming from now, since my changes will open up the possibility that a valid paged response to have `nextToken: undefined`

Perhaps i shall:
- add this specific condition `if (data.length === limit && data.length > 0 && limit > 0)` explicitly to line 89 in `apps/whale-api/src/module.api/_core/api.paged.response.ts`
- and revert the change i did on line 88 in `apps/whale-api/src/module.api/_core/api.paged.response.ts` ?

fuxingloh: > Perhaps i shall:

Yeah this is better, but also:

Currently there is also a `@Min(1)` where only `size > 0` would be allowed.

https://github.com/JellyfishSDK/jellyfish/blob/bee44d910bf817e3c41d0117b0cca60b1d86dd55/apps/whale-api/src/module.api/_core/api.query.ts#L15

Would this case be valid in the first place?

kyleleow: > Would this case be valid in the first place?

Yea, i didn't want to change the `size` behaviour at controller level, so i changed the endpoint to accept an `all` boolean flag and handle the `size = 0` at the service level.
https://github.com/JellyfishSDK/jellyfish/blob/05af3dc1b6756114d0823389fc2fd278c25c66a9/apps/whale-api/src/module.api/governance.service.ts#L38

> Yeah this is better

Will update

fuxingloh (approved):

PR #2020, opened 2023-02-02
feat(whale-api): whitelist getaddressinfo

pierregee:
- Whitelisting `getaddressinfo` to be able to verify if address is owned by the wallet in bridge
- [ ] Checking if this can be used to verify if address is owned

Fixes #

### DIFF
--- a/apps/whale-api/src/module.api/rpc.controller.ts
+++ b/apps/whale-api/src/module.api/rpc.controller.ts
@@ -42,7 +42,8 @@ export class MethodWhitelist implements PipeTransform {
     'getrawtransaction',
     'getgovproposal',
     'listgovproposals',
-    'listgovproposalvotes'
+    'listgovproposalvotes',
+    'getaddressinfo'
   ]
 
   transform (value: string, metadata: ArgumentMetadata): string {


fuxingloh:
> * Whitelisting `getaddressinfo` to be able to verify if address is owned by the wallet in bridge
> * [ ]  Checking if this can be used to verify if address is owned

Nope it doesn't work like this, `apps/whale-api` doesn't hold an keys.

pierregee:
> > * Whitelisting `getaddressinfo` to be able to verify if address is owned by the wallet in bridge
> > * [ ]   Checking if this can be used to verify if address is owned
> 
> Nope it doesn't work like this, `apps/whale-api` doesn't hold an keys.

@fuxingloh Thanks, closing this PR then. Any thoughts on how to verify if an address is owned by a wallet?

fuxingloh:
> > > * Whitelisting `getaddressinfo` to be able to verify if address is owned by the wallet in bridge
> > > * [ ]    Checking if this can be used to verify if address is owned
> > 
> > 
> > Nope it doesn't work like this, `apps/whale-api` doesn't hold an keys.
> 
> @fuxingloh Thanks, closing this PR then. Any thoughts on how to verify if an address is owned by a wallet?

If you have the privateKey for it then you check with it; what wallet system/provider are you using?

pierregee:
WhaleWalletProvider, ok I have an idea on what to do know as discussed. Will check via the index

PR #2046, opened 2023-02-09
feat(jellyfish-api-core): update listGovProposalVotes args and output

shohamc1:
#### What this PR does / why we need it:

- Fixes broken tests after 3.2.4 changes
- Exposes new `aggregate` and `valid` args
- Adds new `valid` output

PR #2047, opened 2023-02-09
chore(jellyfish-api-core): update voteGov tests and docs after v3.2.4

shohamc1:

voteGov can now accept an operator address in addition to the masternode ID. The error was updated to account for this change, which breaks the tests. This PR updates the tests to expect the new error.

### DIFF
--- a/packages/jellyfish-api-core/src/category/governance.ts
+++ b/packages/jellyfish-api-core/src/category/governance.ts
@@ -124,7 +124,7 @@ export class Governance {
    *
    * @param {VoteData} data Vote data
    * @param {string} data.proposalId Proposal id
-   * @param {number} data.masternodeId Masternode id
+   * @param {number} data.masternodeId Masternode id/owner address/operator address
    * @param {VoteDecision} data.decision Vote decision. See VoteDecision.
    * @param {UTXO[]} [utxos = []] Specific utxos to spend
    * @param {string} [utxos.txid] The transaction id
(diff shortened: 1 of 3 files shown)


packages/jellyfish-api-core/src/category/governance.ts
```diff
@@ -124,7 +124,7 @@ export class Governance {
    *
    * @param {VoteData} data Vote data
    * @param {string} data.proposalId Proposal id
-   * @param {number} data.masternodeId Masternode id
+   * @param {number} data.masternodeId Masternode id/owner address/operator address
```
kyleleow, line 127: I wanted to a suggest a different variable name to better represent this param, but i cant think of any 😅

cwkang1998 (approved):
Looks good, but I think the PR title can be `chore(jellyfish-api-core): adjust tests and docs after defichain update` which is more accurate.

Extremely minor `nit` though, so don't worry about it.

fuxingloh:
Could you target all `v3.2.4` and related changes to a new branch from main called `ain/3.2.4`. We can't (or rather shouldn't) merge anything breaking tests into main.

shohamc1:
> Could you target all `v3.2.4` and related changes to a new branch from main called `ain/3.2.4`. We can't (or rather shouldn't) merge anything breaking tests into main.

This is the only PR that would work on 3.2.4. Both #2046 and #2048 require 3.2.5 changes. Do you want to merge this one and then combine #2046 and #2048 into a 3.2.5 branch? Or would you rather wait for 3.2.5 to be released and merge all three from a new branch?

PR #2060, opened 2023-02-20
feat(jellyfish-api-core): add wallet.listtransactions RPC

andyrobert3:

/kind feature

Implemented RPC call for https://github.com/JellyfishSDK/jellyfish/issues/48
1. wallet.listtransactions

#### Additional comments?:
- Unsure how to test the "skip" parameter for the `wallet.listtransactions` function

### DIFF
--- a/packages/jellyfish-api-core/__tests__/category/wallet/listTransactions.test.ts
+++ b/packages/jellyfish-api-core/__tests__/category/wallet/listTransactions.test.ts
@@ -64,7 +64,7 @@ describe('listTransactions', () => {
       expect(inWalletTransaction.label).toStrictEqual('owner')
     })
 
-    expect(inWalletTransactions.length).toBeGreaterThanOrEqual(numTx)
+    expect(inWalletTransactions.length).toBe(numTx)
 
     for (const inWalletTransaction of inWalletTransactions) {
       expect(inWalletTransaction).toMatchObject({
--- a/packages/jellyfish-api-core/__tests__/category/wallet/listTransactions.test.ts
+++ b/packages/jellyfish-api-core/__tests__/category/wallet/listTransactions.test.ts
@@ -171,7 +171,6 @@ describe('listTransactions', () => {
     }
   })
 
-  // TODO: it should listTransactions with excludeCustomTx = false
   it('should listTransactions with excludeCustomTx = true', async () => {
     const inWalletTransactions = await client.wallet.listTransactions('*', 10, 0, undefined, true)
 
--- a/packages/jellyfish-api-core/__tests__/category/wallet/listTransactions.test.ts
+++ b/packages/jellyfish-api-core/__tests__/category/wallet/listTransactions.test.ts
@@ -22,7 +22,7 @@ describe('listTransactions', () => {
     await client.wallet.sendToAddress(address, 0.0001)
     await container.generate(1, address)
 
-    const inWalletTransactions = await client.wallet.listTransactions()
+    const inWalletTransactions = await client.wallet.listTransactions({})
 
     expect(inWalletTransactions.length).toBeGreaterThanOrEqual(1)
 
@@ -49,7 +49,7 @@ describe('listTransactions', () => {
     await client.wallet.sendToAddress(address, 0.0001)
     await container.generate(1, address)
 
-    const inWalletTransactions = await client.wallet.listTransactions('owner')
+    const inWalletTransactions = await client.wallet.listTransactions({ label: 'owner' })
 
     inWalletTransactions.forEach((inWalletTransaction) => {
       expect(inWalletTransaction.label).toStrictEqual('owner')
@@ -80,7 +80,7 @@ describe('listTransactions', () => {
     await client.wallet.sendToAddress(address, 0.0001)
     await container.generate(1, address)
 
-    const inWalletTransactions = await client.wallet.listTransactions('owner')
+    const inWalletTransactions = await client.wallet.listTransactions({ label: 'owner' })
 
     inWalletTransactions.forEach((inWalletTransaction) => {
       expect(inWalletTransaction.label).toStrictEqual('owner')
@@ -111,7 +111,7 @@ describe('listTransactions', () => {
     await client.wallet.sendToAddress(address, 0.0001)
     await container.generate(1, address)
 
-    const inWalletTransactions = await client.wallet.listTransactions('*', 5)
+    const inWalletTransactions = await client.wallet.listTransactions({ count: 5 })
 
     expect(inWalletTransactions.length).toStrictEqual(5)
 
@@ -135,16 +135,16 @@ describe('listTransactions', () => {
   })
 
   it('should not listTransactions with count = -1', async () => {
-    await expect(client.wallet.listTransactions('*', -1)).rejects.toThrow('RpcApiError: \'Negative count\', code: -8, method: listtransactions')
+    await expect(client.wallet.listTransactions({ count: -1 })).rejects.toThrow('RpcApiError: \'Negative count\', code: -8, method: listtransactions')
   })
 
   it('should listTransactions with count = 0', async () => {
-    const inWalletTransactions = await client.wallet.listTransactions('*', 0)
+    const inWalletTransactions = await client.wallet.listTransactions({ count: 0 })
     expect(inWalletTransactions.length).toStrictEqual(0)
   })
 
   it('should listTransactions with includeWatchOnly false', async () => {
-    const inWalletTransactions = await client.wallet.listTransactions('*', 10, 0, false, undefined)
+    const inWalletTransactions = await client.wallet.listTransactions({ includeWatchOnly: false })
 
     inWalletTransactions.forEach((inWalletTransaction) => {
       expect(inWalletTransaction.address).toStrictEqual(address)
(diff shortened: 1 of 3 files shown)


packages/jellyfish-api-core/__tests__/category/wallet/listTransactions.test.ts
```diff
@@ -0,0 +1,201 @@
...
+    await container.stop()
+  })
+
+  it('should listTransactions', async () => {
+    await client.wallet.sendToAddress(address, 0.0001)
+    await container.generate(1, address)
+
+    const inWalletTransactions = await client.wallet.listTransactions()
+
+    expect(inWalletTransactions.length).toBeGreaterThanOrEqual(1)
```
eli-lim: how many are there exactly? If it's exactly `1`, then we should just test it as such?
Alternatively we can create a bunch in the test setup

packages/jellyfish-api-core/__tests__/category/wallet/listTransactions.test.ts
```diff
@@ -0,0 +1,201 @@
...
+
+    for (const inWalletTransaction of inWalletTransactions) {
+      expect(typeof inWalletTransaction.address).toStrictEqual('string')
+      expect(typeof inWalletTransaction.txid).toStrictEqual('string')
```
eli-lim: A cleaner / more helpful alternative to this is

```typescript
expect(inWalletTransaction).toStrictEqual({
  address: expect.any(String),
  txid: expect.any(String),
  ...
})
```

We get the benefit of:
- readability; seeing the whole shape of the expected object
- better diffs; jest + the IDE will show the diff between the expected object and the actual object as a whole, instead of erroring out at individual fields.
  
image

versus

image

packages/jellyfish-api-core/src/category/wallet.ts
```diff
@@ -325,6 +325,21 @@ export class Wallet {
...
+   * Returns up to 'count' most recent transactions skipping the first 'from' transactions.
+   *
+   * @param {string} label If set, should be a valid label name to return only incoming transactions with the specified label, or '*' to disable filtering and return all transactions. (default = '*')
+   * @param {string} count The number of transactions to return (default = 10)
+   * @param {string} skip The number of transactions to skip (default = 0)
+   * @param {boolean} includeWatchOnly Whether to include watch-only addresses (default = true)
+   * @param {boolean} excludeCustomTx False to include all transactions, otherwise exclude custom transactions (default = false)
+   * @return {Promise<Array<InWalletTransactionWithFlatDetails>>}
+   */
+  async listTransactions (label: string = '*', count: number = 10, skip: number = 0, includeWatchOnly: boolean = true, excludeCustomTx: boolean = false): Promise<InWalletTransactionWithFlatDetails[]> {
```
eli-lim: Since there are so many default params, should we use an object (typescript's equivalent for named parameters)?

eli-lim (commented):
Generally LGTM! Some comments

packages/jellyfish-api-core/__tests__/category/wallet/listTransactions.test.ts
andyrobert3: thanks for this, will update

andyrobert3: what tool / extension on VSCode did you use to generate the screen tab on the right

image

packages/jellyfish-api-core/src/category/wallet.ts
andyrobert3: sure, structure will be like this:

`{ label?: string, count?: number, skip?: number, includeWatchOnly?: boolean, excludeCustomTx?: boolean }`

packages/jellyfish-api-core/__tests__/category/wallet/listTransactions.test.ts
andyrobert3: how should we create a bunch of wallet transactions in the test setup? 

Can you point me to any code that does so?

andyrobert3: nvm I generated N transfers in the beginning of the tests, so we can get at least N transactions when fetching from RPC call

Then we just update the test to be
`expect(inWalletTransactions.length).toBe(X)`

packages/jellyfish-api-core/__tests__/category/wallet/listTransactions.test.ts
```diff
@@ -0,0 +1,230 @@
...
+      return await client.wallet.sendToAddress(address, 0.0001)
+    }))
+    await container.generate(1, address)
+
+    const inWalletTransactions = await client.wallet.listTransactions({})
+
+    expect(inWalletTransactions.length).toStrictEqual(10)
+
+    for (const inWalletTransaction of inWalletTransactions) {
+      expect(inWalletTransaction).toMatchObject({
```
eli-lim, line 31: Is there a reason why we prefer `toMatchObject` here over `toStrictEqual`?

eli-lim (approved):
Generally LGTM

packages/jellyfish-api-core/src/category/wallet.ts
andyrobert3: @eli-lim 

Seeing some additional returned optional parameters being returned from the RPC call that are not part of the specifications from the `.cpp` code

https://github.com/DeFiCh/ain/blob/master/src/wallet/rpcwallet.cpp
image

There are additional optional returned parameters that are not part of the official specification, which are: 
- `generated` -> boolean
- `walletconflicts` -> array

These parameters would only appear in some results, being non-existent in others

Using `toStrictEqual` will cause the tests to fail when the `generated` & `walletconflicts` parameters appear

`toMatchObject` only checks that a JavaScript object matches a subset of the properties of an object, which ignores the additional optional returned parameters `generated` & `wallet conflicts`

packages/jellyfish-api-core/__tests__/category/wallet/listTransactions.test.ts
andyrobert3: Here are the specifications on what attributes are returned by the `listtransactions` function on DeFiChain

https://github.com/DeFiCh/ain/blob/master/src/wallet/rpcwallet.cpp
```
RPCResult {
            "[\n"
            "  {\n"
            "    \"address\":\"address\",    (string) The defi address of the transaction.\n"
            "    \"category\":               (string) The transaction category.\n"
            "                \"send\"                  Transactions sent.\n"
            "                \"receive\"               Non-coinbase transactions received.\n"
            "                \"generate\"              Coinbase transactions received with more than 100 confirmations.\n"
            "                \"immature\"              Coinbase transactions received with 100 or fewer confirmations.\n"
            "                \"orphan\"                Orphaned coinbase transactions received.\n"
            "    \"amount\": x.xxx,          (numeric) The amount in " + CURRENCY_UNIT + ". This is negative for the 'send' category, and is positive\n"
            "                                        for all other categories\n"
            "    \"label\": \"label\",       (string) A comment for the address/transaction, if any\n"
            "    \"vout\": n,                (numeric) the vout value\n"
            "    \"fee\": x.xxx,             (numeric) The amount of the fee in " + CURRENCY_UNIT + ". This is negative and only available for the \n"
            "                                         'send' category of transactions.\n"
            "    \"confirmations\": n,       (numeric) The number of confirmations for the transaction. Negative confirmations indicate the\n"
            "                                         transaction conflicts with the block chain\n"
            "    \"trusted\": xxx,           (bool) Whether we consider the outputs of this unconfirmed transaction safe to spend.\n"
            "    \"blockhash\": \"hashvalue\", (string) The block hash containing the transaction.\n"
            "    \"blockindex\": n,          (numeric) The index of the transaction in the block that includes it.\n"
            "    \"blocktime\": xxx,         (numeric) The block time in seconds since epoch (1 Jan 1970 GMT).\n"
            "    \"txid\": \"transactionid\", (string) The transaction id.\n"
            "    \"time\": xxx,              (numeric) The transaction time in seconds since epoch (midnight Jan 1 1970 GMT).\n"
            "    \"timereceived\": xxx,      (numeric) The time received in seconds since epoch (midnight Jan 1 1970 GMT).\n"
            "    \"comment\": \"...\",       (string) If a comment is associated with the transaction.\n"
            "    \"bip125-replaceable\": \"yes|no|unknown\",  (string) Whether this transaction could be replaced due to BIP125 (replace-by-fee);\n"
            "                                                     may be unknown for unconfirmed transactions not in the mempool\n"
            "    \"abandoned\": xxx          (bool) 'true' if the transaction has been abandoned (inputs are respendable). Only available for the \n"
            "                                         'send' category of transactions.\n"
            "  }\n"
            "]\n"
}
```

When running the tests, sometimes there are additional optional returned parameters that are not included in the specifications above:
- `generated` -> boolean 
- `walletconflicts` -> Array

These parameters appear in some responses, and not others.

Using `toStrictEqual` may fail tests here, the extra parameters `generated` & `walletconflicts` are not always included in the response. 

However, `toMatchObject` is defined as "check that a JavaScript object matches a subset of the properties of an object"

In this case, we can just compare the response result if it matches a subset of the specification parameters from the DeFiChain repository

andyrobert3: If anyone from the DeFi chain team knows when & why the 2 extra parameters appear, feel free to add on here so I can modify my tests to use `toStrictEqual`

eli-lim: Ah okay that makes sense 

I guess the RPC doc needs updating in `ain` then

andyrobert3: do you know who should i ask here? @eli-lim 

I'll probably raise a ticket to the `ain` repo?

eli-lim: Sure, just raise a ticket

In the mean time, let's wait for others to review this PR

PR #2085, opened 2023-03-22
chore(apps/whale-api): update poolpairs to ignore token symbols with `BURN`

veralygit:
Updated `poolpair.controller.ts` 
to ignore symbols with `BURN` eg `BURN`, `BURN2` and not only `BURN`

and `poolpair.prices.services.ts` to include symbols with `BURN` as untradeableToken

Fixes #

#### Additional comments?:
This PR is to be put on hold as we are exploring on showing `BURN` without implying that they are tradable

### DIFF
--- a/apps/whale-api/src/module.api/poolpair.controller.ts
+++ b/apps/whale-api/src/module.api/poolpair.controller.ts
@@ -64,7 +64,7 @@ export class PoolPairController {
       return item.id
     })
     response.data = response.data.filter(value => {
-      return !value.symbol.startsWith('BURN-')
+      return !value.symbol.includes('BURN')
     })
     return response
   }
--- a/apps/whale-api/src/module.api/poolpair.controller.e2e.ts
+++ b/apps/whale-api/src/module.api/poolpair.controller.e2e.ts
@@ -9,7 +9,6 @@ import { DeFiDCache } from './cache/defid.cache'
 import { CachePrefix } from '@defichain-apps/libs/caches'
 import { Cache } from 'cache-manager'
 import { TokenInfo } from '@defichain/jellyfish-api-core/dist/category/token'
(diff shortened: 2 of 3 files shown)


kyleleow:
I am thinking about this again, should we consider filtering out the poolpair that has `status: false` instead?
I checked that currently `status: false` poolpair are the one after stock split and our burn bot, e.g.: `dTSLA-dDUSD/v1`, `dGOOGL-dDUSD/v1`, `dBURN2-DUSD`

And on product side we are manually filtering again `..v1` poolpair, e.g.: 
https://github.com/WavesHQ/walletkit/blob/8a3cc3d66697a0ca483c4cd29100f088ab6b104b/packages/walletkit-ui/src/store/wallet.ts#L224-L226

For backward compatibility, maybe we should by default return all pair, and accept a new param that will filter out `status: false`? Or the other way round, i.e.: default to return filtered pair, accept a new param to return unfiltered pairs

Update from `ain` team: `status: false` in means that poolpair will not be able to execute swap, but it is still free to add/remove liquidity

Update (23/03/2023 16:23): we will not be checking `status: false` because we have use case to display the poolpair before we want to make it live

apps/whale-api/src/module.api/poolpair.controller.ts
```diff
@@ -64,7 +64,7 @@ export class PoolPairController {
       return item.id
     })
     response.data = response.data.filter(value => {
-      return !value.symbol.startsWith('BURN-')
+      return !value.symbol.includes('BURN')
```
kyleleow, line 67: can we add a test for this in https://github.com/JellyfishSDK/jellyfish/blob/main/apps/whale-api/src/module.api/poolpair.controller.e2e.ts

kyleleow (approved):

PR #2093, opened 2023-04-13
feat(jellyfish-api-core, jellyfish-transaction, jellyfish-transaction-builder): add evmTx [WIP]

lykalabrada:
Adds support for creation of EVM transaction.

Fixes #


packages/testcontainers/src/containers/DeFiDContainer.ts
```diff
@@ -65,7 +66,7 @@ export abstract class DeFiDContainer extends DockerContainer {
    */
   protected getCmd (opts: StartOptions): string[] {
     return [
-      'defid',
+      'bin/defid',
```
fullstackninja864: We can remove this as in new working build defid path will get set as expected
```suggestion
      'defid',
```

lykalabrada:
Will recreate PR

PR #2134, opened 2023-07-31
feat(jellyfish-crypto): add fromAddress to convert eth address into script

lykalabrada:
To get EVM script for an ETH address that is not owned by the same wallet. Will be used on the `transferDomain` transaction builder that only accepts script as address

Fixes DFC-170

### DIFF
--- a/packages/jellyfish-crypto/package.json
+++ b/packages/jellyfish-crypto/package.json
@@ -14,6 +14,7 @@
     "build": "tsc -b ./tsconfig.build.json"
   },
   "dependencies": {
+    "@defichain/jellyfish-transaction": "^0.0.0",
     "bech32": "^2.0.0",
     "bip66": "^1.1.5",
     "browserify-aes": "^1.2.0",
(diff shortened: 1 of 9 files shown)


fullstackninja864 (dismissed):

canonbrother (dismissed):

pierregee (dismissed):

packages/jellyfish-crypto/package.json
```diff
@@ -14,6 +14,7 @@
     "build": "tsc -b ./tsconfig.build.json"
   },
   "dependencies": {
+    "@defichain/jellyfish-transaction": "^0.0.0",
```
fuxingloh: This will cause cyclic dependency, you need to remove this.

fuxingloh (changes requested):

packages/jellyfish-crypto/package.json
lykalabrada: Thanks @fuxingloh! Moved the change inside `jellyfish-address` instead. LMK if this is fine 🙏

canonbrother (approved):

fuxingloh (approved):

pierregee (approved):

PR #2139, opened 2023-08-06
feat(jellyfish-transaction-builder): Fix fees in createVOC

### LINKED ISSUE #3946: creating a DFIP fails
Trying to create a DFIP in the LW fails with "malformed tx vouts (wrong creation fee) (code 16)"

The address has enough DFI and everything seems fine until the tx is sent to ocean. 

Just tried with another address: with 51 DFI UTXOs in the address, the wallet itself says "not enough DFI UTXOs" which is also wrong.

kuegi:
currently, sending a DFIP via jellyfishSDK fails, cause it sets the wrong fees. 
Fixes https://github.com/BirthdayResearch/defichain-wallet/issues/3946

### DIFF
--- a/packages/jellyfish-transaction-builder/src/txn/txn_builder_governance.ts
+++ b/packages/jellyfish-transaction-builder/src/txn/txn_builder_governance.ts
@@ -45,13 +45,19 @@ export class TxnBuilderGovernance extends P2WPKHTxnBuilder {
     }
     let creationFee = new BigNumber('5')
     if (this.network.name === 'mainnet') {
-      creationFee = new BigNumber('50')
-    }
-    if (this.network.name === 'testnet') {
-      creationFee = new BigNumber('50')
-    }
-    if (createVoc.options === 1) {
-      creationFee = new BigNumber('5000')
+      if (createVoc.options === 1) {
+        creationFee = new BigNumber('5000')
+      } else {
+        creationFee = new BigNumber('50')
+      }
+    } else if (this.network.name === 'testnet') {
+      if (createVoc.options === 1) {
+        creationFee = new BigNumber('50')
+      } else {
+        creationFee = new BigNumber('50')
+      }
+    } else if (createVoc.options === 1) { //so we can test it with regtest params
+      creationFee = new BigNumber('10000')
     }
     return await this.createDeFiTx(
       OP_CODES.OP_DEFI_TX_CREATE_VOC(createVoc),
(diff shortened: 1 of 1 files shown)


packages/jellyfish-transaction-builder/src/txn/txn_builder_governance.ts
```diff
@@ -45,13 +45,13 @@ export class TxnBuilderGovernance extends P2WPKHTxnBuilder {
...
     if (this.network.name === 'mainnet') {
-      creationFee = new BigNumber('100')
+      creationFee = new BigNumber('50')
     }
     if (this.network.name === 'testnet') {
       creationFee = new BigNumber('50')
     }
     if (createVoc.options === 1) {
-      creationFee = new BigNumber('10000')
+      creationFee = new BigNumber('5000')
```
lykalabrada: Hi @kuegi some test is failing, can you help keep this `10000` for testnet, and only set `5000` for mainnet? thanks!

lykalabrada (changes requested):

lykalabrada:
For more info, on chain we have set the following gov attributes:
image

packages/jellyfish-transaction-builder/src/txn/txn_builder_governance.ts
kuegi: on testnet, the 10000 was false before already: 
image
longterm, ocean should provide the govs and this should take it directly from the chain. But IMHo now its most important to get this fixed for users. Cause noone is able to create a VOC via the LW.

kuegi: @lykalabrada what network/networkparams are the tests running on?

kuegi: adapted it so that testnet is correct too now (but still hardcoded) and regtest in tests keep working for now.

lykalabrada (approved):

packages/jellyfish-transaction-builder/src/txn/txn_builder_governance.ts
```diff
@@ -45,12 +45,18 @@ export class TxnBuilderGovernance extends P2WPKHTxnBuilder {
...
+      } else {
+        creationFee = new BigNumber('50')
+      }
+    } else if (this.network.name === 'testnet') {
+      if (createVoc.options === 1) {
+        creationFee = new BigNumber('50')
+      } else {
+        creationFee = new BigNumber('50')
+      }
+    } else if (createVoc.options === 1) { //so we can test it with regtest params
```
lykalabrada:
```suggestion
    } else if (createVoc.options === 1) {
```

lykalabrada (changes requested):

kuegi:
@lykalabrada can this be merged? Imho it's important to get this in soon so that users can finally make dfips via lightwallet.
This bug is really bad for the governance!

pierregee (approved):

lykalabrada (approved):

fuxingloh (approved):

PR #2140, opened 2023-08-07
chore(tests): add tests for evm feature flags

lykalabrada:
Added feature gating tests for EVM

Fixes DFC-167


packages/jellyfish-api-core/__tests__/category/evm/evmFeatureFlag.test.ts
```diff
@@ -0,0 +1,284 @@
...
+          'v0/transferdomain/dvm-evm/enabled': 'false',
+          'v0/transferdomain/evm-dvm/enabled': 'false'
+        }
+      })
+      await container.generate(1)
+    })
+
+    it('should still successfully get eth address', async () => {
+      const newEvmAddr = await container.getNewAddress('eth', 'eth')
+      expect(newEvmAddr).toBeDefined()
```
fullstackninja864, line 99: we can add this checks too here 
```
expect(newEvmAddr).toMatch(/^0x[a-fA-F0-9]{40}$/gm)
expect(newEvmAddr.length).toStrictEqual(42)

File: packages/jellyfish-api-core/__tests__/category/evm/evmFeatureFlag.test.ts [hunk 2]
```diff
@@ -0,0 +1,284 @@
...
+    })
+
+    it('should still successfully get eth balance', async () => {
```
fullstackninja864, line 104: this test case is of no use as we are not getting any change in balance, I guess we can remove this

fullstackninja864 (changes requested):

packages/jellyfish-api-core/__tests__/category/evm/evmFeatureFlag.test.ts
lykalabrada: i removed the dvm balance, but keeping the evm balance here just so we know includeEth is not affected by EVM flags at all

fullstackninja864: in that case lets first transfer some token to evm and then only we will be able to check that diff

fullstackninja864 (dismissed):

PR #2155, opened 2023-09-22
Fix(apps/whale-api): rm error before index

canonbrother:

- should not throw error before index as err will trigger cleanup to invalidate non-existent data
- ensure poolpair.pathing caches the latest

Fixes #


canonbrother:
close https://github.com/BirthdayResearch/jellyfishsdk/pull/2170

PR #2171, opened 2023-11-22
chore(apps/whale-api): add enable index env

canonbrother:

debugging purpose only


asyrafnorafandi (commented):
So far looks good to me. Is this by default `TRUE`?

PR #2173, opened 2023-11-27
chore(): 4.0.5

canonbrother:

Fixes #

### DIFF
--- a/docs/node/CATEGORIES/08-account.md
+++ b/docs/node/CATEGORIES/08-account.md
@@ -159,7 +159,12 @@ Create a transfer domain transaction submitted to a connected node.
 
 ```ts title="client.account.transferDomain()"
 interface account {
-  transferDomain (payload: Array<Record<string, TransferDomainInfo>>): Promise<string>
+  transferDomain (payload: Array<Record<TransferDomainKey, TransferDomainInfo> | Record<TransferDomainOptionalKey, TransferDomainOptionalInfo>>): Promise<string>
+}
+
+enum TransferDomainKey {
+  SRC = 'src',
+  DST = 'dst'
 }
 
 interface TransferDomainInfo {
@@ -176,7 +181,15 @@ enum TransferDomainType {
   DVM = 2,
   /** type for EVM To DVM Token transfer */
   EVM = 3,
-};
+}
+
+enum TransferDomainOptionalKey {
+  SINGLE_KEY_CHECK = 'singlekeycheck',
+}
+
+interface TransferDomainOptionalInfo {
+  singleKeyCheck: true
+}
 ```
 
 ## accountToUtxos
--- a/docs/node/CATEGORIES/08-account.md
+++ b/docs/node/CATEGORIES/08-account.md
@@ -188,7 +188,7 @@ enum TransferDomainOptionalKey {
 }
 
 interface TransferDomainOptionalInfo {
-  singleKeyCheck: true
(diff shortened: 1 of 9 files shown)


docs/node/CATEGORIES/08-account.md
```diff
@@ -176,7 +181,15 @@ enum TransferDomainType {
...
   EVM = 3,
-};
+}
+
+enum TransferDomainOptionalKey {
+  SINGLE_KEY_CHECK = 'singlekeycheck',
+}
+
+interface TransferDomainOptionalInfo {
+  singleKeyCheck: true
```
fullstackninja864:
```suggestion
  singleKeyCheck: boolean
```
it should be boolean ?

sieniven:
singlekeycheck flag can be passed both on the RPC and/or set directly as a node config on start up. Reference: https://github.com/DeFiCh/ain/pull/2714 and https://github.com/DeFiCh/ain/pull/2709. To maintain backwards compatibility, a better fix could potentially be to pass the singlekeycheck flag directly on the node start up, which will default to skipping the singlekeycheck pipeline on transferdomain RPCs.

packages/jellyfish-api-core/__tests__/category/account/transferDomain.test.ts
```diff
@@ -233,7 +248,8 @@ describe('TransferDomain', () => {
             address: dvmAddr,
             amount: '3@DFI',
             domain: TransferDomainType.DVM
-          }
+          },
+          singlekeycheck: false as unknown as TransferDomainOptionalInfo
```
sieniven: singlekeycheck flag can be passed both on the RPC and/or set directly as a node config on start up. Reference: https://github.com/DeFiCh/ain/pull/2714 and https://github.com/DeFiCh/ain/pull/2709. To maintain backwards compatibility, a better fix could potentially be to pass the singlekeycheck flag directly on the node start up, which will default to skipping the singlekeycheck pipeline on transferdomain RPCs.

docs/node/CATEGORIES/08-account.md
canonbrother: true | false is boolean.. specify `true` means "true by default"
kind of
```
function get(item: any: isChecked: boolean)
vs
function get(item: any, isChecked: true)
```

packages/jellyfish-api-core/__tests__/category/account/transferDomain.test.ts
canonbrother: cleaner now
[7525d04](https://github.com/BirthdayResearch/jellyfishsdk/pull/2173/commits/7525d04f86ff7cf48d3ee9b971a214fa44289193)

hmm.. we still need show case how to TD with singlekeycheck params
so added a simple one [e7927cc](https://github.com/BirthdayResearch/jellyfishsdk/pull/2173/commits/e7927cc9842029559eee4c2763321936db1905a4)

docs/node/CATEGORIES/08-account.md
canonbrother: @fullstackninja864 
my bad.. that's not the right way on TS
as `Typescript - type 'false' is not assignable to type 'true'`...
[3df346c](https://github.com/BirthdayResearch/jellyfishsdk/pull/2173/commits/3df346c1ec939132dab28b6cfd995d3870380d27)

PR #2174, opened 2023-12-01
chore(): support whale test suite for defid ocean

canonbrother:

Allow ocean test-suites running on Defid
Its tricky to map Defid Ocean API and function with Whale Ocean
eg:
```ts
if (app instanceof DefidBin) {
      controller = app.addressController
} else {
      controller = app.get(AddressController)
}
```

`Sinonjs`-like is an ideal approach


canonbrother:
refer to https://github.com/BirthdayResearch/jellyfishsdk/pull/2177

